PR #7, opened 2021-09-02
Fix model imports in sample code

sk-zk:


generall:
Hi, thanks for reporting this!  The documentation for in this project is actually auto-generated from OpenAPI schema. And the template for `*.md` files are unfortunately mismatch generated classes. I am going to completely remove this doc files so they won't create a confusion. The proper documentation for python client will be released soon.

PR #18, opened 2022-02-02
V0.5.x

generall:


agourlay:
How did you validate the new client version? Have you tried it in the [benchmarks](https://github.com/qdrant/benchmark)?

generall:
> How did you validate the new client version? Have you tried it in the [benchmarks](https://github.com/qdrant/benchmark)?

There are some integration tests in place, which require actual service instance running. But haven't check benchmark

PR #26, opened 2022-03-04
Upd version - 0.7.0

generall:


tests/test_qdrant_client.py
```diff
@@ -69,7 +70,7 @@ def test_qdrant_client_integration(prefer_grpc):
         vectors=vectors,
         payload=payload,
         ids=None,  # Let client auto-assign sequential ids
-        parallel=2
+        parallel=1
```
monatis, line 73: Why this change? It would be nice to keep it to catch any strange behavior due to multithreading.

monatis: but it's definitely a NIT.

tests/test_qdrant_client.py
```diff
@@ -79,7 +80,7 @@ def test_qdrant_client_integration(prefer_grpc):
 
     # Create payload index for field `random_num`
     # If indexed field appear in filtering condition - search operation could be performed faster
-    index_create_result = client.create_payload_index(COLLECTION_NAME, "random_num")
+    index_create_result = client.create_payload_index(COLLECTION_NAME, "random_num", PayloadSchemaType.FLOAT)
```
monatis, line 83: I understand that the name `PayloadSchemaType` is a convention from the previous API with restricted data types, but it might be a good idea to rename it to `PayloadIndexType` in a future release before v1.0.

monatis (approved):
LGTM --I only left some NIT comments.

tests/test_qdrant_client.py
generall: Actually, I changed this to debug something. Might be it would be better to improve error reporting in parallel uploading mode :writing_hand:

tests/test_qdrant_client.py
generall: There should be no problem to rename, will look at it

tests/test_qdrant_client.py
monatis: yeah you're right --it would be better to test them in isolation 👍

tests/test_qdrant_client.py
generall: `PayloadIndexType` is already taken by the type of index itself :disappointed:, might be `PayloadSchemaType` is not too bad? :thinking:

tests/test_qdrant_client.py
generall: I found out that error was actually logged, but as a side-effect I also fixed the proper shutdown in case of exception

tests/test_qdrant_client.py
monatis: `PayloadSchemaType` is not too bad

I think so. Let's release v0.7 as is, and we may think of a workaround in a future release if we observe a confusion among users?

PR #27, opened 2022-03-07
move qdrant_openapi_client into qdrant_client.http

generall:


joein:
There is an outdated import in `README.md`

joein (changes requested):
Also there is an outdated import in `README.md`

joein (approved):

monatis:
We should also update the code snippet in readme, shouldn't we?

PR #28, opened 2022-03-08
upload collection with gPRC

generall:


qdrant_client/qdrant_client.py
```diff
@@ -1,73 +1,35 @@
...
     # If True - single-value payload arrays will be replaced with just values.
     # Warn: Deprecated
     unwrap_payload = False
 
-    def __init__(self, host="localhost", port=6333, **kwargs):
+    def __init__(self,
+                 host="localhost",
+                 port=6333,
+                 grpc_port=6334,
+                 prefer_grpc=False,
```
monatis, line 25: Currently, it seems that GRPC is only used in `upload` method. Since we confirmed that GRPC provides a huge speedup in qdrant/qdrant#351, I believe we can invest more to support GRPC as a first-class citizen in the library. 

Things we can improve:
- Create GRPC client and initialize clients with it as class properties: It may be something like `QdrantClient.grpc.collections`, `QdrantClient.grpc.points` as opposed to `QdrantClient.http`.
- Or, we can separate client classes, but I guess the first one is more user-friendly.

tools/generate_grpc_client.sh
```diff
@@ -0,0 +1,22 @@
...
+PROTO_DIR="$(pwd)/src/tonic/proto"
+
+# Ensure current path is project root
+cd "$(dirname "$0")/../"
+
+CLIENT_DIR="qdrant_client/grpc/"
+
+cp $PROTO_DIR/*.proto $CLIENT_DIR
+
+python -m grpc_tools.protoc -I . --python_betterproto_out=$CLIENT_DIR $CLIENT_DIR/collections.proto $CLIENT_DIR/points.proto
```
monatis, line 18: Why not include `qdrant.proto`?

qdrant_client/uploader/grpc_uploader.py
```diff
@@ -0,0 +1,113 @@
...
+            if all(isinstance(v, dict) and 'lon' in v and 'lat' in v for v in val):
+                res[key] = grpc.Payload(geo=GeoPayload(
+                    values=[grpc.GeoPoint(lat=v['lat'], lon=v['lon']) for v in val],
+                ))
+                continue
+        raise RuntimeError(f"Payload {key} have unsupported type {type(val)}")
+    return res
+
+
+async def upload_batch_grpc(service: PointsStub, collection_name: str, batch):
```
monatis, line 74: In GRPC context, the name `service` is usually used for server-side implementations, and instances of `XxxStub` are usually named `xxx_client`. This is just a nit, but I believe it's better to stick to the convention.

monatis (changes requested):

monatis:
Also, maybe we should mention the GRPC client in readme as well, with a note of "at an early stage," and citing the benchmarking results. This can help early adoption and getting useful feedback.

qdrant_client/grpc/collections.proto
```diff
@@ -0,0 +1,205 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
joein: Yes, it makes sense not to keep `.proto` right here and for example no one could want to change it in this repo and there are no excess files here.

But I think that I'd like to see an exact version of `.proto` files from which current `.py` files were compiled. Maybe we can write a link to `README.md` to an exact commit in `qdrant` from which `.proto` files have been fetched or something similar.

qdrant_client/qdrant_client.py
generall: will do in later PRs

qdrant_client/grpc/collections.proto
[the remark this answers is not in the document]
generall: Documentation for client - is one of the big further tasks, fyi

monatis (approved):

joein (approved):
Functions in tests could be wrapped as fixtures, also there are some tools for benchmarking like `pytest-benchmark`. But I think that it is not a case at the moment.

PR #59, opened 2022-08-20
Fixed #58

### LINKED ISSUE #58: Qdrant throws 'NoneType' object has no attribute 'CLOSED' when program exits
## Current Behavior

When the program exits, I see this error:
````
Exception ignored in: <function QdrantClient.__del__ at 0x7f6d7848e280>
Traceback (most recent call last):
  File "/anaconda3/lib/python3.8/site-packages/qdrant_client/qdrant_client.py", line 74, in __del__
  File "/anaconda3/lib/python3.8/site-packages/httpx/_client.py", line 1256, in close
  File "/anaconda3/lib/python3.8/site-packages/httpx/_transports/default.py", line 230, in close
  File "/anaconda3/lib/python3.8/site-packages/httpcore/_sync/connection_pool.py", line 307, in close
  File "/anaconda3/lib/python3.8/site-packages/httpcore/_sync/connection.py", line 159, in close
  File "/anaconda3/lib/python3.8/site-packages/httpcore/_sync/http11.py", line 216, in close
AttributeError: 'NoneType' object has no attribute 'CLOSED'
````

## Steps to Reproduce

````python

client = QdrantClient(host="localhost", port=6333)
print(client.get_collections())
````

## Expected Behavior

Either the program should exit without any exceptions or I should be able to catch it.

## Context (Environment)

- I am using `qdrant-client==0.9.0` on `python 3.8`
- The quadrant server is run via docker compose
````
version: "3.8"

services:
  qdrant:
    image: qdrant/qdrant:v0.9.1
    container_name: qdrant_embedding_search
    ports:
      - 6333:6333
      - 6334:6334
    env_file:
      - ./.env
    volumes:
      - ./data:/qdrant/storage
      - ./config/qdrant.yaml:/qdrant/config/production.yaml
   ````

## Detailed Description

It seems the problem is with the `__del__` [implementation](https://github.com/qdrant/qdrant_client/blob/master/qdrant_client/qdrant_client.py#L71-L74]) when the program exists it tries to close the grpc and openapi connections but ends up throwing exception.

## Possible Implementation

A workaround I found is to close the connections manually before exiting the program:
````python
import time
import numpy as np
from qdrant_client.http.models import Distance

client = QdrantClient(host="localhost", port=6333)
print(client.get_collections())

# Close connection manually
if client._grpc_channel is not None:
    client._grpc_channel.close()
client.openapi_client.client._client.close()
````

Please wrap the closing method inside try blocks and handle the exception or provide a `.close()` method for `QdrantClient` so the user can close it manually.

### PR DESCRIPTION
**Pull request by imneonizer**

### CODE REVIEW

File: qdrant_client/qdrant_client.py [hunk 1]
```diff
@@ -69,9 +69,17 @@ def __init__(self,
...
+
+    def close(self):
         if self._grpc_channel is not None:
             self._grpc_channel.close()
-        self.openapi_client.client._client.close()
+
+        try:
+            self.openapi_client.client._client.close()
+        except AttributeError as e:
+            if str(e) != "'NoneType' object has no attribute 'CLOSED'":
```
generall, line 81: That is usually a bad practice to check directly for an error message.
It would be best to have a way to reliably reproduce the error. After which we could wind the cause of this error instead of just patching the consequences.

generall (changes requested):
@imneonizer please provide a reproducible scenario for the error, so we could investigate the cause of this behavior.

imneonizer:
I think the problem is specific to my system, I tested inside docker It didn't appear.

PR #61, opened 2022-08-22
api-key auth for rest and grpc

generall:


qdrant_client/qdrant_client.py
```diff
@@ -53,57 +55,77 @@ def __init__(self,
...
+
+        self._https = https
+        self._api_key = api_key
+
+        http2 = kwargs.pop("http2", False)
+        grpc_headers = {}
+        rest_headers = kwargs.pop("headers", {})
+        if api_key is not None:
+            if self._https is False:
+                warnings.warn("Api key is used with unsecure connection.")
```
agourlay, line 74: Did you check if the load balancer accept HTTP traffic with an API key?

agourlay (approved):

qdrant_client/qdrant_client.py
generall: our cloud is https only

PR #67, opened 2022-09-01
Fast protobuf

generall:
Replace betterproto with default google grpcio, as it is 20x faster serialization


agourlay:
Do you remember what was the rationale for using `betterproto` in first place?
Just want to make sure we have enough context captured here.

generall:
> Do you remember what was the rationale for using betterproto in first place?


The reason was, that `betterproto` generated structures looks like this:

```
@dataclass(eq=False, repr=False)
class UpdateCollection(betterproto.Message):
    collection_name: str = betterproto.string_field(1)
    optimizers_config: Optional["OptimizersConfigDiff"] = betterproto.message_field(
        2, optional=True, group="_optimizers_config"
    )
    timeout: Optional[int] = betterproto.uint64_field(
        3, optional=True, group="_timeout"
    )

```

not like this

```
UpdateCollection = _reflection.GeneratedProtocolMessageType('UpdateCollection', (_message.Message,), {
  'DESCRIPTOR' : _UPDATECOLLECTION,
  '__module__' : 'collections_pb2'
  # @@protoc_insertion_point(class_scope:qdrant.UpdateCollection)
  })
_sym_db.RegisterMessage(UpdateCollection)
```

agourlay:
How much impact is there for the users of the client? Do they need to change some call sites or is it completely transparent?

generall:
> How much impact is there for the users of the client? Do they need to change some call sites or is it completely transparent?

I would say, it is 95% same, buy there are corner cases

PR #73, opened 2022-09-25
updated loguru and grpcio vers

fursovia:
- There is a `0.6.0` version of `loguru`. Can qdrant support it?
- Unfortunately, tritonclient only supports [grpcio==1.41.0](https://github.com/triton-inference-server/client/blob/efd1cbe0cca60e750aed395c78b3ee5c90139add/src/python/library/requirements/requirements_grpc.txt#L34). Can we downgrade?

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -639,7 +639,7 @@ testing = ["func-timeout", "jaraco.itertools", "pytest (>=6)", "pytest-black (>=
 [metadata]
 lock-version = "1.1"
 python-versions = ">=3.7,<4.0"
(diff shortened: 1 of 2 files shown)


generall:
Hi @fursovia, nice to see you here! 

Regarding loguru, I don't mind even excluding it from dependencies, but downgrade grpcio does not look like a good long-term strategy. Is it possible to set minimal working version instead?

generall:
Will this work for your case https://github.com/qdrant/qdrant_client/pull/74 ?

PR #74, opened 2022-09-26
lower dependency requirements

generall:
- remove loguru
- lower httpx requirements
- lower grpcio requirements

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,14 +15,12 @@ keywords = ["vector", "search", "neural", "matching", "client"]
 
 [tool.poetry.dependencies]
 python = ">=3.7,<4.0"
-httpx = { version = "^0.23.0", extras = ["http2"] }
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -15,14 +15,12 @@ keywords = ["vector", "search", "neural", "matching", "client"]
 
 [tool.poetry.dependencies]
 python = ">=3.7,<4.0"
-httpx = { version = "^0.23.0", extras = ["http2"] }
-numpy = "^1.21"
+httpx = { version = ">=0.14.0", extras = ["http2"] }
```
fursovia, line 18: can it be at least `0.13.3`?

generall: if this is critical for your use-case, you can always just make raw http requests without any client

fursovia (approved):
super!

PR #83, opened 2022-12-27
Add `prefix` optional argument

cesumilo:
Hi 👋  I didn't find a contributing documentation but I took the freedom to create this PR so maybe you will agree to add this feature to the python client.

## Context
We would like to send requests to our Qdrant cluster on the same domain name as our API but with a prefix `/qdrant` for example (e.g. `api.inarix.com/qdrant`).

## Solution
I added an optional `prefix` argument to the constructor and merged it to the rest uri if it exists.

Feel free to tell me what I need to do to be accepted and merged 😄

### DIFF
--- a/qdrant_client/qdrant_client.py
+++ b/qdrant_client/qdrant_client.py
@@ -59,11 +59,16 @@ class QdrantClient:
                  prefer_grpc=False,
                  https=None,
                  api_key=None,
+                 prefix=None,
                  **kwargs):
         self._prefer_grpc = prefer_grpc
         self._grpc_port = grpc_port
         self._host = host
         self._port = port
+        self._prefix = prefix
+
+        if prefix is not None and len(prefix) > 0 and prefix[0] != '/':
+            self._prefix = f"/{prefix}"
 
         self._https = https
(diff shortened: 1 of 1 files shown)


qdrant_client/qdrant_client.py
```diff
@@ -49,6 +49,7 @@ class QdrantClient:
         prefer_grpc: If `true` - use gPRC interface whenever possible in custom methods.
         https: If `true` - use HTTPS(SSL) protocol. Default: `false`
         api_key: API key for authentication in Qdrant Cloud. Default: `None`
+        prefix: If not `None` - add `prefix` after `port` in the rest uri. Default: `None`
```
generall, line 52:
```suggestion
        prefix:
            If not `None` - add `prefix` to the REST URL path.
            Example: `service/v1` will result in `http://localhost:6333/service/v1/{qdrant-endpoint}` for REST API.
            Default: `None`
```

cesumilo: Thanks for the suggestion 👍

generall (approved):

PR #85, opened 2022-12-29
Fix `delete` missing return when HTTP

gabrielmbmb:
I've added the missing return statement to the `QdrantClient.delete` method for when HTTP is used.


generall (approved):

generall:
Hi @gabrielmbmb! Thanks for the fix!

PR #86, opened 2022-12-29
explicit timeout param

generall:
Related issue: https://github.com/qdrant/qdrant_client/issues/84

Added explicit `timeout` parameter to `QdrantClient`.
Parameter affects both REST and gRPC requests.


ncoder (approved):
Good start. I would have done this the same way.

PR #91, opened 2023-01-19
new: add py.typed, mypy config and fix some types

joein:


joein:
Install mypy with `pip3 install mypy`

Run mypy from the root directory with `mypy .`
It should show 43 errors in our files, which we probably can handle

Run mypy with `mypy . --strict` to see 133 errors :)

The following directories are ignored:

`qdrant_client.grpc` - autogenerated
`qdrant_client.http` - autogenerated

`qdrant_client.conversions` - might be tricky due to protobuf imports, postponed

`tests` - not sure if we want to apply mypy to tests

joein:
Well, seems like I broke some tests with my updates to `parallel_preprocessor.py`, I'll have a look at that

joein:
At this point `mypy .` does not complain about anything.
Let's look at current state to decide if we want to proceed with `mypy . --strict`

generall (approved):

Issue #88, opened 2023-01-12
Providing a hostname that contains `https://` causes connection error


nleroy917:
### Overview
I was using the qdrant client to connect to a qdrant instance hosted on qdrant cloud. When I provided `QdrantClient` with my hostname that contained `https://` I noticed that it refused to connect as it couldn't find the host. Removing `https://` solves the problem.

### Steps to reproduce:
1. Launch qdrant instance on qdrant cloud:
2. Set env variables:
```
export QDRANT_HOST="https://my-hostname.qdrant.io"
export QDRANT_PORT=6333
export QDRANT_API_KEY=my-api-key
```
3. Attempt connection to qdrant and pull information:
```python
import os
from qdrant_client import QdrantClient

qdrant = QdrantClient(
  host=os.environ.get("QDRANT_HOST"),
  port=os.environ.get("QDRANT_PORT"),
  api_key=os.environ.get("QDRANT_API_KEY")
)
collection_info = qdrant.get_collection("my-collection")
```
### Observed Behavior:
Error received:
```
ResponseHandlingException
[Errno 8] nodename nor servname provided, or not known
```

### Expected Behavior:
I would expect `QdrantClient` to successfully connect to a host regardless if it was prefixed with `https://` or not.

### Notes
I was able to figure out that removing `https://` would resolve the issue by inspecting the code. I [see](https://github.com/qdrant/qdrant_client/blob/9b64ada9f569c9f46e2c9d8eea74d683716f9a6a/qdrant_client/qdrant_client.py#L106) it's building up the scheme based on parameters. It'd be nice if it sanitized the URI prior to this, or at least warned a user that they have `http` in their host.

generall:
Hi @nleroy917, in URL, the host name does not contain the schema by definition.
But I would agree, that it might be better to accept full URL, rather than just a host name

nleroy917:
Hi @generall, thanks for the reply. I had started on a PR but got tripped up on some things in the development guide. Namely, the generation of the REST code: `bash -x tools/generate_rest_client.sh`.

I can try again.

generall:
Fixed in https://github.com/qdrant/qdrant_client/pull/99

Issue #92, opened 2023-01-19
Automatically get host and port from an environment variable


ChickenTarm:
It would be useful to be able to do
```python
qc.QdrantClient()
```
where the client will check if there the environment variables "QDRANT_IP, QDRANT_PORT, QDRANT_GRPC_PORT" present and if so will use those value.

generall:
Closed in favor of url param, please consider using

```
qc.QdrantClient(url=os.getenv("QDRANT_URL"))
```

PR #94, opened 2023-01-23
Strip scheme from host parameter if provided when creating `QdrantClient` instance

nleroy917:
Details are in #88 

This just adds a helper function to the `QdrantClient` class that strips the [scheme](https://en.wikipedia.org/wiki/List_of_URI_schemes) from the hostname prior to self assigning the attribute and generating the `rest_uri`.

### DIFF
--- a/qdrant_client/qdrant_client.py
+++ b/qdrant_client/qdrant_client.py
@@ -5,6 +5,7 @@ from typing import Optional, Iterable, List, Union, Tuple, Type, Dict
 import httpx
 import numpy as np
 import logging
+from urllib.parse import urlparse
 
 from qdrant_client import grpc as grpc
 from qdrant_client.connection import get_channel
@@ -70,7 +71,7 @@ class QdrantClient:
                  **kwargs):
         self._prefer_grpc = prefer_grpc
         self._grpc_port = grpc_port
-        self._host = host
+        self._host = self._remove_scheme_from_host(host)
         self._port = port
         self._prefix = prefix
         self._timeout = timeout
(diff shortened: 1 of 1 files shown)


generall:
Hi @nleroy917, thanks for the PR!

Would it be possible to try to avoid mixing of URL and Host concepts, as it could lead to misinterpretations?
For example, `https://localhost` parsed as URL should mean that we are intended to use 443 port and https protocol for communication. But code in the PR will discard this information without any warnings.

Would it be possible to make additional optional argument  `url`, which will be fully parsed into correct client configuration?

nleroy917:
Sure. Do you mean sort of the paradigm that database connection libraries use? Where you can provide a connection string (`postgres://user:password@localhost:5432`) **or** the individual params (`host=localhost`, `password=...`, etc).

generall:
> Sure. Do you mean sort of the paradigm that database connection libraries use? Where you can provide a connection string (`postgres://user:password@localhost:5432`) **or** the individual params (`host=localhost`, `password=...`, etc).
> 

Exactly! We could have additional optional parameter `url`, it it is provided, we can parse it into schema, port and host

generall:
> Do you intend this to be in __init__ or in the new helper function on the class?

I think extending `__init__` will be reasonable

nleroy917:
I'll work on this in the coming days. In the meantime, wondering if this flow seems reasonable for establishing the connection:

```mermaid
graph LR
  A[<code>__init__</code>] --> B{<code> url == None? </code>}
  B -- No --> C[Parse url into host and port]
  B -- Yes --> D[use provided host and port]
```

I also saw #92 I don't know if it's reasonable to add that functionality. But the diagram could be updated to look more like:

```mermaid
graph LR
  A[__init__] --> B{env vars?}
  B -- Yes --> C[connect with env vars]
  B -- No --> D{<code> url == None ? </code>}
  D -- No --> E[Parse url into host and port]
  D -- Yes --> F[use provided host and port]
```

I'm personally *opposed* to initiating from env variables, but thats just my opinion 😅

generall:
I also don't think ENV should be parsed internally

generall:
Hi @nleroy917, we intercepted your PR a bit https://github.com/qdrant/qdrant_client/pull/99, hope it solves the issue for you!
Please feel free to suggest any other improvements.

nleroy917:
@generall no worries! I was moving too slowly. I'll let you know if I run into other issues...

PR #126, opened 2023-03-01
new: add pre-commit

joein:


joein:
In current setup all checks are applied to changed files.

Hence, once we try to update auto-generated code, black will be applied to the corresponding files.

I haven't included black reformatting into this PR yet, I will add it when @generall approves changes in config and README.

joein:
@generall you might want to increase black line length to 99

generall (approved):

PR #131, opened 2023-03-16
Constraint grpcio-tools to be fine with protobuf > 4.0.0

monatis:
A fix attempt for [this issue on Discord](https://discord.com/channels/907569970500743200/928649303898075196/1085781334028451840).


generall:
That is not the right approach to solve the problem.

PR #142, opened 2023-03-28
wip: universal entry point

generall:


tests/test_qdrant_client.py
```diff
@@ -62,44 +64,66 @@ def create_random_vectors():
     return vectors_path
 
 
-def test_rest_init():
+def test_client_init():
+    import tempfile
```
joein, line 68: why is it here

joein (approved):
Well, yeah, considered a similar option too

Looks ok, don't really like the naming though)) this `impl` thing and the fact that the attr is not protected

tests/test_qdrant_client.py
generall: to test `path=/tmp/...`

joein: I mean why the import is inside the function, any particular reason?

joein: it's just a nit comment

PR #154, opened 2023-04-13
Fix MatchText filters in local mode

kacperlukawski:
I fixed the case when a particular payload does not have a corresponding attribute. In such a case, we have `value = None`, and the condition ends with `TypeError: argument of type 'NoneType' is not iterable`


joein:
Isn't it better to have `value is not None` ?
Like `"" in ""` is `True`, but `value and "" in ""` is False

kacperlukawski:
@joein That's right! I didn't think somebody might be using an empty filter, but that may happen.

generall (approved):

PR #158, opened 2023-04-20
Fix local upload collection

joein:


generall:
What was the problem there?

joein:
In `QdrantLocal.upload_collection`

`zip_longest` caused infinite loop when `ids` were an infinite sequence (same problem with payload)

generall (approved):

PR #159, opened 2023-04-20
expose raw async grpc client

generall:
Expose raw async gRPC client. Channel is created by default Qdrant client, but raw queries function could be used to speed optimization


joein:
Seems like there is no convenient way to instantiate async grpc channels with custom headers (e.g. to pass `api-key`), since those channels are created only on property invocation.

We also always create sync grpc channels if `prefer_grpc` is set, even if users don't want to use sync grpc.

First thing that came to my mind was to stop creating sync channels in `__init__`, instead create them on property invocation (if they haven't been created yet) and use `self._grpc_header` instead of passing it as an argument to `_init_..._client`.

generall:
@joein missed that, thanks!

joein (approved):

Issue #166, opened 2023-04-28
Persisting high-dimensional vectors in QdrantLocal


joein:
We have problems with persisting of high-dimensional vectors in QdrantLocal
If we try to save vectors with dimensionality >= 433, it fails (see traceback in the bottom)

We also don't properly close db in `LocalCollection`

I read some sources about dbm or shelve (shelve is a layer on top of dbm which provides interface to store values in any format) and there were written, that these modules work with fairly small values. 
[dbm docs](https://docs.python.org/3/library/dbm.html) (look for `Warning` in the bottom of the page)
[shelve docs](https://docs.python.org/3/library/shelve.html#restrictions)
[blog](https://www.zachburchill.ml/scraping_problems/#fnref:4) (some guy also encountered NegativeSize exception)

Code snippet to reproduce:

```Python
import random
import numpy as np
import qdrant_client
import qdrant_client.http.models as rest

default_collection_name = "example"
qdrant_args = {"path": "qdrant"}


def ingest_data(vector_size=433):  # vector_size < 433: works, vector_size >= 433: crashes
    lines = [x for x in range(10)]
    embeddings = np.random.randn(len(lines), vector_size).tolist()
    client = qdrant_client.QdrantClient(**qdrant_args)
    client.recreate_collection(default_collection_name, vectors_config=rest.VectorParams(
                size=vector_size,
                distance=rest.Distance.COSINE,
            ),)
    client.upsert(
        collection_name=default_collection_name,
        points=rest.Batch.construct(
            ids=[random.randint(0, 100) for _ in range(len(lines))],
            vectors=embeddings,
        )
    )


if __name__ == '__main__':
    ingest_data()
    ingest_data()  # Crashes here. Here we have the second instance

```

Traceback
```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/qdrant/qdrant-local-issue/raw.py", line 29, in <module>
    ingest_data()
  File "/Users/qdrant/qdrant-local-issue/raw.py", line 13, in ingest_data
    client = qdrant_client.QdrantClient(**qdrant_args)
[...]

PR #167, opened 2023-05-01
test for large vectors persistence

generall:
Switch persistence storage to sqlite for the local mode


generall:
Fixes https://github.com/qdrant/qdrant-client/issues/166

joein:
We could just in case explicitly close the connection in `__del__`. (However it might not be necessary)

joein:
I think we might need to test local client not only with `:memory:`, but also with `path=tmpdir` in congruence tests in the future (to be more confident about it)

joein (approved):

tests/test_local_persistence.py
```diff
@@ -0,0 +1,40 @@
...
+            distance=rest.Distance.COSINE,
+        ),
+    )
+
+    client.upsert(
+        collection_name=default_collection_name,
+        points=rest.Batch.construct(
+            ids=[random.randint(0, 100) for _ in range(len(lines))],
+            vectors=embeddings,
+        ),
```
kacperlukawski, line 33: Sometimes that may lead to duplicated ids in the collection. Since the test doesn't check how many points were stored, that wasn't visible. It should be instead changed to:

```python
points=rest.Batch.construct(
    ids=random.sample(range(100), len(lines)),
    vectors=embeddings,
),
```

tests/test_local_persistence.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+
+def test_local_persistence():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        ingest_data(path=tmpdir)
+        ingest_data(path=tmpdir)
```
kacperlukawski, line 40: The test doesn't check if the points were really stored in the collection.

kacperlukawski (changes requested):
The persistence layer looks ok. I just made some comments on the test code.

However, the `QdrantLocal` class should be adapted to check the presence of the collection whenever `_get_collection` is called, not on init only. Otherwise, if we have two instances of `QdrantClient`, we won't see the collections created in the second one when we call the first and the other way around.

kacperlukawski:
PR #169 contains the changes to my comments.

kacperlukawski (approved):

PR #173, opened 2023-05-17
v1.2.0

generall:


qdrant_client/qdrant_remote.py
```diff
@@ -24,6 +24,7 @@
 from qdrant_client.conversions import common_types as types
 from qdrant_client.conversions.conversion import GrpcToRest, RestToGrpc
 from qdrant_client.http import ApiClient, SyncApis
+from qdrant_client.http import models
 from qdrant_client.http import models as rest_models
```
joein, line 28: duplicated import

tests/congruence_tests/test_nested_filter.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    remote_result, _next_page = remote_client.scroll(
+        collection_name=COLLECTION_NAME,
+        scroll_filter=filter_,
+        limit=100,
+        with_payload=True,
+    )
+
+    # assert len(local_result) == len(remote_result)
```
joein, line 68: should it actually be commented out?

tests/fixtures/points.py
```diff
@@ -37,10 +39,17 @@ def generate_records(
         if random_ids:
             idx = str(uuid.uuid4())
 
+        vectors = random_vectors(vector_sizes)
+
+        if skip_vectors:
+            if random.random() > 0.8:
+                vector_to_skip = random.choice(list(vectors.keys()))
```
joein, line 46: do we want to fail with exception if someone set `vector_sizes = int` and `skip_vectors = True` ?

PR #184, opened 2023-06-08
fix: fix retry mechanism

joein:


generall (approved):

almostimplemented:
@generall @joein Can you explain the logic of this retry method? 

To me, it looks like it will fire off `max_retries` API calls every time, and if it happens to have a failure on the last one, it will raise an exception. I would expect a `return True` when `upsert_points` completes successfully, and the final `return` should be `False`. This is essentially how the code worked before this, so I don't want to put up an issue / pull request without first understanding this change.

almostimplemented:
Optimistically opened #530 , breaking out of the retry loop upon a successful upsert request.

PR #185, opened 2023-06-08
new: assert for with_vector

joein:


generall:
with_vectors was actually an example. Maybe we can do this check for any extra param?

joein:
I thought that we need to fix with_vector(s) due to the amount of questions we are getting in Discord

We can just assert that kwargs is empty

generall:
> We can just assert that kwargs is empty

that's what I am talking about

Issue #192, opened 2023-06-16
local client values count differs from remote client 


joein:
Current implementation of `payload_value_extractor.value_by_key` do not distinguish empty lists, key absence, and None values

```python
from pprint import pprint

from qdrant_client.http.models import PointStruct, ValuesCount
from qdrant_client import QdrantClient
from qdrant_client.http.models import Filter, FieldCondition, VectorParams, Distance


COLLECTION_NAME = "test"

local_client = QdrantClient(":memory:")
remote_client = QdrantClient()
local_client.recreate_collection(
    collection_name=COLLECTION_NAME,
    vectors_config=VectorParams(size=4, distance=Distance.EUCLID),
)
remote_client.recreate_collection(
    collection_name=COLLECTION_NAME,
    vectors_config=VectorParams(size=4, distance=Distance.EUCLID),
)
points = [
        PointStruct(id=1, vector=[0.05, 0.61, 0.76, 0.74], payload={"city": "Berlin"}),
        PointStruct(id=2, vector=[0.19, 0.81, 0.75, 0.11], payload={"city": ["Berlin", "London"]}),
        PointStruct(id=3, vector=[0.36, 0.55, 0.47, 0.94], payload={"city": ["Berlin", "Moscow"]}),
        PointStruct(id=4, vector=[0.18, 0.01, 0.85, 0.80], payload={"city": ["London", "Moscow"]}),
        PointStruct(id=5, vector=[0.24, 0.18, 0.22, 0.44], payload={"count": [0]}),
        PointStruct(id=6, vector=[0.35, 0.08, 0.11, 0.44]),
        PointStruct(id=7, vector=[0.14, 0.15, 0.32, 0.64], payload={"city": []}),
        PointStruct(id=8, vector=[0.11, 0.22, 0.33, 0.44], payload={"city": None}),
]
local_client.upsert(
    collection_name=COLLECTION_NAME,
    points=points
)
remote_client.upsert(
    collection_name=COLLECTION_NAME,
    points=points
)

local_points = local_client.scroll(COLLECTION_NAME, limit=10)
remote_points = remote_client.scroll(COLLECTION_NAME, limit=10)


local_search_values_count_result = local_client.search(
    collection_name=COLLECTION_NAME,
    query_vector=[0.2, 0.1, 0.9, 0.7],
    query_filter=Filter(
        must=[
            FieldCondition(key="city", values_count=ValuesCount(lt=1))
        ]
    ),
    limit=5
)
[...]

Issue #187, opened 2023-06-08
Error using Python client with local storage, search filter and missing payload


todowa:
I'm getting an error when running the following the code from [Quickstart](https://qdrant.tech/documentation/quick-start/) using the Python client and local storage. I'm using qdrant-client 1.2.0; python3.10.7.

```
from qdrant_client.http.models import PointStruct

operation_info = client.upsert(
    collection_name=COLLECTION_NAME,
    wait=True,
    points=[
        PointStruct(id=1, vector=[0.05, 0.61, 0.76, 0.74], payload={"city": "Berlin"}),
        PointStruct(id=2, vector=[0.19, 0.81, 0.75, 0.11], payload={"city": ["Berlin", "London"]}),
        PointStruct(id=3, vector=[0.36, 0.55, 0.47, 0.94], payload={"city": ["Berlin", "Moscow"]}),
        PointStruct(id=4, vector=[0.18, 0.01, 0.85, 0.80], payload={"city": ["London", "Moscow"]}),
        PointStruct(id=5, vector=[0.24, 0.18, 0.22, 0.44], payload={"count": [0]}),
        PointStruct(id=6, vector=[0.35, 0.08, 0.11, 0.44]),
    ]
)
```

Then later:
```
from qdrant_client.http.models import Filter, FieldCondition, MatchValue
search_result = client.search(
    collection_name="test_collection",
    query_vector=[0.2, 0.1, 0.9, 0.7],
    query_filter=Filter(
        must=[
            FieldCondition(
                key="city",
                match=MatchValue(value="London")
            )
        ]
    ),
    limit=3
)
```

The error is:

```
File ~/projects/.../.venv/lib/python3.10/site-packages/qdrant_client/qdrant_client.py:283, in QdrantClient.search(self, collection_name, query_vector, query_filter, search_params, limit, offset, with_payload, with_vectors, score_threshold, append_payload, consistency, **kwargs)
    199 def search(
    200     self,
    201     collection_name: str,
   (...)
    217     **kwargs: Any,
...
---> 29     if k not in data:
     30         return
     31     value = data[k]

TypeError: argument of type 'NoneType' is not iterable
```

The error seems to be because neither `qdrant_client.local.payload_filters.calculate_payload_mask` nor `qdrant_client.local.payload_value_extractors.value_by_key` can handle missing (`None`) payloads, yet this is the case for `PointStruct` with `id=6`.

[...]

PR #199, opened 2023-07-04
[WIP] Enable testing on macos

kacperlukawski:
We were not launching the tests on platforms different than Ubuntu. This PR enables testing on MacOS.
There are still some issues with Docker on Windows, so it wasn't enabled.


generall (approved):

joein:
Is this test from #187 ?

The behaviour was fixed in #191 
However not everything was fixed - #192 

So I think this test should not be here and should not be in a separate file as well. Instead, it should be added to  `tests.congruence_tests.test_updates.py::test_upsert`

Issue #175, opened 2023-05-23
TypeError: issubclass() arg 1 must be a class


gaurav274:
`import qdrant_client` does not work for `python3.8` and `python3.9`.

joein:
Hello @gaurav274 
Could you please provide qdrant-client version and full log?

gaurav274:
> Hello @gaurav274 Could you please provide qdrant-client version and full log?

Yes, @joein 

```
Name: qdrant-client
Version: 1.1.4
Summary: Client library for the Qdrant vector search engine
Home-page: https://github.com/qdrant/qdrant-client
Author: Andrey Vasnetsov
Author-email: andrey@qdrant.tech
License: 
Location: /test38/lib/python3.8/site-packages
Requires: grpcio, grpcio-tools, httpx, numpy, pydantic, typing-extensions, urllib3
```

Logs:

```In [1]: import qdrant_client
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 import qdrant_client

File /test38/lib/python3.8/site-packages/qdrant_client/__init__.py:1
----> 1 from .qdrant_client import QdrantClient as QdrantClient

File /test38/lib/python3.8/site-packages/qdrant_client/qdrant_client.py:4
      1 from typing import Any, Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Union
      3 from qdrant_client import grpc as grpc
----> 4 from qdrant_client.client_base import QdrantBase
      5 from qdrant_client.conversions import common_types as types
      6 from qdrant_client.http import ApiClient, SyncApis

File /test38/lib/python3.8/site-packages/qdrant_client/client_base.py:3
      1 from typing import Any, Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Union
----> 3 from qdrant_client.conversions import common_types as types
      4 from qdrant_client.http import models
      7 class QdrantBase:

File /test38/lib/python3.8/site-packages/qdrant_client/conversions/common_types.py:13
     10 from typing import List, Union
     12 from qdrant_client import grpc as grpc
---> 13 from qdrant_client.http import models as rest
     15 Filter = Union[rest.Filter, grpc.Filter]
     16 SearchParams = Union[rest.SearchParams, grpc.SearchParams]

File /test38/lib/python3.8/site-packages/qdrant_client/http/__init__.py:4
      1 import inspect
      3 from pydantic import BaseModel
[...]

PR #203, opened 2023-07-08
new: update pydantic, unfreeze typing-extensions

joein:
We had an issue #175 

It seemed like it was due to a problem with `typing-extensions` version, but actually it was `pydantic`'s incorrect behaviour.
As it stated [here](https://github.com/pydantic/pydantic/pull/5826).

`pydantic<1.10.8` does not work with `typing-extensions>=4.6.0`, however `pydantic>=1.10.8` works with `typing-extensions>=4.2.0`

I think it is not right to fix `typing-extensions>=4.0.0,<=4.6.0` and suggest to update pydantic instead.

Also `pydantic>=2.0.0` has requires `typing-extensions>=4.6.1`


cmaycumber:
Same would absolutely love to see this merged!

joein:
can be closed in favour of #214

generall (approved):

PR #210, opened 2023-07-12
Add fastembed integration

NirantK:
This PR adds two new functions `add` and `query` and a new return object type `QueryResponse`. This makes it a lot easier for folks coming from NLP and looking to use SoTA Embedding (beating OpenAI) but several times faster. 

Improvements:
1. Configurable ONNX runtime — allows users to use GPU, Mac Metal M1/M2, CPU and more runtimes for creating batch embedding at insertion time. This is done via the ONNX Runtime and we default to the CPU Runtime
1. Quantized model using `optimum` — this makes the model ~2x faster compared to the PyTorch runtime on CPU

### DIFF
--- a/qdrant_client/qdrant_fastembed.py
+++ b/qdrant_client/qdrant_fastembed.py
@@ -20,7 +20,7 @@ SUPPORTED_EMBEDDING_MODELS: Dict[str, Tuple[int, models.Distance]] = {
 }
 
 
-class QueryResponse(BaseModel, extra="forbid"):
+class QueryResponse(BaseModel, extra="forbid"):  # type: ignore
     id: Union[str, int]
     embedding: Optional[List[float]]
     metadata: Dict[str, Any]
@@ -212,7 +212,7 @@ class QdrantFastembedMixin(QdrantBase):
 
         assert (
             embeddings_size == vector_params.size
-        ), f"Embedding size mismatch: {embeddings_size} != {collection_info.config.params.vectors.size}"
+        ), f"Embedding size mismatch: {embeddings_size} != {vector_params.size}"
 
         assert (
             distance == vector_params.distance


qdrant_client/qdrant_client.py
```diff
@@ -73,6 +73,19 @@ def __init__(
         path: Optional[str] = None,
         **kwargs: Any,
     ):
+        # Check if fastvector is installed
+        try:
+            from fastvector.qdrant_mixin import QdrantClientMixin
```
generall, line 78: If you import `fastvector`, then `fastvectors` should not depend on qdrant-client here https://github.com/NirantK/fastvector/blob/main/pyproject.toml#L15

Otherwise it is circular dependency

NirantK: Great catch! I've removed the `qdrant-client` dependency from `fastvectors`

qdrant_client/qdrant_client.py
```diff
@@ -72,6 +72,18 @@ def __init__(
         path: Optional[str] = None,
         **kwargs: Any,
     ):
+        # Check if fastvector is installed
+        try:
+            from fastembed.qdrant_mixin import QdrantClientMixin
+
+            # If it is, add the mixin methods to this instance
+            for name, method in QdrantClientMixin.__dict__.items():
```
generall, line 80: dynamic assignment of methods might be problematic for IDE to provide completions and docs.
Might it be possible to define methods within QdrantClient, while keeping the implementations in the separate package?

generall: Those placeholder methods could have an additional instructions for the user, if the `fastembed` is not installed.

generall:
Overall it would be nice to have tests for this integration

qdrant_client/qdrant_client.py
NirantK: We'd proposed the dynamic assignment because it keeps the API design/definition in `fastembed`, instead of splitting it across the client and fastembed. I agree that this means certain IDEs can trip and not register the functions properly.

We can definitely implement the functions within the client. We had agreed upon to avoid making the client bulkier in our initial conversations. Doing that would also mean, we no longer need the `fastembed` dependency at all

What do you recommend?

NirantK: Do you recommend adding 1-4 placeholder functions to the Qdrant client with the same definitions which in turn call the `fastembed` package and check for the import error?

generall: > Do you recommend adding 1-4 placeholder functions to the Qdrant client with the same definitions which in turn call the fastembed package and check for the import error?

Yeah, I think i would be enough. The separate package is still useful, cause it isolates heavy dependencies.

As I understood, the idea it to only have minimal additional interface for update and search. All other APIs do not require embedding inference.

Additional benefit of having it as a separate package might be interoperability. So that we define mixin base class in Qdrant Client and implement it in separate packages. In this case some other project might implement the same interface and introduce other embeddings support without pushing into our repo

qdrant_client/qdrant_client.py
```diff
@@ -1699,3 +1766,5 @@ def get_locks(self, **kwargs: Any) -> types.LocksOption:
         assert len(kwargs) == 0, f"Unknown arguments: {list(kwargs.keys())}"
 
         return self._client.get_locks(**kwargs)
+        return self._client.get_locks(**kwargs)
```
generall, line 1769: Typo?

NirantK: Arrgh. Yes.

tests/test_qdrant_client.py
```diff
@@ -1105,10 +1105,62 @@ def test_client_close():
         local_client_persist_2.close()
     # endregion local
 
+@pytest.fixture(autouse=True)
+def fastembed_setup(request):
+    if request.node.get_closest_marker('fastembed'):
+        os.system('pip install fastembed')
+    elif request.node.get_closest_marker('no_fastembed'):
+        os.system('pip uninstall -y fastembed')
```
joein, line 1113: Not sure whether i would like to install dependencies inside python tests via `os.system`
Is not it better to delegate package installation to a bash script and run tests twice with a corresponding marker? 
Imho, might be more flexible during the development process and a bash script is a more obvious place to install dependencies

NirantK: No longer doing this

joein:
haven't yet looked deep into the PR, but all the packages inside `qdrant_client.http` are auto-generated and should not be modified manually

if it is actually required to modify them, https://github.com/qdrant/pydantic_openapi_v3 should be upgraded first

NirantK:
@joein for return types which are exclusive to the Python client for now — is the OpenAPI where the changes to be made?

Since it auto-generates from the Rust models directly, would not prefer doing that. Alternate proposal is to keep this in the Python client itself

generall:
Hi @NirantK, I made some changes into PR:

- Moved fastembed function into middleware, so it is easier to navigate
- Dropped support for 3.7 python, as fastembed requires >3.7. If it is not actually a case, please change that in fastembed manifest and revert my changes regarding it. But we wanted to drop 3.7 sooner or later either way
- Made interface a bit more compatible and consistent with existing functions. I hope it didn't compromise the "pythonistic" approach too much

I am going to suggest some changes in `fastembed` as well

qdrant_client/qdrant_fastembed.py
```diff
@@ -0,0 +1,287 @@
...
+            query_vector=query_vector,
+            query_filter=query_filter,
+            search_params=search_params,
+            limit=limit,
+            offset=offset,
+            with_payload=with_payload,
+            with_vectors=with_vectors,
+            score_threshold=score_threshold,
+            **kwargs,
+        )
```
NirantK, line 243: A major point of API improvement is a simpler Response type, this is why I created `QueryResponse` and simplified the `ScoredPoint` and discarded a lot of information. 

Would it possible to retain that here?

qdrant_client/qdrant_fastembed.py
```diff
@@ -0,0 +1,287 @@
...
+        collection_name: str,
+        query_text: str,
+        query_filter: Optional[models.Filter] = None,
+        search_params: Optional[models.SearchParams] = None,
+        limit: int = 10,
+        offset: int = 0,
+        with_payload: Union[bool, Sequence[str], models.PayloadSelector] = True,
+        with_vectors: Union[bool, Sequence[str]] = True,
+        score_threshold: Optional[float] = None,
+        embedding_model: str = DEFAULT_EMBEDDING_MODEL,
```
NirantK, line 181: The `query` endpoint should NOT accept `embedding_model`!

This can lead to oversight e.g. folks trying to use different models for `add` and `query`. This means we've to persist the state across multiple client sessions. One way to do so would be to have that as part of the payload itself perhaps?

qdrant_client/qdrant_fastembed.py
```diff
@@ -0,0 +1,287 @@
...
+    def query(
+        self,
+        collection_name: str,
+        query_text: str,
+        query_filter: Optional[models.Filter] = None,
+        search_params: Optional[models.SearchParams] = None,
+        limit: int = 10,
+        offset: int = 0,
+        with_payload: Union[bool, Sequence[str], models.PayloadSelector] = True,
+        with_vectors: Union[bool, Sequence[str]] = True,
```
NirantK, line 179: Do these added params makes the function more consistent with the client at the cost of simplicity? 

We'd want to have only `collection_name`, `limit` and `query_text` ideally! 

Anyone who cares about these params enough to read the docs and understand those — should use the APIs which our client already has!

qdrant_client/qdrant_fastembed.py
```diff
@@ -0,0 +1,287 @@
...
+        score_threshold: Optional[float] = None,
+        embedding_model: str = DEFAULT_EMBEDDING_MODEL,
+        **kwargs: Any,
+    ) -> List[List[types.ScoredPoint]]:
```
NirantK, line 258: The most notable difference is that this accepts a `List[str]` as queries — this is not worth learning for most MLE! 

Can we make a call from the `query` endpoint to support this behaviour instead? Thereby removing the need for engineer to learn this — and keeping it as a separate function at the same time?

qdrant_client/qdrant_fastembed.py
generall: Could you please elaborate how plain dict it better than a type?
Previous version converted `QueryResponse` into `dict` before returning it, in fact  `QueryResponse` was only used internally and gave no info to the user about the response structure.

If you propose to replace return type `ScoredPoint -> QueryResponse` that I can agree on

qdrant_client/qdrant_fastembed.py
generall: We can hide those under `**kwargs`

qdrant_client/qdrant_fastembed.py
```diff
@@ -73,14 +103,38 @@ def _embed_documents(
                 for vector in vector_batch.tolist():
                     yield vector
 
+    def _get_vector_field_name(self) -> str:
+        model_name = self.embedding_model_name.split("/")[-1].lower()
+        return f"text-{model_name}"
+
+    def _scored_points_to_query_responses(
```
NirantK, line 110: Yaay

NirantK (commented):
Looks a lot cleaner to me! Good to go from my end after the CI clears

generall (approved):

NirantK:
@generall merge when you're satisfied?

PR #211, opened 2023-07-14
Fix typing extensions version issue for python 3.8 and 3.9

ehsanmok:
Have issues with poetry complaining about the version collision with other libraries. This seems to fix the typing version issue though haven't tested it locally as I'm using python 3.10 and relying on CI here and [this changelog](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-463-june-1-2023).


joein:
hey, there is already a pull request for it #203

joein:
#214 contains required updates, it has been already merged into master, release should be soon

Issue #212, opened 2023-07-14
Qdrant local mode does not normalize vectors


kacperlukawski:
Qdrant server normalizes vectors when Cosine distance is used. However, this is not done when Qdrant local in-memory mode is used. That might not be clear for a locally created application pointing to the Qdrant server in production.

# How to reproduce?

```python
import qdrant_client

client = qdrant_client.QdrantClient(":memory:")
client.create_collection("test", vectors_config={"size": 10, "distance": "Cosine"})
client.upsert("test", points=[models.PointStruct(id=1, vector=[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0])])
```

Even though the point was stored in a collection using Cosine distance, it was not normalized.

```python
client.scroll("test", with_vectors=True)
# out: ([Record(id=1, payload={}, vector=[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0])], None)
```

PR #213, opened 2023-07-14
Fix cosine normalization

kacperlukawski:
This PR enables the normalization of vectors if cosine distance is used as reported in #212. It also enables vector comparison in tests, as the previous scroll call did not return vectors. In addition to that, the vectors are compared with `atol=1e-3` as random vectors are rounded upon generation: https://github.com/qdrant/qdrant-client/blob/bc08733e77fa3c2860177c90247ddcc8492a34a9/tests/fixtures/points.py#L15

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -554,6 +554,9 @@ class LocalCollection:
         for vector_name, named_vectors in self.vectors.items():
             vector = vectors.get(vector_name)
             if vector is not None:
+                params = self.get_vector_params(vector_name)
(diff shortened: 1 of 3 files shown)


joein (approved):

poutyface:
eps is needed if all vectors are zero

joein:
fair point

but why would anyone like to store zeroed vectors?

joein:
#236

poutyface:
That's true. I found it when I was inputting the raw sensor data. But  it might happen when you want to use high dimensional learned latent vectors (Black  images, or vq-vae encoder vector)

Issue #179, opened 2023-06-01
TypeError when connecting via insecure GRPC and with API key 


ArthurMelin:
Hello,
The grpc library throws the following exception when attempting to connect via insecure GRPC and with an API key:
```
Python 3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from qdrant_client import QdrantClient
>>> client = QdrantClient(host="qdrant", https=False, prefer_grpc=True, api_key="foobar")
/usr/local/lib/python3.11/site-packages/qdrant_client/qdrant_remote.py:104: UserWarning: Api key is used with unsecure connection.
  warnings.warn("Api key is used with unsecure connection.")
>>> client.get_collections()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/qdrant_client/qdrant_client.py", line 1130, in get_collections
    return self._client.get_collections(**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/qdrant_client/qdrant_remote.py", line 1489, in get_collections
    response = self.grpc_collections.List(
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/qdrant_client/qdrant_remote.py", line 217, in grpc_collections
    self._init_grpc_collections_client()
  File "/usr/local/lib/python3.11/site-packages/qdrant_client/qdrant_remote.py", line 159, in _init_grpc_collections_client
    self._grpc_channel = get_channel(
                         ^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/qdrant_client/connection.py", line 214, in get_channel
    channel = grpc.insecure_channel(f"{host}:{port}", metadata, options)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/grpc/__init__.py", line 1964, in insecure_channel
    return _channel.Channel(target, () if options is None else options, None,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #216, opened 2023-07-17
Fix incorrect args for grpc.insecure_channel()

ArthurMelin:
This fixes issue #179 by removing the incorrect `metadata` argument to the `grpc.insecure_channel()` call


joein:
Sorry, we have some problems with macos CI, thanks for this PR

Issue #196, opened 2023-06-25
Docu:  module 'qdrant_client.models' has no attribute 'NestedContainer'


hopkins385:
As per docu the nested filter for the python client should be configured as following:

```python
client.scroll(
    collection_name="{collection_name}",
    scroll_filter=models.Filter(
        must=[
            models.NestedContainer(
                nested=models.NestedCondition(
                    key="diet",
                    filter=(
                        must=[
                            models.FieldCondition(
                                key="food",
                                match=models.MatchValue(value="meat")
                            ),
                            models.FieldCondition(
                                key="likes",
                                match=models.MatchValue(value=True)
                            ),
                        ]
                    )
                )
            )
        ],
    ),
)

````

But after running into errors and looking into the code it appears to be this seems to be the valid syntax:

```python
   filter=models.Filter(
        must=[
            models.NestedCondition(
                nested=models.Nested(
                    key="diet",
                    filter=(
                        must=[
                            models.FieldCondition(
                                key="food",
                                match=models.MatchValue(value="meat")
                            ),
                            models.FieldCondition(
                                key="likes",
                                match=models.MatchValue(value=True)
                            ),
                        ]
                    )
                )
            )
        ],
    ),
```

Or am I missing something?

joein:
Hi @hopkins385 

Yes, you are right, it is a typo in docs, thanks for pointing it out! We will update it soon

If you will have any other problems with filters feel free to reach us out or take a look at tests e.g. [nested filter example](https://github.com/qdrant/qdrant-client/blob/master/tests/fixtures/filters.py#L80)

joein:
Also it should be `filter=models.Filter(must=[models.FieldCondition....`, and not just `filter=(must=...)`

hopkins385:
cool. happy I was able to help. 
Maybe someone can explain shortly which approach would be more suitable for my usecase.
Scenario: In a collection are multiple "documents" (langchain) clustered by custom metadata field "media_id". 
Problem: I want to filter the documents to say maybe just 2 out of 5 by passing an array of media_ids.

Which approach would be better?
Sidenote: For me its not clear if the database is first looking for "similar" vectors and after that the result is filtered, 
or if first the filter is applied and after that "similar" vectors are located

Approach 1:
```python
def get_qdrant_documents(query: str, collection_name: str, media_ids: List[str] | None) -> List[Document]:
  
# ...

    filtr = models.Filter(
        must=[]
    )

    if media_ids is not None:
[...]

PR #233, opened 2023-07-27
Change `top` kwarg to `limit` as per code

prrao87:
The `top` keyword argument [is now](https://github.com/qdrant/qdrant-client/blob/ba8063ba897007886393ca403885be1faf8b21c5/qdrant_client/client_base.py#L27) `limit` in v1.3.x. The docs didn't reflect that, so this has been fixed.


joein:
Thanks for pointing it out :)

joein:
@prrao87 we are currently don't use this documentation, current documentation is available on https://qdrant.tech/documentation/concepts/
However we are going to have a new documentation according to @davidmyriel in #196

generall (approved):

PR #238, opened 2023-07-31
V1.4.0

generall:


tests/congruence_tests/test_collections.py
```diff
@@ -42,3 +45,93 @@ def test_get_collection():
...
+            ),
+        ),
+    )
+
+    local_collection_info = local_client.get_collection(COLLECTION_NAME)
+    remote_collection_info = remote_client.get_collection(COLLECTION_NAME)
+
+    assert local_collection_info.config.params.vectors["text"].hnsw_config is None
+    assert local_collection_info.config.params.vectors["text"].quantization_config is None
+    assert local_collection_info.config.hnsw_config.ef_construct == 123
```
timvisee, line 126: In the local client `ef_construct` is set to 123 in an `update_collection` request, but it remains 100 which was the default.

Is this a problem or intended behavior for the local client?

The remote client works as expected.

timvisee (commented):
I added a simple test for the new update collection parameters feature.

I did find a potential problem with the local client, please see the comment below.

tests/congruence_tests/test_collections.py
generall: collection params doesn't have an effect in local client

timvisee (approved):
Good to see the tests succeed now :raised_hands: 

Take my :heavy_check_mark: with a grain of salt, I'm no Python wizard.

PR #243, opened 2023-08-05
Migration tool

joein:


joein:
review comments:
create a method in the client to display it in auto-complete

joein:
Not sure about `Self` type-hint in migrate method though

generall (approved):

PR #253, opened 2023-08-11
Allow a user to specify the grpc_host

graelo:
Hi,

In the current state, a user cannot use qdrant-client if qdrant is behind a reverse proxy: the rest endpoint and the gRPC endpoints have different server names and potentially different certificates, and qdrant-client only offers one host parameter.

This PR brings the ability to specify the gRPC host. The newly introduced `self._grpc_host` default to `self._host`.

Usage:

```python
client = qdrant_client.QdrantClient(
    port=443,
    grpc_port=443,
    prefer_grpc=True,
    https=True,
    host="qdrant.domain.com",
    grpc_host="qdrant-grpc.domain.com",
    api_key=qdrant_api_key,
)
```

For completeness, here is how to configure the gRPC reverse proxy on NGINX:

```conf
server {
  server_name qdrant-grpc.domain.com;
  listen 80;

  location / {
    return 301 https://$host$request_uri;
  }

  access_log /var/log/nginx/qdrant-grpc.domain.com-access.log;
  error_log /var/log/nginx/qdrant-grpc.domain.com-error.log;
}

server {
  server_name qdrant-grpc.domain.com;
  listen 443 ssl http2;

  ssl_certificate          /path/to/chained.crt;
  ssl_certificate_key      /path/to/service.key;
  add_header               Strict-Transport-Security "max-age=31536000" always;

  # Allow any size file to be uploaded.
  # Set to a value such as 1000m; to restrict file size to a specific value
  client_max_body_size 0;
  # Disable buffering
  proxy_buffering off;
  proxy_request_buffering off;

  location / {
    grpc_pass grpc://127.0.0.1:6334;
  }

  access_log /var/log/nginx/qdrant-grpc.domain.com-access.log;
  error_log /var/log/nginx/qdrant-grpc.domain.com-error.log;
}
```

### DIFF
--- a/qdrant_client/qdrant_remote.py
+++ b/qdrant_client/qdrant_remote.py
@@ -50,6 +50,7 @@ class QdrantRemote(QdrantBase):
         prefix: Optional[str] = None,
         timeout: Optional[float] = None,
         host: Optional[str] = None,
+        grpc_host: Optional[str] = None,
         **kwargs: Any,
     ):
         self._prefer_grpc = prefer_grpc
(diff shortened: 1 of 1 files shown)


generall:
Hi @graelo, have you considered creating 2 instances of qdrant client instead?
I would prefer not to allow different hosts in the single client, as it might be potentially error-prone: in some scenarios one client might point to different qdrant instances, which might be very hard to debug.

graelo:
Hi @generall, thanks, no I did not think of that. It would be great indeed.

Could you provide an example of using only grpc?

Thanks

generall:
Here is an example of using async gRPC: https://github.com/qdrant/qdrant-client/blob/master/tests/test_async_qdrant_client.py

You can re-use the same code for sync version as well.

PR #265, opened 2023-08-23
* fix(qdrant_fastembed.py): change method name from encode to embed in QdrantFastembedMixin class

NirantK:
* fix(qdrant_fastembed.py): change method name from encode to embed


generall (approved):

NirantK:
The changes from this PR were merged as part of #266. This branch and PR are no longer needed.

PR #268, opened 2023-08-25
Congruence values count

joein:
Methods `value_counts`, `isempty`, `isnull` didn't work properly in local mode due to an approach when we flatten all input values into one array

For instance if we have the following payload: `{"nested": [{"numeric": 1, "numeric": 2, "numeric": 3}]}` and want to filter by value count of `numeric`, then local mode piles up everything into one array and counts them together (`values_count=3`), but it should count them separately (`values_count=1`)

Another problem with payload like 
`{"nested": [{"empty": [None]}, {"empty": [None]}, {"empty": [None]}`
or
`{"nones": [None, None]}`

Local mode considers such fields equal to `None`, however `[None, None] != None`.

This PR adjusts interface of `payload_value_extractor` and allows to disable flattening during an extraction of values.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


tests/congruence_tests/test_unflatten_values_methods.py
```diff
@@ -0,0 +1,236 @@
...
+        scroll_filter=filter_,
+        limit=10,
+        with_payload=True,
+    )
+    assert [record.payload for record in local_result] == [
+        record.payload for record in remote_result
+    ]
+
+
+@pytest.mark.parametrize(
```
timvisee, line 128: Maybe we should try `{}` in `is_empty` and `is_null` as well.

timvisee (approved):

generall (approved):

Issue #270, opened 2023-08-29
Bug: `update_vectors` is broken


timvisee:
It appears that the [`update_vectors()`](https://github.com/qdrant/qdrant-client/blob/fd0f52a2735ccd37875b1960d005356690a6543e/qdrant_client/qdrant_client.py#L779-L786) function in our client is broken.

It expects a `vectors=` parameter, but this should be called `points=` instead:

- gRPC uses `points`: https://github.com/qdrant/qdrant/blob/0f102a5575ac33df03f06563844198f3ea26136b/docs/grpc/docs.md#qdrant-UpdatePointVectors  
  ([this](https://github.com/qdrant/qdrant-client/blob/fd0f52a2735ccd37875b1960d005356690a6543e/qdrant_client/qdrant_remote.py#L1078) breaks)
- Documentation uses `points=`: https://qdrant.tech/documentation/concepts/points/#update-vectors

I'm afraid this call isn't covered by our testing somehow.

Originally discovered because a user in Discord was having problems with this call: https://discord.com/channels/907569970500743200/1146070954653528074

timvisee:
This is fixed in v1.5.0.

PR #275, opened 2023-08-31
add tests for vector operations

generall:
Fixes <https://github.com/qdrant/qdrant-client/issues/270>.

- Add tests for vector updates and deletes
- Fix problems with those methods
- Fix type matching
- Fix test error in case of deleted vectors (related to the new logic of vector deletion, changed in https://github.com/qdrant/qdrant/pull/2514 )

### DIFF
--- a/qdrant_client/qdrant_remote.py
+++ b/qdrant_client/qdrant_remote.py
@@ -1075,7 +1075,7 @@ class QdrantRemote(QdrantBase):
                 grpc.UpdatePointVectors(
                     collection_name=collection_name,
                     wait=wait,
-                    vectors=vectors,
+                    points=vectors,
                     ordering=ordering,
                 )
             ).result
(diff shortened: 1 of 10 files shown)


qdrant_client/qdrant_remote.py
```diff
@@ -1075,7 +1075,7 @@ def update_vectors(
                 grpc.UpdatePointVectors(
                     collection_name=collection_name,
                     wait=wait,
-                    vectors=vectors,
```
timvisee, line 1078: Our Python client now still uses `vectors`, while our gRPC/REST API uses `points`.

Our documentation also suggests to use `points`: https://qdrant.tech/documentation/concepts/points/#update-vectors

I prefer `points` because you provide a list of points (with vectors), and not a list of vectors directly.

Do we leave as is and update our documentation, or do we switch to `points`? I like the latter better.

timvisee (approved):
Thanks for the fix. Tested it locally with the snippet that previously failed as well, and that works fine now (with `vectors=`).

I do see an issue with our argument name though, please see my comment below:

qdrant_client/qdrant_remote.py
generall: let's switch

PR #280, opened 2023-09-05
sync API

generall:
- Batch update API
- shard snapshot API


timvisee:
Looks good based on a static code analysis. I'll do some local testing as well before approving.

Would it be good to add congruence tests for the new batch update to ensure both implementations work the same?

timvisee (approved):

PR #317, opened 2023-09-28
fix: add init from conversion, support init from in local mode

joein:


joein:
tests are failing because `qdrant/qdrant:dev` returns `read_fan_out` on `get_collection` request

generall (approved):

PR #319, opened 2023-09-30
Async qdrant client

joein:
It is an mvp of async qdrant client generator
It should be refactored a lot, it is a minimal code to produce a workable client


joein:
update ci to fail if async client was updated

PR #326, opened 2023-10-05
python 3.12 support

generall:


timvisee (approved):
Not very familiar with this. But looks alright as far as I can judge!

PR #329, opened 2023-10-06
To avoid sqlite3.ProgrammingError when interleaving threads

chan150:
When using multi-threads application (e.g., web server gate interface; wsgi), each thread attempt to access client object which has sqlite connection.
In this case, current version of Qdrant occurs sqlite3.ProgrammingError due to the simultaneous access.

Error messages is here:
```
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread.
```

This error should be solved when your program runs on web server application (e.g., wgsi)

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


qdrant_client/local/persistence.py
```diff
@@ -25,7 +25,7 @@ def try_migrate_to_sqlite(location: str) -> None:
     try:
         dbm_storage = dbm.open(str(dbm_path), "c")
 
-        con = sqlite3.connect(str(sql_path))
+        con = sqlite3.connect(str(sql_path), , check_same_thread=False)
```
generall, line 28: Probably a typo?

chan150: Yes, it is a typo. 
comma with a blank should be removed.

I missed a commit before merging.

generall:
@chan150 do you know how safe this option is in terms of data consistency? Why is it disabled by default?

chan150:
> @chan150 do you know how safe this option is in terms of data consistency? Why is it disabled by default?

This option is safe within my knowledge.
In general, a compile macro, SQLITE_THREADSAFE, is set to 1 by default, and it make all transactions serialized.

pieroit:
We also made some tests with concurrent websocket conversations, all of them writing in the file-based db. All is right!

+1 for merge

imartinez:
@generall I've tested the new version 1.6.5 containing the changes in https://github.com/qdrant/qdrant-client/pull/369
Unfortunately it still doesn't work on OSX

Looking at the code in that merged branch, multi-thread is still disabled on OSX because check_same_thread is passed as True following the THREADSAFE variable.

I'd suggest to enable it by default (as proposed in this PR) or to make it optional by exposing it in the constructor (up to the client).

Note: This is preventing adoption in PrivateGPT project local setup.

generall:
> make it optional by exposing it in the constructor (up to the client).

I think we can do this. I am still not sure enough about safety to enable it by default

generall:
@imartinez pls check v1.6.6

imartinez:
> @imartinez pls check v1.6.6

Hey @generall, thanks for the solution, I agree it is the right approach. 
I'm afraid it still doesn't fully work. The reason, I think, is that you forgot to pass the new param to the LocalClient here:

https://github.com/qdrant/qdrant-client/blob/f1faca44f3c78160a8500e4db8cc82f83c187aa9/qdrant_client/local/qdrant_local.py#L75

Let me know if I can help.

generall:
@imartinez, you are right, missed that, thanks! Will fix in a sec

generall:
@imartinez v1.6.7 :pray:

generall:
v1.6.8

imartinez:
@generall still fails. I've checked the commit on 1.6.8 and I don't fully get it: instead of passing the new parameter in qdrant_local:L75 (as I pointed in my prev comment), you've removed it from async_qdrant_local. Why? As far as I understand, you should pass the new param in in both cases.  

https://github.com/qdrant/qdrant-client/commit/a3748acf46f886586c2fb49c6b948c4deb69a3e6

generall:
@imartinez I am sorry, I am just so un-focused today =(

generall:
I hope now it is fixed

imartinez:
It now works with 1.69. Thanks @generall. This PR can be closed because the suggestion here is already implemented.

joein:
@chan150 thanks for the initial proposal, we've added support for this functionality in 1.6.9

PR #356, opened 2023-10-26
Change to Property

NirantK:
- Removes `()` from notes in the grpc_points and other Properties


qdrant_client/async_qdrant_client.py
joein, line 1: i am not sure how you change these files, but you need to generate them via ./tools/generate_async_client.sh after sync client versions have been changed

NirantK:
Will include these changes as part of the formatting fixes

PR #359, opened 2023-10-30
Rewrite Docstrings + Change Netlify build to branch from Pip

NirantK:


NirantK:
BEFORE: 
image

AFTER: 
image

tools/generate_docs_netlify.sh
```diff
@@ -7,10 +7,13 @@ cd "$(dirname "$0")/../"
 
 brew install pandoc
 
-pip install qdrant-client
+curl -sSL https://install.python-poetry.org | python3 -
+export PATH="/opt/buildhome/.local/bin:$PATH"
+poetry install
 
 pip install sphinx==4.5.0
+pip install ipython
```
joein, line 15: why isn't it in pyproject.toml dev group?

NirantK: This is needed only for the docs build. Nothing else. Should we include that in the dev group? If yes, will add this, Sphinx, pygments and other docs dependencies there.

joein: they are already there

NirantK: Will add these to a new `docs` group in a different PR

joein (approved):

Issue #355, opened 2023-10-26
update examples in index.rst in python docs 


joein:

NirantK:
Changes made in #361

joein:
add examples for scroll, retrieve, upload_collection / upload_records (the latter two might not be in index.rst)

PR #361, opened 2023-10-30
Add New API to Docs

NirantK:
- Updated the Docs to include latest changes from dev including FastEmbed
- Updated index.rst as per #355
- Added `Async` information to index.rst

### DIFF
--- a/docs/source/qdrant_client.local.rst
+++ b/docs/source/qdrant_client.local.rst
@@ -1,0 +1,33 @@
+qdrant\_client.local package
+============================
+
+Subpackages
+-----------
+
+.. toctree::
+   :maxdepth: 4
+
+   qdrant_client.local.tests
+
+Submodules
+----------
+
+.. toctree::
+   :maxdepth: 4
+
+   qdrant_client.local.async_qdrant_local
+   qdrant_client.local.distances
+   qdrant_client.local.geo
+   qdrant_client.local.local_collection
+   qdrant_client.local.payload_filters
+   qdrant_client.local.payload_value_extractor
+   qdrant_client.local.persistence
+   qdrant_client.local.qdrant_local
+
+Module contents
+---------------
+
+.. automodule:: qdrant_client.local
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/qdrant_client.local.tests.rst
+++ b/docs/source/qdrant_client.local.tests.rst
@@ -1,0 +1,18 @@
+qdrant\_client.local.tests package
+==================================
+
+Submodules
+----------
+
+.. toctree::
+   :maxdepth: 4
+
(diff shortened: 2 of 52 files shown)


NirantK:
Fixes #355

docs/source/qdrant_client.local.rst
```diff
@@ -0,0 +1,33 @@
+qdrant\_client.local package
+============================
+
+Subpackages
+-----------
+
+.. toctree::
+   :maxdepth: 4
+
+   qdrant_client.local.tests
```
joein, line 10: not sure whether it should be here

docs/source/qdrant_client.local.tests.rst
```diff
@@ -0,0 +1,18 @@
+qdrant\_client.local.tests package
```
joein, line 1: not sure if this file should exist

docs/source/qdrant_client.local.tests.test_payload_filters.rst
```diff
@@ -0,0 +1,7 @@
+qdrant\_client.local.tests.test\_payload\_filters module
```
joein, line 1: not sure if this file should exist

qdrant_client/qdrant_client.py
```diff
@@ -23,15 +23,14 @@ class QdrantClient(QdrantFastembedMixin):
         This module methods are wrappers around generated client code for gRPC and REST methods.
         If you need lower-level access to generated clients, use following properties:
 
-        - :meth:`QdrantClient.grpc_points`
-        - :meth:`QdrantClient.grpc_collections`
-        - :meth:`QdrantClient.rest`
+        - :py:attr:`QdrantClient.grpc_points`
```
joein, line 26: I am wondering if QdrantClient can be replaced with some reference to the class itself? (not sure if it is possible in sphinx)

I am worried because AsyncQdrantClient has links to the properties of QdrantClient :( (and I started thinking about the ways to handle it)

docs/source/qdrant_client.local.rst
NirantK: This set of files is autogenerated. We don't want to delete it, but this is not rendered in the HTML

docs/source/qdrant_client.local.tests.rst
NirantK: This set of files is autogenerated. We don't want to delete it, but this is not rendered in the HTML

docs/source/qdrant_client.local.tests.test_payload_filters.rst
NirantK: This set of files is autogenerated. We don't want to delete it, but this is not rendered in the HTML

joein (approved):

Issue #375, opened 2023-11-13
Inconsistent python versions for grpc and async client generators


coszio:
In short, grpc client generator needs python 3.10, while async client generator should use 3.11

- `grpc` generator is capped at 3.10 because of [this dependency](https://github.com/qdrant/qdrant-client/blob/d6100614fd2b8413781763d57013f0ee376741e1/tools/DEVELOPMENT.md?plain=1#L19-L20) (`grpcio == 1.48.2`)

- `async` generator handles parentheses differently in 3.11 than in previous versions. See [CI failure](https://github.com/qdrant/qdrant-client/actions/runs/6850151448/job/18623753629?pr=368) when generated from python 3.10

agourlay:
Thanks for documenting this issue, I am facing it as well.

Do you happen to have a script that create a dedicated python env. with the right dependencies to generate the async. client client?

PR #387, opened 2023-11-27
new: switch async generator version to 3.10.x

joein:
addresses #375 

* [x] Does your submission pass tests?
* [x] Have you installed `pre-commit` with `pip3 install pre-commit` and set up hooks with `pre-commit install`?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "qdrant-client"
-version = "1.6.9"
+version = "1.4.0"
 description = "Client library for the Qdrant vector search engine"
 authors = ["Andrey Vasnetsov <andrey@qdrant.tech>"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "qdrant-client"
-version = "1.4.0"
+version = "1.6.9"
```
generall, line 3: please exclude it from the PR, otherwise there will be a conflict on merging dev -> master

agourlay:
quick sanity question :)

Why did we use Python 3.11 in first place for the async. generation?

joein:
I don't think there was a particular reason, probably because it's just newer and doesn't introduce redundant parentheses.

However, the final decision has been made by @generall , so I might be wrong here.

Issue #388, opened 2023-11-27
local mode accepts nans in payload, while remote mode does not


joein:

joein:
should be fixed as soon as #397 released

PR #397, opened 2023-12-08
fix: forbid nan in payload in local mode to restore congruence

joein:
#388


generall:
one extra serialization is too expensive

joein:
it is for local mode only

generall (approved):

Issue #415, opened 2024-01-01
Unable to set cache_dir or max_threads when using client.add(...)


praveen-palanisamy:
The "native" integration between the `qdrant-client` and [`fastembed`](https://github.com/qdrant/fastembed) packages allows for a simple and neat abstraction via the `client.add(...)` method to automatically create/update/insert `collection` and `Points` using `fastembed`.
But, I am unable to set the `cache_dir` used for the (ONNX) embedding model storage or modify the ONNX Runtime's `intra_op_num_threads` session option via this native integration.
A workaround is to use `fastembed`'s API without using the `qdrant_client.add(...)` method but it defeats the purpose of having this simpler, native API support.

joein:
Should be available in the next version of qdrant-client, thanks for the contribution!

PR #360, opened 2023-10-30
Allow specifying cache dir for fastembed

KaQuMiQ:
This commit adds ability to specify cache dir for fastembed when setting up the client embedding model. This allows managing models data locally and ensure proper/altered paths on docker when needed.

### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you installed `pre-commit` with `pip3 install pre-commit` and set up hooks with `pre-commit install`?

PR #416, opened 2024-01-01
Allow fastembed embedding model params configuration in parity with fastembed APIs

praveen-palanisamy:
This PR allows the user to configure the following parameters for the `fastembed` Embeddings via the `qdrant_client.set_model_params(...)` method:
```
max_length (int, optional): The maximum number of tokens.
cache_dir (str, optional): The path to the cache directory.
threads (int, optional): The number of threads single onnxruntime session can use.
```
in addition to the `embedding_model_name` parameter.  This brings parity with the init args of `fastembed`'s `Embedding` classes: `DefaultEmbedding`, `FlagEmbedding` and `JinaEmbedding`. This would allow a clean migration for existing `fastembed` users and switch to the simpler `qdrant_client.add(...)` API.
- [x] Added tests for the `qdrant_client.set_model_params(...)` method.

- [x] This fixes the issue described in #415 .

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/qdrant_fastembed.py
+++ b/qdrant_client/qdrant_fastembed.py
@@ -40,6 +40,7 @@ class QdrantFastembedMixin(QdrantBase):
         max_length: int = 512,
         cache_dir: Optional[str] = None,
         threads: Optional[int] = None,
+        **kwargs: Any,
     ) -> None:
         """
         Set embedding model to use for encoding documents and queries.
--- a/qdrant_client/qdrant_fastembed.py
+++ b/qdrant_client/qdrant_fastembed.py
@@ -34,7 +34,7 @@ class QdrantFastembedMixin(QdrantBase):
         self.embedding_model_name = self.DEFAULT_EMBEDDING_MODEL
         super().__init__(**kwargs)
 
-    def set_model_params(
+    def set_model(
         self,
         embedding_model_name: str,
         max_length: int = 512,
--- a/qdrant_client/qdrant_fastembed.py
+++ b/qdrant_client/qdrant_fastembed.py
@@ -63,6 +63,7 @@ class QdrantFastembedMixin(QdrantBase):
             max_length=max_length,
             cache_dir=cache_dir,
             threads=threads,
+            **kwargs,
         )
         self.embedding_model_name = embedding_model_name
 
@@ -93,6 +94,7 @@ class QdrantFastembedMixin(QdrantBase):
         max_length: int = 512,
         cache_dir: Optional[str] = None,
         threads: Optional[int] = None,
+        **kwargs: Any,
     ) -> "DefaultEmbedding":  # -> Embedding: # noqa: F821
         if model_name in cls.embedding_models:
             return cls.embedding_models[model_name]
@@ -109,6 +111,7 @@ class QdrantFastembedMixin(QdrantBase):
             max_length=max_length,
             cache_dir=cache_dir,
             threads=threads,
+            **kwargs,
         )
         return cls.embedding_models[model_name]
 
--- a/qdrant_client/qdrant_fastembed.py
+++ b/qdrant_client/qdrant_fastembed.py
@@ -38,8 +38,8 @@ class QdrantFastembedMixin(QdrantBase):
         self,
         embedding_model_name: str,
         max_length: int = 512,
-        cache_dir: str = None,
-        threads: int = None,
+        cache_dir: Optional[str] = None,
+        threads: Optional[int] = None,
     ) -> None:
         """
         Set embedding model to use for encoding documents and queries.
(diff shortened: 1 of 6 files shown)


DiTo97:
@praveen-palanisamy, when will it be merged?

joein:
Hi @praveen-palanisamy , thanks for your contribution!

Hi, @DiTo97 I'll try to review it today and see whether we can merge it as is or not

praveen-palanisamy:
@DiTo97 : Thanks for your interest and the ping. I am waiting for the review/approval before this can be merged. I'll follow-it up.

@joein : Do you have any feedback/review comments to finalize & merge this PR?

joein:
Hi @praveen-palanisamy 

Yeap, I have some comments

Firstly, I'd like to avoid changing names of the existing API

Secondly, I would like to have `**kwargs` in the method's signature

Tbh, I am not even sure at the moment if we actually need to have `max_length`, `cache_dir` and `threads` as explicit params in the signature, or not. I need to get some context from the team. 

P.S. We use python3.10 to generate async client (there is a difference with parentheses starting from python3.11, CI will fail if you generate it with python != 3.10)

joein:
Related https://github.com/qdrant/qdrant-client/pull/360

praveen-palanisamy:
@joein Thanks for your comments.

> Firstly, I'd like to avoid changing names of the existing API
1. Okay. I proposed to change the name from `set_model` to `set_model_params` because it was more appropriate since this method is not only setting the model but also the associated config params. This new name also matched with the other existing (internal) method naming: `_get_model_params`. I can revert the name change if you prefer to keep it unchanged.

> Secondly, I would like to have **kwargs in the method's signature
2. Okay. I will update the signature. I am keeping the positional args as is for now with the `**kwargs` unused. Please see response below:

>Tbh, I am not even sure at the moment if we actually need to have max_length, cache_dir and threads as explicit params in the signature, or not. I need to get some context from the team.

3. Are you suggesting to just fold them all into `**kwargs`? They are kept as positional arguments in the same order as the underlying init args of fastembed's Embedding classes: `DefaultEmbedding`, `FlagEmbedding` and `JinaEmbedding`. Keeping this parity is preferred from a user point of view. I guess the `fastembed` API signature may change as it is early in its development cycle. If we want to avoid updating this signature frequently and bring back the parity at a later stage, it makes sense to me. Please confirm the change request after your sync with the team.

>P.S. We use python3.10 to generate async client (there is a difference with parentheses starting from python3.11, CI will fail if you generate it with python != 3.10)
4. Thanks for the tip! I just noticed that the CI runs [this check](https://github.com/qdrant/qdrant-client/blob/master/tests/async-client-consistency-check.sh) only with Python `3.10.x` while the test matrix includes Python versions `>=3.8 <=3.12`. [This check](https://github.com/qdrant/qdrant-client/blob/master/tests/async-client-consistency-check.sh) passes with Python 3.11 as well. Any particular reason why Python `3.10.x` alone was chosen to run that check/test?

Thanks for the pointer to # 360. That was not enough to allow me as a user to use `qdrant_client.add(...)` with the desired ONNX session option.

qdrant_client/qdrant_fastembed.py
```diff
@@ -34,20 +34,36 @@ def __init__(self, **kwargs: Any):
...
 
         Returns:
             None
         """
-        self._get_or_init_model(model_name=embedding_model_name)
+        self._get_or_init_model(
+            model_name=embedding_model_name,
+            max_length=max_length,
+            cache_dir=cache_dir,
+            threads=threads,
```
joein, line 65: **kwargs

qdrant_client/qdrant_fastembed.py
```diff
@@ -84,7 +104,12 @@ def _get_or_init_model(
 
         cls._import_fastembed()
 
-        cls.embedding_models[model_name] = DefaultEmbedding(model_name=model_name)
+        cls.embedding_models[model_name] = DefaultEmbedding(
+            model_name=model_name,
+            max_length=max_length,
+            cache_dir=cache_dir,
+            threads=threads,
```
joein, line 111: **kwargs

qdrant_client/qdrant_fastembed.py
```diff
@@ -72,7 +88,11 @@ def _get_model_params(cls, model_name: str) -> Tuple[int, models.Distance]:
 
     @classmethod
     def _get_or_init_model(
-        cls, model_name: str
+        cls,
+        model_name: str,
+        max_length: int = 512,
+        cache_dir: Optional[str] = None,
+        threads: Optional[int] = None,
```
joein, line 95: **kwargs

joein:
1. Yes, I prefer to avoid changes to the existing interface, if there is no necessity.
2-3. I think that we can have these parameters as keyword arguments, it is not hard to hide them inside `**kwargs` later if there is a need for this.
4. As I've told, python3.10 and python3.11 generates different code (3.10 adds parentheses, 3.11 does not). We can't have this check turned on for 3.10 and 3.11 simultaneously, since only one version can pass at the same moment. 3.10 was chosen due to the problems with gRPC generator.

praveen-palanisamy:
@joein : Thanks for the comments and responses! I have resolved the remaining items and also regenerated the async APIs using Python 3.10.x.
Please let me know if any more changes are required before final CI check & merge.

joein:
@praveen-palanisamy 

Fine by me, thank you for the contribution!

joein (approved):

Issue #365, opened 2023-11-01
"'NoneType' object is not callable" on program close when using local data file


sjpritchard:
Description:

When using a local data file, the following exception is generated on program termination (with no apparent adverse effects on the data):

```
Exception ignored in: <function QdrantClient.__del__ at 0x7f0eb98f1e40>
Traceback (most recent call last):
  File "/xxxx/lib/python3.11/site-packages/qdrant_client/qdrant_client.py", line 116, in __del__
  File "/xxxx/lib/python3.11/site-packages/qdrant_client/qdrant_client.py", line 123, in close
  File "/xxxx/lib/python3.11/site-packages/qdrant_client/local/qdrant_local.py", line 57, in close
TypeError: 'NoneType' object is not callable
```

Code sample:
```
qdrant_client = QdrantClient(path="./qdrant_data")
qdrant_client.recreate_collection(
    collection_name="xxxxx",
    vectors_config=models.VectorParams(size=1024, distance=models.Distance.COSINE),
)
for idx, item in enumerate(input_data, start=1):
    print(f"Processing {idx} of {len(input_data)}")
    text_string = " ".join(item["text"])
    if not text_string:
        continue
    embedding = get_embedding(text_string)
    id = str(uuid.uuid4())
    qdrant_client.upsert(
        collection_name="simcortex", points=[models.PointStruct(id=id, vector=embedding, payload={"text": text_string})]
    )
```

System Info:
- Python 3.11
- Ubuntu 22.04 on WSL
- qdrant-client 1.6.4

generall:
could you please provide a code example of how to reproduce the issue? I am afraid, your explanation leaves a room for misinterpreting

joein:
It seems like it might be related to the usage of `portalocker` on WSL. At least that's what can be assumed from the log.

sjpritchard:
> could you please provide a code example of how to reproduce the issue? I am afraid, your explanation leaves a room for misinterpreting

I've added some example code. It's pretty simple. Create a client, and multiple upserts. I tried adding an explicit qdrant_client.close() at the end with no effect.

XXXHlown:
Has this issue been resolved? I got the similar Exception while data upsert

Code sample:
``` python
class Model(object):
    """singleton & lazy load"""

    def __init__(self) -> None:
        self._qclient = None
        self._hf_embeddings = None

    @property
    def qclient(self):
        if not self._qclient:
[...]

PR #421, opened 2024-01-04
fix: fix portalocker being deleted before closing the client

joein:
Sometimes portalocker module being deleted even before deletion of QdrantLocal

In this case when `__del__` method in QdrantClient being called, we can already don't have portalocker object and calling unlock on it will cause an error `NoneType...` https://github.com/qdrant/qdrant-client/issues/365

### DIFF
--- a/qdrant_client/local/qdrant_local.py
+++ b/qdrant_client/local/qdrant_local.py
@@ -66,7 +66,7 @@ class QdrantLocal(QdrantBase):
                 portalocker.unlock(self._flock_file)
             except (
                 TypeError
-            ):  # sometimes portalocker module can be deleted even before qdrant local instance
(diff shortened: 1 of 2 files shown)


qdrant_client/local/qdrant_local.py
```diff
@@ -62,7 +62,12 @@ def close(self, **kwargs: Any) -> None:
                     f"{list(self.collections.keys())}"
                 )
         if self._flock_file is not None:
-            portalocker.unlock(self._flock_file)
+            try:
+                portalocker.unlock(self._flock_file)
+            except (
+                TypeError
+            ):  # sometimes portalocker module can be garbage collected before QdrantLocal instance
+                pass
```
generall, line 70: please fix formatiing

generall (approved):

Issue #403, opened 2023-12-12
upload_records() calls in _migrate_collection() should use wait=True


andrew-titus:
Since the calls to upload_records ([here](https://github.com/qdrant/qdrant-client/blob/6d019e67a133bea87a96bce388f5b901cdae1287/qdrant_client/migrate/migrate.py#L119) and [here](https://github.com/qdrant/qdrant-client/blob/6d019e67a133bea87a96bce388f5b901cdae1287/qdrant_client/migrate/migrate.py#L124)) in _migrate_collection() do not use `wait=True` (which defaults to `False`), the assertion statement at the end of the function that checks that the vector counts match often fails when migrating content from a local client to a remote client. Since we are checking that they match, we should wait for any uploads to complete for consistency, so this should either be set to `wait=True` directly or at least be exposed as a kwarg for users. Thanks!

shivas1516:
I would like to take on this issue. After reviewing the code, I understand that the calls to `upload_records` in `_migrate_collection()` may not wait for uploads to complete, potentially leading to inconsistencies in vector counts. I plan to address this.
If there are any specific considerations or preferences regarding the solution, please let me know. I'll keep you updated on my progress.

Looking forward to contributing!

joein:
Thanks for highlighting this issue!

I believe the main issue with `upload_records` in the loop had already been resolved by @agourlay 

The remaining part of it with a corner case should be fixed with #428 and available with the next release, thanks @shivas1516 

Feel free to re-open the issue if the problems is reproducible in `qdrant-client 1.7.1` (should be available this week)

PR #428, opened 2024-01-08
Fix Vector Count Mismatch during Client Migration #403

shivas1516:
# Client Migration Fix

Resolve vector count mismatch during client migration. The problem was due to missing `<wait=True>` in the `<upload_records>` calls. This commit adds `<wait=True>` to ensure proper synchronization and resolves the vector count mismatch.

### All Submissions:

* [x] Contributions should target the `dev` branch. (Yes, created branch from `dev`)
* [x] Have you followed the guidelines in our Contributing document? (Yes)

### New Feature Submissions:

### Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them? (Yes, explanation included in the commit message)
* [ ] Have you successfully run tests with your changes locally?

### DIFF
--- a/qdrant_client/migrate/migrate.py
+++ b/qdrant_client/migrate/migrate.py
@@ -106,6 +106,7 @@ def _migrate_collection(
     dest_client: QdrantBase,
     collection_name: str,
     batch_size: int = 100,
+    wait: bool = True,
 ) -> None:
     """Migrate collection from source client to destination client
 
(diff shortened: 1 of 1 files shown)


qdrant_client/migrate/migrate.py
```diff
@@ -106,6 +106,7 @@ def _migrate_collection(
     dest_client: QdrantBase,
     collection_name: str,
     batch_size: int = 100,
+    wait: bool = True,
```
joein, line 109:
```suggestion
```

qdrant_client/migrate/migrate.py
```diff
@@ -116,12 +117,12 @@ def _migrate_collection(
...
     """
     records, next_offset = source_client.scroll(collection_name, limit=2, with_vectors=True)
-    dest_client.upload_records(collection_name, records)
+    dest_client.upload_records(collection_name, records, wait=True)
     while next_offset is not None:
         records, next_offset = source_client.scroll(
             collection_name, offset=next_offset, limit=batch_size, with_vectors=True
         )
-        dest_client.upload_records(collection_name, records, wait=True)
+        dest_client.upload_records(collection_name, records, wait=wait)
```
joein, line 125:
```suggestion
        dest_client.upload_records(collection_name, records, wait=True)
```

joein (commented):
I guess we may have more complaints about timeouts with these changes, but let's try it out

It also does not make a lot of sense to add `wait` parameter to the `_migrate_collection` signature, since it is not exposed to the user

And I don't want to make the signature of `migrate` more complex

Issue #431, opened 2024-01-11
Local mode does not raise an error on an attempt to write a non-existent named vector


joein:
```python

from qdrant_client import QdrantClient, models

client = QdrantClient(":memory:")
client.recreate_collection(
    "test", vectors_config=models.VectorParams(size=2, distance=models.Distance.COSINE)
)
client.upsert("test", points=[models.PointStruct(id=1, vector={"what_am_I": [0.1, 0.3]})])
```

joein:
available as of `qdrant-client==1.7.1`

PR #432, opened 2024-01-11
fix: fix upsert check in local mode

joein:
#431


joein:
tests fails due to a new field in snapshot (checksum)

joein:
tested locally on qdrant v1.7.0 - tests pass

joein:
tested locally on v1.7.0

generall (approved):

PR #433, opened 2024-01-11
fix: fix local mode loading with dense and sparse vectors

joein:


joein:
tests fails due to new fields in dev

tested locally on qdrant v1.7.0 - all green

generall (approved):

PR #436, opened 2024-01-15
fix: Add document type validation in Fastembed and update test case.

kartik-gupta-ij:
* [x] Add document type validation to Fastembed as it only supports strings.
* [x] Update test case.


joein:
Hey, I don't think we need this :(

1) Fastembed has been refactored
2) Current PR implementation reads all data into RAM (`all`), we want to preserve lazy reading when possible

Thank you

Issue #439, opened 2024-01-17
TypeError: only integer scalar arrays can be converted to a scalar index


dabasmoti:
I am trying to create a collection with sparse vectors 
I am using `mmh3.hash(token, signed=False)` as hashmap.
and I get the error above.

## Steps to Reproduce
```
from qdrant_client import QdrantClient
from qdrant_client import models

client = QdrantClient(":memory:")
COLLECTION_NAME = "bm25"

client.recreate_collection(
    collection_name=COLLECTION_NAME,
    vectors_config={},
    sparse_vectors_config={"text": models.SparseVectorParams()}
)


client.upsert(
collection_name=COLLECTION_NAME,
points=[
    models.PointStruct(
        id=1,
        payload={},  # Add any additional payload if necessary
        vector={
            "text": models.SparseVector(
        indices=[1479876206, 696618268],
        values=[0.5,0.6]
            )
        },
    )
],
)
```

agourlay:
I took the liberty to transfer the issue to the qdrant Python client repository.

joein:
Hi @dabasmoti 

We have a bug on our side.
In case you need to use local mode and don't want to wait for the release, you can sort indices and corresponding values before vector creation, e.g.:

OK:
```
SparseVector(
    indices=[0, 5],
    values=[0.3, 0.2]
)
```

ERROR:
```
SparseVector(
    indices=[5, 0],
    values=[0.2, 0.3]
)
```

joein:
available as of `qdrant-client==1.7.1`

PR #442, opened 2024-01-19
Fix sort sparse vectors

joein:
#439

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -30,7 +30,7 @@ from qdrant_client.local.persistence import CollectionPersistence
 from qdrant_client.local.sparse import (
     calculate_distance_sparse,
     empty_sparse_vector,
-    sort,
+    sort_sparse_vector,
     validate_sparse_vector,
 )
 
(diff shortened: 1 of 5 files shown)


qdrant_client/local/local_collection.py
```diff
@@ -401,7 +401,7 @@ def search(
             scores = calculate_context_scores(query_vector, vectors[: len(self.payload)], distance)
         elif isinstance(query_vector, SparseVector):
             # sparse vector query must be sorted by indices for dot product to work with persisted vectors
-            sort(query_vector)
+            query_vector = sort_sparse_vector(query_vector)
```
agourlay, line 404: That's embarrassing, I guess I thought this was a sort in place :face_with_spiral_eyes:

qdrant_client/local/sparse.py
```diff
@@ -24,18 +28,20 @@ def is_sorted(vector: SparseVector) -> bool:
...
-def sort(vector: SparseVector) -> SparseVector:
+def sort_sparse_vector(vector: SparseVector) -> SparseVector:
     if is_sorted(vector):
         return vector
 
     sorted_indices = np.argsort(vector.indices)
     return SparseVector(
-        indices=vector.indices[sorted_indices],
-        values=vector.values[sorted_indices],
+        indices=[vector.indices[i] for i in sorted_indices],
```
agourlay, line 37: Why is this change necessary?

qdrant_client/local/local_collection.py
joein: 🙈

qdrant_client/local/sparse.py
joein: `vector.indices` and `vector.values` are built-in python lists
This kind of lists does not support indexing by array of indices

To take elements from an array by several indices, this array should support such an operation (e.g. np.array)

agourlay (approved):
Thank you!

generall (approved):

Issue #441, opened 2024-01-18
Timeout parameter has different types in QdrantClient and in its methods


joein:
Context:
https://discord.com/channels/907569970500743200/1196696606012092459/1196696606012092459

joein:
available as of `qdrant-client==1.7.1`

PR #443, opened 2024-01-19
fix: align timeout type for qdrant client and its methods

joein:
#441


joein:
tests fail due to an update to the interface in qdrant dev, everything passes locally on qdrant v1.7.0

generall (approved):

LouisAuneau:
Hello @joein,

I feel this PR introduces a regression. We used to provide an `httpx.Timeout` object to our `QdrantClient`, which, although it wasn't compliant with the typings, allowed fine-grained timeout management between connect, read and write timeouts. Maybe it would be more appropriate to change the parameter type to `Union[float, httpx.Timeout, None]` and handle the types with dedicated logic in the constructor ?

joein:
Hi @LouisAuneau 

I can't say that it is a regression, since we have never supported this case 🤷‍♂️  
And this case would indeed require to have some additional logic around `timeout` in the constructor.
The concern for this use case was initially raised by one of our users on discord, since we used to have different types for `timeout` in QdrantClient and its methods.

What I suggest to do at the moment

1) Create an issue in qdrant-client, so we could think on a decision for you
2) If you need functionality from `qdrant-client==1.7.1`, then you probably need to make a fork and change it on your own
3) Otherwise, you can pin `qdrant-client==1.7.0` and wait until we resolve the issue

Btw, if you ignore our type hints in some other cases as well, feel free to open issues in the repository, maybe we are missing something and can extend our library :)

Issue #427, opened 2024-01-06
Replace types.Record in upload_records with Points


joein:
Currently, `upload_records` accept an array of `Record`s.

`Record` contains a field for a shard key. 

We don't want to allow to specify a shard key per record in `upload_records` (neither do we in `upload_collection`).

If we allow to specify a shard key for each record in `upload_records`, it will make code more complex and affect the performance.

The performance might be affected since `upsert` API requires `shard key` to be set for the whole batch, rather than per point. Thus if an auto-assembled batch contains different shard keys, we need to split it, it increases the number of requests.

Due to this, we don't want to allow to use a structure which can cause ambiguity, instead we want to use `common_types.Points` or `List[models.PointStruct]`

joein:
fastembed also uses records

generall:
The proposal is to accept both, but print deprecation warning if Record is used. Over one minor version remove record

joein:
available as of `qdrant-client==1.7.1`

PR #447, opened 2024-01-19
new: deprecate upload records, update tests, prohibit migration of co…

joein:
- Deprecate usage of `upload_records` due to the deprecation of `models.Record`. `models.Record` was deprecated due to a `shard_key` which can't be specified item-by-item in uploading processes #427 
- Update corresponding tests
- Prohibit migration of collections with custom shard keys since it is not implemented and is not a frequent use case (yet a hard one)


qdrant_client/qdrant_client.py
```diff
@@ -1817,6 +1824,55 @@ def upload_records(
...
+        return self._client.upload_points(
+            collection_name=collection_name,
+            points=points,
+            batch_size=batch_size,
+            parallel=parallel,
+            method=method,
+            max_retries=max_retries,
+            wait=wait,
+            shard_key_selector=shard_key_selector,
+            **kwargs,
```
generall, line 1873: if `kwargs` is empty, let's not pass it here

generall, line 1873:
```suggestion
```

qdrant_client/async_qdrant_client.py
```diff
@@ -1754,6 +1760,54 @@ def upload_records(
...
+        return await self._client.upload_points(
+            collection_name=collection_name,
+            points=points,
+            batch_size=batch_size,
+            parallel=parallel,
+            method=method,
+            max_retries=max_retries,
+            wait=wait,
+            shard_key_selector=shard_key_selector,
+            **kwargs,
```
generall, line 1808:
```suggestion
```

generall (approved):

Issue #452, opened 2024-01-20
Fine-grained timeout for httpx


joein:
Hello @joein,

I feel this PR introduces a regression. We used to provide an `httpx.Timeout` object to our `QdrantClient`, which, although it wasn't compliant with the typings, allowed fine-grained timeout management between connect, read and write timeouts. Maybe it would be more appropriate to change the parameter type to `Union[float, httpx.Timeout, None]` and handle the types with dedicated logic in the constructor ?

_Originally posted by @LouisAuneau in https://github.com/qdrant/qdrant-client/issues/443#issuecomment-1901308986_

joein:
@LouisAuneau unfortunately, we can't just ease the condition due to mypy complaints, we would need to do another workaround, but it won't be available in 1.7.2 :( more like 1.8.0

LouisAuneau:
Thanks for your feedback @joein. For now we set an upper bound to `1.7.0` to solve the problem.

PR #453, opened 2024-01-20
fix: replace condition to allow illegal timeout type usage

joein:
#452


joein:
we would probably need to convert timeout to int in each operation if we want this feature to be available

Issue #459, opened 2024-01-24
upload_collection can't be launched with parallel and without explicit ids


joein:
```
cl.upload_collection('test1', vectors=[[1.0, 0.2], [0.1, 0.2], [0.3, 0.4], [0.2, 0.1], [0.1, 0.2]], batch_size=2, parallel=3)
Traceback (most recent call last):
  File "/Users/x/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/Users/x/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle 'generator' object
Traceback (most recent call last):
  File "/Users/x/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/Users/x/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle 'generator' object
Traceback (most recent call last):
  File "/Users/x/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/Users/x/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle 'generator' object
```

PR #460, opened 2024-01-24
fix: fix implicit ids in upload collection with paralell > 1

joein:


joein:
#459

generall (approved):

tests/test_qdrant_client.py
```diff
@@ -257,6 +259,57 @@ def test_point_upload(prefer_grpc):
...
+    )
+    vectors = [
+        [1.0, 2.0, 3.0],
+        [4.0, 5.0, 6.0],
+        [7.0, 8.0, 9.0],
+        [10.0, 11.0, 12.0],
+        [13.0, 14.0, 15.0],
+    ]
+    payload = [{"a": 2}, {"b": 3}, {"c": 4}, {"d": 5}, {"e": 6}]
+    ids = [1, 2, 3, 4, 5]
```
coszio, line 282: We can make another test for locking in the behavior of auto-generating ids when `ids = None`

tests/test_qdrant_client.py
```diff
@@ -257,6 +259,57 @@ def test_point_upload(prefer_grpc):
...
+        vectors_config=VectorParams(size=size, distance=Distance.DOT),
+        timeout=TIMEOUT,
+    )
+    vectors = [
+        [1.0, 2.0, 3.0],
+        [4.0, 5.0, 6.0],
+        [7.0, 8.0, 9.0],
+        [10.0, 11.0, 12.0],
+        [13.0, 14.0, 15.0],
+    ]
```
coszio, line 280: Right now the behavior is to stop at the shortest iterator of any of `ids`, `vectors`, or `ids`. Is it possible to emit a warning when this happens? E.g, for when it stopped with any of those un-exhausted

coszio (commented):
I'm particularly interested in the suggested warning, otherwise it looks good to me

tests/test_qdrant_client.py
joein: It is actually already there, the first call to upload_collection does not provide ids and payload, only vectors

The second call provides all of them - vectors, ids and payload. I put the data into one place because if we change vectors, then ids and payload should also be changed

tests/test_qdrant_client.py
joein: Yeah, we can consider it as a separate issue

coszio (approved):

tests/test_qdrant_client.py
joein: I've just realised that having iterators of different length is a valid scenario, e.g. it is valid when ids iterator is infinite.

We can only check the number of ids/payloads/vectors right before making a request, however this check won't help when the smallest iterator is divisible by batch_size

PR #474, opened 2024-01-31
Fix potential edge case scoring in context search

coszio:
Follows the same fix that was made in core (qdrant/qdrant#3374) for congruence

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/local/distances.py
+++ b/qdrant_client/local/distances.py
@@ -147,6 +147,14 @@ def calculate_distance_core(
         return calculate_distance(query, vectors, distance_type)
 
 
+def fast_sigmoid(x: np.float32) -> np.float32:
+    if np.isfinite(x):
+        return x / (1.0 + abs(x))
+    else:
+        # To avoid NaNs, which gets: RuntimeWarning: invalid value encountered in scalar divide
+        return x
+
+
 def scaled_fast_sigmoid(x: np.float32) -> np.float32:
     if np.isfinite(x):
         return 0.5 * (x / (1.0 + abs(x)) + 1.0)
@@ -234,7 +242,9 @@ def calculate_context_scores(
         neg = calculate_distance_core(pair.negative, vectors, distance_type)
 
         difference = pos - neg - EPSILON
-        pair_scores = np.minimum(difference, 0.0)
+        pair_scores = np.fromiter(
+            (fast_sigmoid(xi) for xi in np.minimum(difference, 0.0)), np.float32
+        )
         overall_scores += pair_scores
 
     return overall_scores
--- a/qdrant_client/local/distances.py
+++ b/qdrant_client/local/distances.py
@@ -2,6 +2,7 @@ from enum import Enum
 from typing import List, Optional, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


qdrant_client/local/distances.py
```diff
@@ -147,6 +147,14 @@ def calculate_distance_core(
         return calculate_distance(query, vectors, distance_type)
 
 
+def fast_sigmoid(x: np.float32) -> np.float32:
+    if np.isfinite(x):
+        return x / (1.0 + abs(x))
+    else:
```
SamirPS, line 153: Small thing:

You can remove the else and directly do return X no?

SamirPS: Or https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expit.html ?

coszio: True, we can avoid the redundant else statement. 

> expit

We use the same function as in core implementation to get scoring congruence

SamirPS (approved):

qdrant_client/local/distances.py
SamirPS: Oh I see

qdrant_client/local/distances.py
```diff
@@ -147,13 +147,21 @@ def calculate_distance_core(
         return calculate_distance(query, vectors, distance_type)
 
 
+def fast_sigmoid(x: np.float32) -> np.float32:
+    if not np.isfinite(x):
```
joein, line 151: I am a bit confused by it, I know, that we have had such a check for quite some time now though..

Do we wait for a particular not finite type here?

Like -inf, +inf, nan ? (if we wait -inf, won't it be better to return a constant value?)

It seems for me that at the moment, If vector has a non-finite difference with any of context pairs, then it is excluded from result search, because it will have non-finite score in `overall_scores`

SamirPS: Yeah, I don't know if I understand well, but if x equal +inf or -inf the function return x is good or you want to return for example 2 or 3 ?

qdrant_client/local/distances.py
```diff
@@ -2,6 +2,7 @@
 from typing import List, Optional, Union
 
 import numpy as np
+from numpy.lib.ufunclike import isneginf
```
joein, line 5: you are a numpy expert

coszio: wait I need to remove that

qdrant_client/local/distances.py
```diff
@@ -148,19 +149,15 @@ def calculate_distance_core(
 
 
 def fast_sigmoid(x: np.float32) -> np.float32:
-    if not np.isfinite(x):
-        # To avoid NaNs, which gets: RuntimeWarning: invalid value encountered in scalar divide
-        return x
+    if np.isnan(x):
+        # To avoid divisions on NaNs, which gets: RuntimeWarning: invalid value encountered in scalar divide
+        return x  # NaN
```
joein: do we want to just hide the warning? :D

PR #479, opened 2024-02-03
feat: Add URL information in UnexpectedResponse

### LINKED ISSUE #456: Add information about requested URL to UnexpectedResponse exception to simplify troubleshooting
If you specify an incorrect URL for the QdrantClient, and make a call to the server such as `upsert` the request will fail with an exception.

The exception does not specify which server it tried to call, only the response.

If for example, you specify the URL for a web server that is not the cluster server (e.g. one of the nodes) you can get an exception with HTTP status 404 "not found" like this when you call `upsert`:

```
qdrant_client.http.exceptions.UnexpectedResponse: Unexpected Response: 404 (Not Found)
Raw response content:
b'404 page not found\n'
```

To make troubleshooting configuration mistakes easier, it would be nice to include the actual URL in the exception.

*Proposed Solution*

The exception comes from the `send` function in `api_client.py`.

One solution would be to add an optional `url` field to  `UnexpectedResponse` in `exceptions.py` and rewrite the `UnexpectedResponse.for_response` function  to include the URL data from the request (the request is an optional field on `response` which is used to construct the exception).

SamirPS:
This commit add the url in the UnexpectedResponse class, and modify the for_response and __str__ to display the URL.

Signed_Off_by: Samir Akarioh <akariohsamir@yahoo.com>
resolve #456

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/http/exceptions.py
+++ b/qdrant_client/http/exceptions.py
@@ -1,7 +1,7 @@
 import json
 from typing import Any, Dict, Optional
 
-from httpx import Headers, Response
+from httpx import Headers, Response, URL
 
 MAX_CONTENT = 200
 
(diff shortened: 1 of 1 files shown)


joein:
Unfortunately, we can't just change files in `http/` (or `grpc` because they are auto-generated.)

And the generator is not public

SamirPS:
oh i see, do you want me to close the PR ?

joein:
It's not necessary atm

PR #484, opened 2024-02-08
Fix gRPC conversion for sparse search batch

agourlay:
Fixes the gRPC conversion for batch search requests containing sparse vectors.

### DIFF
--- a/qdrant_client/conversions/conversion.py
+++ b/qdrant_client/conversions/conversion.py
@@ -1955,13 +1955,13 @@ class RestToGrpc:
     @classmethod
     def convert_named_vector_struct(
         cls, model: rest.NamedVectorStruct
-    ) -> Tuple[List[float], Optional[List[int]], Optional[str]]:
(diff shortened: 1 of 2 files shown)


qdrant_client/conversions/conversion.py
```diff
@@ -1955,23 +1955,26 @@ def convert_batch_vector_struct(
     @classmethod
     def convert_named_vector_struct(
         cls, model: rest.NamedVectorStruct
-    ) -> Tuple[List[float], Optional[str]]:
+    ) -> Tuple[List[float], Optional[grpc.SparseIndices], Optional[str]]:
         if isinstance(model, list):
```
joein, line 1959: I am a little bit confused by our naming...

Why do we have `NamedVectorStruct` which can be a `List[StrictFloat]` :(

Code is correct though 👍

joein (approved):

qdrant_client/conversions/conversion.py
agourlay: Anonymous named vector #legacy

Issue #490, opened 2024-02-12
Tracking issue: local mode for Qdrant v1.8


coszio:
There are some new features scheduled for Qdrant v1.8 that need to be implemented in local mode:

- [x] `order_by` in scroll (qdrant/qdrant#3521) => #491 
- [x] `min_should` filtering condition (qdrant/qdrant#3466) => no PR yet
- [x] remove deprecated `clear_payload_deprecated` field => #502
    - https://github.com/qdrant/qdrant-client/blob/66cb6dff45a98d915e572d35a46e2e6e9d9c3b0c/tests/conversions/fixtures.py#L801

generall:
- [ ] Support for datetime

generall:
- [ ] API for checking if collection exists: https://github.com/qdrant/qdrant/pull/3472

joein:
- [] update fastembed integration https://github.com/qdrant/qdrant-client/pull/493

PR #491, opened 2024-02-12
Local mode of `order_by` parameter in `scroll`

coszio:
Tracked by #490 

- updates rest and grpc clients as little as possible
- implements local mode of `order_by` parameter of `scroll` API

## TODO
- [x] congruence tests
- [x] make type checker happy
- [x] ~wait for qdrant/qdrant#3574~, won't introduce these changes for now

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


coszio:
CI is now failing because of a semi-bug in core, which is fixed here qdrant/qdrant#3622. It will pass once that gets merged and built to `qdrant/qdrant:dev` image

joein:
Hey @coszio 

Are you using the latest python client generator? I think there should be some changes in models regarding `extra="forbidden"`

qdrant_client/local/local_collection.py
```diff
@@ -964,14 +1009,76 @@ def scroll(
...
+        with_vectors: Union[bool, Sequence[str]] = False,
+    ) -> Tuple[List[types.Record], Optional[types.PointId]]:
+        if isinstance(order_by, grpc.OrderBy):
+            order_by = GrpcToRest.convert_order_by(order_by)
+        if isinstance(order_by, str):
+            order_by = models.OrderBy(key=order_by)
+
+        value_and_ids: List[Tuple[OrderingValue, Tuple[ExtendedPointId, int]]] = []
+
+        for id in self.ids.items():
```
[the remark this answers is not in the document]
joein: in this case it is probably even better to unwrap `id` into `external_id`, `internal_id`.

qdrant_client/local/order_by.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __gt__(self, other: "OrderingValue") -> bool:
+        return self.value > other.value
+
+    def __ge__(self, other: "OrderingValue") -> bool:
+        return self.value >= other.value
+
+    def __le__(self, other: "OrderingValue") -> bool:
+        return self.value <= other.value
+
```
joein, line 24: Why do we need this structure?

tests/congruence_tests/test_scroll_order_by.py
```diff
@@ -0,0 +1,77 @@
...
+def test_simple_scroll() -> None:
+    fixture_points = generate_fixtures(200)
+
+    local_client = init_local()
+    init_client(local_client, fixture_points)
+
+    http_client = init_remote()
+    init_client(http_client, fixture_points)
+    http_client.create_payload_index(
+        COLLECTION_NAME, "rand_number", models.PayloadSchemaType.FLOAT, wait=True
```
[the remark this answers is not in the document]
coszio: Id ordering is not guaranteed. Sorry, just design choice to allow backward compatibility

qdrant_client/conversions/conversion.py
```diff
@@ -1851,6 +1932,52 @@ def convert_with_payload_interface(
...
+    def convert_start_from(cls, model: rest.StartFrom) -> grpc.StartFrom:
+        if isinstance(model, int):
+            return grpc.StartFrom(integer=model)
+        if isinstance(model, float):
+            return grpc.StartFrom(float=model)
+        if isinstance(model, datetime):
+            ts = cls.convert_datetime(model)
+            return grpc.StartFrom(timestamp=ts)
+        if isinstance(model, str):
+            # Pydantic also accepts strings as datetime if they are correctly formatted
```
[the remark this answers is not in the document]
coszio: Not yet in documentation, but they are here
https://github.com/qdrant/qdrant/blob/7ac7398bb129946c7c770f882272a0917d647237/lib/segment/src/types.rs#L1416-L1432

qdrant_client/local/order_by.py
coszio: I feel like it makes types simpler, instead of applying cmp operators on a Union + converting datetimes to a specific type

qdrant_client/conversions/conversion.py
```diff
@@ -1326,6 +1383,25 @@ def convert_range(cls, model: rest.Range) -> grpc.Range:
             lte=model.lte,
         )
 
+    @classmethod
+    def convert_datetime(
+        cls, model: datetime
+    ) -> grpc.google_dot_protobuf_dot_timestamp__pb2.Timestamp:
```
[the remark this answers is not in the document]
coszio: I guess yes 😅

tests/conversions/fixtures.py
```diff
@@ -831,6 +843,29 @@
...
+)
+
+datetime_start_from = grpc.StartFrom(
+    datetime=datetime.datetime.now().isoformat(),
+)
+
+
+order_by = grpc.OrderBy(
+    key="my_field",
+    direction=grpc.Desc,
```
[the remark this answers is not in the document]
coszio: `grpc.Asc` is the default but I will add an extra fixture for this

tests/conversions/fixtures.py
```diff
@@ -54,7 +54,19 @@
     gte=4.0,
 )
 
+timestamp = Timestamp(seconds=12345678, nanos=123456000)
```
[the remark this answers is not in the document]
coszio: It is to test that we actually support microseconds correctly. Is it failing again? I fixed a conversion error a few days back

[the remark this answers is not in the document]
coszio: There is a bug with amount of zeroes, I'll fix it

tests/fixtures/payload.py
```diff
@@ -162,6 +165,11 @@ def one_random_payload_please(idx):
         "text_data": uuid.uuid4().hex,
         "rand_digit": random.randint(0, 9),
         "rand_number": round(random.random(), 5),
+        "rand_datetime": datetime.datetime.fromtimestamp(
+            # random date between 1999 and 2000, with microseconds
+            random.randint(915159600000000, 946695600000000)
```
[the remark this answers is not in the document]
coszio: I agree, let me change it

pyproject.toml
```diff
@@ -55,6 +55,7 @@ Pygments= "2.16.1"
 pyright = "^1.1.293"
 mypy = "^1.0.0"
 types-protobuf = "^4.21.0.5"
+types-python-dateutil = "^2.8.19"
```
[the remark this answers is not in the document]
coszio: I'm gonna try to parse using some custom implementation to avoid this dependency

qdrant_client/local/order_by.py
coszio: We don't actually need this, switching to type alias should be fine

tests/congruence_tests/test_scroll_order_by.py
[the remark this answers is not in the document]
coszio: I will fix this flakyness of the test on Monday

pyproject.toml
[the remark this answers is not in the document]
joein:
```
# These are the formats that should work with just datetime.strptime,
# but it does not handle a variable number of digits in the microseconds section,
# so we have to handle it manually to match what the qdrant core does.
```

Could you please check it again or provide an example?
I tried to find an example with which `strptime` does not work manually, but failed
Also I could not find an example even some random datetime generator

The function I used was:

```python
def another_parse(dt_str):
    for f in [
        "%Y-%m-%dT%H:%M:%S.%f%z",
        "%Y-%m-%d %H:%M:%S.%f%z",
        "%Y-%m-%dT%H:%M:%S.%f%:z",
        "%Y-%m-%d %H:%M:%S.%f%:z",
        "%Y-%m-%dT%H:%M:%S.%f",
        "%Y-%m-%d %H:%M:%S.%f",
        "%Y-%m-%dT%H:%M:%S%z",
        "%Y-%m-%d %H:%M:%S%z",
        "%Y-%m-%dT%H:%M:%S",
        "%Y-%m-%d %H:%M:%S",
        "%Y-%m-%d %H:%M",
        "%Y-%m-%d",
    ]:
        try:
            return datetime.strptime(dt_str, f)
        except Exception:
            pass
```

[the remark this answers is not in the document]
coszio: You're right, it works. When I tested it, I had a bug in the test itself.

tests/congruence_tests/test_scroll_order_by.py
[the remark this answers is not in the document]
coszio: Fixed this by doing an extra step after finishing scroll. I resort based on ID but only on consecutive same-value records

qdrant_client/http/models/models.py
```diff
@@ -2232,6 +2326,10 @@ def __str__(self) -> str:
     SparseIndexTypeOneOf1,
     SparseIndexTypeOneOf2,
 ]
+StartFrom = Union[
+    StrictFloat,
+    datetime,
+]
```
joein, line 2332: https://github.com/qdrant/pydantic_openapi_v3/issues/24

qdrant_client/local/tests/test_datetimes.py
```diff
@@ -0,0 +1,54 @@
...
+    ],
+)
+def test_datetime_to_timestamp(dt: datetime):
+    rest_to_grpc = RestToGrpc.convert_datetime(dt)
+    grpc_to_rest = GrpcToRest.convert_timestamp(rest_to_grpc)
+
+    print(f"dt: {dt}, rest_to_grpc: {rest_to_grpc}, grpc_to_rest: {grpc_to_rest}")
+    assert (
+        dt.utctimetuple() == grpc_to_rest.utctimetuple()
+    ), f"Failed for {dt}, should be equal to {grpc_to_rest}"
```
joein, line 54: should be in `tests/conversions/test_validate_conversions.py`

tests/test_qdrant_client.py
```diff
@@ -1448,7 +1448,7 @@ def test_conditional_payload_update(prefer_grpc):
 
 
 @pytest.mark.parametrize("prefer_grpc", [False, True])
-def test_conditional_payload_update(prefer_grpc):
+def test_conditional_payload_update_2(prefer_grpc):
```
joein, line 1451: ouch

tests/fixtures/payload.py
```diff
@@ -146,6 +148,17 @@
...
 
+start_datetime = datetime(2000, 1, 1)
+end_datetime = datetime(2001, 1, 31)
+
+
+def random_datetime() -> datetime:
+    return start_datetime + timedelta(
+        seconds=random.randint(0, int((end_datetime - start_datetime).total_seconds())),
+        microseconds=random.randint(0, 999999),
+    )
```
joein, line 159: I am somewhat skeptical when dealing with datetimes, I find them really fragile things to work with
Could you generate more datetime variants, please?
E.g., with timezones and without, with microseconds and without (ideally for each format we support)

qdrant_client/http/models/models.py
joein: we want to accept integers here as well, it will work with pydantic v2 https://github.com/pydantic/pydantic/issues/7504
however, it won't work with pydantic v1 which will crash, we need to add `StrictInt`

coszio: Already added to the spec with qdrant/qdrant#3726. In the meantime I will add it manually to integrate it to the PR.

tests/fixtures/payload.py
coszio: How about this

https://github.com/qdrant/qdrant-client/blob/3337b2ae00ff6b8175bd92556fea730e2b0ed66c/tests/fixtures/payload.py#L157-L165

qdrant_client/local/tests/test_datetimes.py
coszio: Done :)

tests/fixtures/payload.py
joein: sounds good, does not work

joein: "%Y-%m-%dT%H:%M:%S.%f%:z"
"%Y-%m-%d %H:%M:%S.%f%:z"
    
Are these actually correct formats?

tests/fixtures/payload.py
```diff
@@ -152,12 +154,16 @@
...
-    return start_datetime + timedelta(
+def random_datetime_str() -> str:
+    random_datetime = start_datetime + timedelta(
         seconds=random.randint(0, int((end_datetime - start_datetime).total_seconds())),
         microseconds=random.randint(0, 999999),
     )
 
+    format = random.choice(datetime_utils.available_formats)
+
+    return random_datetime.strftime(format)
```
joein, line 165: why can't I leave a suggestion :(

joein:
```suggestion
    fmt = random.choice(datetime_utils.available_formats)
    if 'z' in format:
        random_datetime = random_datetime.replace(tzinfo=timezone(offset=timedelta(hours=random.randint(-12, 12))))
    dt_str = random_datetime.strftime(fmt)
    return dt_str
```

also we need to `from datetime import timezone`

tests/fixtures/payload.py
coszio: They are, but they seem to be introduced in python 3.12, I'll remove them since we don't actually need them

tests/fixtures/payload.py
coszio: nice, you just triggered a local implementation bug. I'll fix it

coszio: Well, it wasn't really a bug. But in core we didn't allow timezones without colons. E.g. `+09:00` was accepted but `+0900` was not. 
I have added this flexibility to add both now in qdrant/qdrant#3729 and qdrant/qdrant#3730, and it works well now

joein (approved):
we've decided to handle this case in a separate PR

`qdrant_client/local/tests/test_datetimes.py`

```
        # this is accepted in core but not here, there is no specifier for only-hour offset
        # (
        #     "2021-01-01 00:00:00.000+01",
        #     datetime(2021, 1, 1, 0, 0, 0, tzinfo=timezone(timedelta(hours=10))),
        # ),
```

qdrant_client/http/models/models.py
coszio: Solved

coszio:
Follow up issue: #511

PR #517, opened 2024-03-01
Add support for datetime ranges

coszio:
- local mode for datetime ranges
- add datetime ranges to random filter fixture
- discover a grpc validation bug in core 😮 -> https://github.com/qdrant/qdrant/pull/3742

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -3,6 +3,7 @@ from typing import Any, List, Optional
 import numpy as np
 
 from qdrant_client.http import models
+from qdrant_client.local import datetime_utils
 from qdrant_client.local.geo import boolean_point_in_polygon, geo_distance
 from qdrant_client.local.payload_value_extractor import value_by_key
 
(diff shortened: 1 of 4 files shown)


qdrant_client/local/payload_filters.py
```diff
@@ -86,18 +87,40 @@ def check_geo_polygon(condition: models.GeoPolygon, values: Any) -> bool:
...
-        return False
-    if condition.lte is not None and value > condition.lte:
-        return False
-    if condition.gt is not None and value <= condition.gt:
+    return (
+        (condition.lt is None or value < condition.lt)
+        and (condition.lte is None or value <= condition.lte)
+        and (condition.gt is None or value > condition.gt)
+        and (condition.gte is None or value >= condition.gte)
+    )
```
joein, line 106: I think it has become harder to read

joein, line 106: it might be harder to read, but ok

coszio: I wanted to keep the same sign for the operator (>= for gte, etc). 😄

qdrant_client/local/payload_filters.py
```diff
@@ -86,18 +87,40 @@ def check_geo_polygon(condition: models.GeoPolygon, values: Any) -> bool:
...
+
+
+def check_datetime_range(condition: models.DatetimeRange, value: Any) -> bool:
+    if not isinstance(value, str):
```
joein, line 110: don't we accept datetime objects?

coszio: can we have datetime objects in the payload?

joein: users can add datetime objects to payload, however pydantic will convert them to strings

so we need to store strings in payload, but interface should support working with datetime objects

I think users would expect passing datetime objects to the methods rather than strings

joein: so `check_datetime_range` is correct, it does not need to handle datetime

coszio: I will update the payload fixture then, to account for passing datetime objects to the payload

joein (approved):

PR #518, opened 2024-03-01
new: add collection exists interface

joein:


joein:
seems like recent black update has caused some styling updates in this PR

generall (approved):

fsyj-123:
dose this interface "QdrantClient.collection_exists()" only support version which is bigger than 1.8?

joein:
@fsyj-123 yes

PR #519, opened 2024-03-01
new: add min should clause

joein:


joein:
seems like a recent update of black interfered with this PR :(

generall (approved):

PR #522, opened 2024-03-03
fix: turn on test no fastembed

joein:


.github/workflows/integration-tests.yml
```diff
@@ -55,6 +55,11 @@ jobs:
           export RUNNER_OS=${{ runner.os }}
           QDRANT_VERSION='v1.7.4' ./tests/integration-tests.sh
         shell: bash
+      - name: Fastembed not installed tests
```
coszio, line 58: This phrasing was very confusing to me, maybe:
```suggestion
      - name: Run fastembed tests without fastembed
```

.github/workflows/integration-tests.yml
```diff
@@ -55,6 +55,11 @@ jobs:
...
+        run: |
+          pip3 uninstall fastembed -y
+          pytest tests/test_fastembed.py
```
coszio, line 61: I think we should create a directory for fastembed tests. We may add more files, and with a dedicated directory we won't need to update this script

coszio (changes requested):

.github/workflows/integration-tests.yml
joein: I suggest putting of creating a separate directory until we need it

coszio (approved):

PR #534, opened 2024-03-12
new: propagate timeout from methods to httpx

joein:


pyproject.toml
```diff
@@ -15,7 +15,7 @@ keywords = ["vector", "search", "neural", "matching", "client"]
 
 [tool.poetry.dependencies]
 python = ">=3.8"
-httpx = { version = ">=0.14.0", extras = ["http2"] }
+httpx = { version = ">=0.20.0", extras = ["http2"] }
```
generall, line 18: Less that 1% uses httpx version < 0.2x, so I assume this should be fine

generall (approved):

PR #536, opened 2024-03-12
Add key set payload

joein:


qdrant_client/qdrant_client.py
```diff
@@ -1240,6 +1241,9 @@ def set_payload(
                 If multiple shard_keys are provided, the update will be written to each of them.
                 Only works for collections with `custom` sharding method.
 
+            key:
+                Assigns payload to each point that satisfy this path of property
```
generall, line 1245: Will try to suggest better comment...

joein: fair point.. I just copied it from openapi 🙃

qdrant_client/local/payload_value_extractor.py
```diff
@@ -63,6 +63,53 @@ def _get_value(data: Any, k_list: List[str]) -> None:
...
+    def _set_value(data: Any, k_list: List[str], new_value: Dict[str, Any]) -> None:
+        if not k_list:
+            return
+
+        k = k_list.pop(0)
+
+        if k.endswith("]"):
+            k, index = k.split("[")
+            if k not in data:
+                data[k] = []
```
xzfc, line 78: It crashes on the following inputs:

```python
set_value_by_key({"a": []}, {}, "a.b[0]")
set_value_by_key({"a": []}, {}, "a.b")
```

qdrant_client/local/payload_value_extractor.py
```diff
@@ -115,3 +162,77 @@ def test_value_by_key() -> None:
     assert value_by_key(payload, "nested[].empty", flat=False) == [[], [], None]
     assert value_by_key(payload, "the_null", flat=False) == [None]
     # endregion
+
+
+def test_set_value_by_key() -> None:
```
xzfc, line 167: Could you please add more tests with "unmatching" paths?

E.g. the path assumes dict key, but the data is an array (`set_value_by_key({"a": []}, {}, "a.b[0]")`), and so on.

qdrant_client/local/payload_value_extractor.py
joein: thanks!
will take a look

qdrant_client/local/payload_value_extractor.py
```diff
@@ -63,6 +63,53 @@ def _get_value(data: Any, k_list: List[str]) -> None:
...
+            if not isinstance(data, list):
+                return
+
+            index = index.strip("]")
+            if index == "":
+                for item in data:
+                    if isinstance(item, dict) and len(k_list) == 0:
+                        item.update(new_value)
+                    else:
+                        _set_value(item, k_list.copy(), new_value)
```
xzfc, line 92: This fragment appears three times in this function with slight changes.
Can it be refactored to appear only once? Perhaps it could be moved to the top of this function.

xzfc:
Also, please note that we switched to a new JSON path parser and logic in the original Rust implementation. It has more robust design rather than old logic based on ad-hoc string manipulation. We might eventually want to implement something similar for the Python local mode.

https://github.com/qdrant/qdrant/blob/v1.8.1/lib/segment/src/json_path/parse.rs
https://github.com/qdrant/qdrant/blob/v1.8.1/lib/segment/src/json_path/v2.rs

In short, the new implementation:
- allows quoting keys, e.g. `a."foo.bar".b` to handle payloads like `{"a": {"foo.bar": {"b": []}}}`
- rejects malformed paths, e.g. `a]`
- allows nested arrays, e.g. `a[0][1]`

joein:
An attempt to refactor spaghetti-code #544

joein:
Todo:

- [ ] extend payload fixtures and generated filters with sequential array indexing like `a[][]`

qdrant_client/local/payload_value_extractor.py
```diff
@@ -18,46 +25,46 @@ def value_by_key(payload: dict, key: str, flat: bool = True) -> Optional[List[An
...
+
+        elif current_key.item_type == JsonPathItemType.INDEX:
+            assert current_key.index is not None
+
+            if current_key.index < len(data):
+                _get_value(data[current_key.index], k_list.copy())
+
+        elif current_key.item_type == JsonPathItemType.WILDCARD_INDEX:
+            for item in data:
+                _get_value(item, k_list.copy())
```
xzfc, line 67: Need to check if `isinstance(data, dict)`/`isinstance(data, list)` here.

xzfc (changes requested):

qdrant_client/local/json_path_parser.py
```diff
@@ -0,0 +1,314 @@
...
+    if '"' not in path or not path.startswith('"'):
+        return None, path
+
+    if path.count('"') % 2 != 0:
+        raise ValueError("Invalid path")
+
+    left_quote_pos = 0
+    right_quote_pos = path.find('"', 1)
+
+    if left_quote_pos == (right_quote_pos + 1):
```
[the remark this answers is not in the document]
joein: We might want to support them, but local mode can't go ahead of the server
So it should be implemented in the server first, and then we can add it to local mode

qdrant_client/local/json_path_parser.py
```diff
@@ -0,0 +1,314 @@
...
+        return path[1:]
+
+    elif path.startswith("["):
+        return path
+    else:
+        raise ValueError("Invalid path")
+
+
+def match_quote(path: str) -> Tuple[Optional[JsonPathItem], str]:
+    if '"' not in path or not path.startswith('"'):
```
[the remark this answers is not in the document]
joein: fair point

qdrant_client/local/json_path_parser.py
```diff
@@ -0,0 +1,314 @@
...
+        return None, path
+
+    if path.count('"') % 2 != 0:
```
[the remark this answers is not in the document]
joein: I can replace it with `quote count <2` or with `right_quote_pos != -1`

Not sure if we actually want to support escaped characters in the keys, we can update it once it is implemented in the server

qdrant_client/local/json_path_parser.py
```diff
@@ -0,0 +1,314 @@
...
+
+    return keys, rest
+
+
+def _match_brackets(path: str) -> Tuple[Optional[JsonPathItem], str]:
+    if "[" not in path or not path.startswith("["):
+        return None, path
+
+    left_bracket_pos = 0
+    right_bracket_pos = path.find("]", left_bracket_pos + 1)
```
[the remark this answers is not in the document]
joein: I guess it might not actually throw an error if there is `[10`, I'll fix it

qdrant_client/local/payload_value_setter.py
```diff
@@ -0,0 +1,536 @@
...
+        data: Any,
+        k_list: List[JsonPathItem],
+        value: Dict[str, Any],
+        prev_data: Any,
+        prev_key: Optional[JsonPathItem],
+    ) -> None:
+        if not k_list:
+            return
+
+        current_key = k_list.pop(0)
```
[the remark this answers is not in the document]
joein: I guess we might leave it as is at the moment, we are not trying to squeeze every second in local mode

joein:
We might need to move tests to `local/tests/test_payload.py`, because otherwise pytest does not pick up tests which are defined in files without `test` prefix

qdrant_client/local/payload_value_extractor.py
joein: I'll take a look at it

joein: I fixed it, good catch, thanks

generall (approved):

qdrant_client/local/json_path_parser.py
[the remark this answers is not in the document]
xzfc: Empty keys are supported by Qdrant right now. I.e. these tests pass:

<details>
<summary>Details</summary>

```python
    response = request_with_validation(
        api='/collections/{collection_name}/points/payload',
        method="POST",
        path_params={'collection_name': collection_name},
        body={
            "payload": {"": "bbb"},
            "filter": {
                "must": [
                    {
                        "key": "key4",
                        "match": {
                            "value": "aaa"
                        }
                    }
                ]
            }
        }
    )
    assert response.ok

    response = request_with_validation(
        api='/collections/{collection_name}/points/scroll',
        method="POST",
        path_params={'collection_name': collection_name},
        body={
            "filter": {
                "must": [
                    {
                        "key": "\"\"",
                        "match": {
                            "value": "bbb"
                        }
                    }
                ]
            }
        }
    )
    assert response.ok
    assert len(response.json()['result']['points']) == 3
```

</details>

qdrant_client/local/json_path_parser.py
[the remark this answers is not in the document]
xzfc: Ah, nvm, looking at wrong diff, sorry.

qdrant_client/local/json_path_parser.py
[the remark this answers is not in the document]
joein: yeap, this code works

```
    response = request_with_validation(
        api='/collections/{collection_name}/points/payload',
        method="POST",
        path_params={'collection_name': collection_name},
        body={
            "payload": {"": "bbb"},
            "filter": {
                "must": [
                    {
                        "key": "key4",
                        "match": {
                            "value": "aaa"
                        }
                    }
                ]
            }
        }
    )
    assert response.ok
```

however, when I try to set `key=""` or add `key=""` into a filter, I get errors from qdrant side, e.g. 

```python
    cl.set_payload(
        CN,
        payload={"": "x"},
        points=models.Filter(must=[models.FieldCondition(key="", match=models.MatchValue(value="xc"))]),
    )
```

> b'{"status":{"error":"Format error in JSON body: data did not match any variant of untagged enum Condition at line 1 column 72"},"time":0.0}'

```python
    cl.set_payload(
        CN,
        payload={"": "x"},
        points=[10],
        key="",
    )
```

>b'{"status":{"error":"Format error in JSON body: Invalid json path at line 1 column 43"},"time":0.0}'

I will add tests for these cases

So now we can set payload to `""`, but we can't filter by it, I guess it should be updated on the server side first

xzfc (approved):

PR #538, opened 2024-03-12
fix: make datetime range conditions tz aware

joein:


Anush008 (approved):
Thank you for the fix @joein.

generall (approved):

PR #541, opened 2024-03-14
chore: Added init_options property

Anush008:
### Why?

This change will be helpful in our integration implementations as we won't have to necessitate users to pass instances of both `QdrantClient` and `AsyncQdrantClient`. Either one will suffice.


joein:
Hey @Anush008 

It does not sound like the right motivation

Could you please provide more details and explain the need?

Anush008:
@joein, yes. For instances like these.

https://discord.com/channels/907569970500743200/1206597766999056435/1206601751747371089

Some frameworks require us to pass SDK instances and not the params to construct those. So instead of forcing users to pass 2 instances(sync, async) of essentially the same thing, we can construct one from the other.

joein:
Could you please explain why do they need to create two instances?

Anush008:
To implement the sync and async interfaces of the frameworks.

joein:
So can they just use one client or another one? Why do they need to use both simultaneously?

Anush008:
For example, the [`Qdrant`](https://python.langchain.com/docs/integrations/vectorstores/qdrant) class from Langchain has both sync and async interfaces.

So for
````python

Qdrant#add_documents(..)
...
await Qdrant#asimiliarity_search(...)

...

await Qdrant#aadd_documents(..)
...

Qdrant#similiarity_search(...)
```

to work, we need the sync and async instances within the `Qdrant` class.

joein (approved):

PR #559, opened 2024-04-01
fix: fix update sparse vectors local mode

joein:


qdrant_client/local/local_collection.py
```diff
@@ -1241,7 +1241,11 @@ def upsert(self, points: Union[List[models.PointStruct], models.Batch]) -> None:
 
     def _update_named_vectors(self, idx: int, vectors: Dict[str, List[float]]) -> None:
         for vector_name, vector in vectors.items():
-            self.vectors[vector_name][idx] = np.array(vector)
+            if isinstance(vector, SparseVector):
```
agourlay, line 1244: I am very confused, `vectors` contains only dense vectors according to the signature.
How could it be an instance of `SparseVector`?

joein: @agourlay oops, you're right!

I forgot to update the signature 🙈

PR #562, opened 2024-04-02
fix: fix QueryVector isinstance check in local mode

joein:
if we pass unexpected type to search, check `isinstance(query_vector, QueryVector)` fails when it gets to numpy arrays


coszio (approved):
☹️ sad that there is no easy way of using `get_args` or `get_args_subscribed` in this scenario. But ok ✅

coszio (approved):
Now we're talking 😎

PR #589, opened 2024-04-09
fix: Fix read consistency when performing batch search

Apmats:
Solves https://github.com/qdrant/qdrant-client/issues/590

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/async_qdrant_remote.py
+++ b/qdrant_client/async_qdrant_remote.py
@@ -262,7 +262,7 @@ class AsyncQdrantRemote(AsyncQdrantBase):
                 else r
                 for r in requests
             ]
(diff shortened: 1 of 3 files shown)


joein:
Hey @Apmats 

That's a good catch, thank you!

However, we also need to regenerate the async client (`./tools/generate_async_client.sh`) and add some tests to avoid these problems in the future. Would you like to do it, or should I take over?

Apmats:
@joein running that script is trivial but the tests don't seem so to me. It seems most (all?) of the tests you have are integration tests, but I don't think whether this option is respected can be tested that way easily.

joein (approved):

PR #601, opened 2024-04-16
deprecate recreate_collection

generall:
Apparently, some users don't expect this method to delete existing collection.

Adds deprecation warning as a preparation for further removal.


generall:
@joein there is some problem with formatting.

the tool which generates async client, pre-commit hook, and CI - all three have a unique opinion on how to format the code.

joein (approved):

PR #612, opened 2024-04-26
new: add gpu support for fastembed, add fastembed providers

joein:


qdrant_client/async_qdrant_fastembed.py
```diff
@@ -109,6 +115,7 @@ def set_sparse_model(
         embedding_model_name: Optional[str],
         cache_dir: Optional[str] = None,
         threads: Optional[int] = None,
+        providers: Optional[Sequence[Union[str, Tuple[str, Dict[Any, Any]]]]] = None,
```
generall, line 118: can we do a simpler type?

generall (approved):
On green CI

PR #622, opened 2024-05-07
Update openapi v1 10 x

generall:
Depends on: https://github.com/qdrant/qdrant/pull/4196


tests/fixtures/points.py
```diff
@@ -36,6 +36,19 @@ def generate_random_sparse_vector(size: int, density: float) -> SparseVector:
...
+    if random.random() > 0.5:
+        size = int(size * 0.3)
+
+    num_non_zero = int(size * density)
+    indices: List[int] = random.sample(range(size), num_non_zero)
+
+    values: List[float] = [round(random.random(), 6) for _ in range(num_non_zero)]
+    sparse_vector = SparseVector(indices=indices, values=values)
+    validate_sparse_vector(sparse_vector)
+    return sparse_vector
```
joein, line 49:
```suggestion
    if random.random() > 0.5:
        size = int(size * 0.3)
    return generate_random_sparse_vector(size, density)
```

qdrant_client/http/models/models.py
```diff
@@ -1981,6 +2003,9 @@ class SparseVectorParams(BaseModel, extra="forbid"):
     index: Optional["SparseIndexParams"] = Field(
         default=None, description="Custom params for index. If none - values from collection configuration are used."
     )
+    modifier: Optional["Modifier"] = Field(
```
joein, line 2006: should it be added to `QdrantFastembed.get_fastembed_sparse_vector_params` as well?

qdrant_client/local/local_collection.py
```diff
@@ -1606,6 +1665,14 @@ def batch_update_points(
             else:
                 raise ValueError(f"Unsupported update operation: {type(update_op)}")
 
+    def update_sparce_vectors_config(
```
joein, line 1668: sparse :)

qdrant_client/local/qdrant_local.py
```diff
@@ -625,8 +627,20 @@ def collection_exists(self, collection_name: str, **kwargs: Any) -> bool:
...
+        self,
+        collection_name: str,
+        sparse_vectors_config: Optional[Mapping[str, types.SparseVectorParams]] = None,
+        **kwargs: Any,
+    ) -> bool:
         _collection = self._get_collection(collection_name)
+
+        if sparse_vectors_config is not None:
+            for vector_name, vector_params in sparse_vectors_config.items():
+                collection = self._get_collection(collection_name)
```
joein, line 640: I think we don't need to do this call in the loop

qdrant_client/local/qdrant_local.py
```diff
@@ -21,9 +21,11 @@
 import numpy as np
 import portalocker
 
+from qdrant_client import grpc
```
joein, line 24: imported not used

qdrant_client/local/qdrant_local.py
```diff
@@ -21,9 +21,11 @@
...
 from qdrant_client._pydantic_compat import to_dict
 from qdrant_client.client_base import QdrantBase
 from qdrant_client.conversions import common_types as types
+from qdrant_client.conversions.conversion import GrpcToRest
```
joein, line 28: imported not used

qdrant_client/local/local_collection.py
```diff
@@ -133,6 +135,33 @@ def close(self) -> None:
...
+            self.sparse_vectors_idf[vector_name][idx] += 1
+
+    def _update_idf_remove(self, vector: SparseVector, vector_name: str) -> None:
+        for idx in vector.indices:
+            self.sparse_vectors_idf[vector_name][idx] -= 1
+
+    @classmethod
+    def _compute_idf(cls, df: int, n: int) -> float:
+        # ((n - df + 0.5) / (df + 0.5) + 1.).ln()
+        return math.log((n - df + 0.5) / (df + 0.5) + 1)
```
joein, line 151: why do we need to add `+1` ?

I thought that it might be to avoid a case when the expression in the parentheses is equal to 0, however it is not possible since `n > df`

qdrant_client/local/local_collection.py
```diff
@@ -133,6 +135,33 @@ def close(self) -> None:
...
+
+    def _rescore_idf(self, vector: SparseVector, vector_name: str) -> SparseVector:
+        num_docs = self.count(count_filter=None).count
+        new_values = []
+        idf_store = self.sparse_vectors_idf[vector_name]
+
+        for idx, value in zip(vector.indices, vector.values):
+            term_frequency = idf_store.get(idx, 0)
```
joein, line 159: it's document frequency, isn't it?

joein (approved):

PR #623, opened 2024-05-07
Drop black

generall:
use ruff instead


.pre-commit-config.yaml
```diff
@@ -13,16 +13,13 @@ repos:
...
-      - id: black
-        name: "Black: The uncompromising Python code formatter"
-        args: ["--line-length", "99"]
-
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.13.2
-    hooks:
-      - id: isort
-        name: "Sort Imports"
-        args: ["--profile", "black", "--py", "310"]
```
xzfc, line 28: Is `isort` missing in the new definition?

https://docs.astral.sh/ruff/formatter/#sorting-imports

generall: not needed as well

PR #645, opened 2024-06-10
Fix grpc port by configure via url

ainyava:
I've noticed that when configuring Qdrant endpoint via URL it does not set grpc_port so I've made a simple change.

### DIFF
--- a/qdrant_client/async_qdrant_remote.py
+++ b/qdrant_client/async_qdrant_remote.py
@@ -55,14 +55,17 @@ class AsyncQdrantRemote(AsyncQdrantBase):
     def __init__(
         self,
         url: Optional[str] = None,
+        grpc_url: Optional[str] = None,
         port: Optional[int] = 6333,
         grpc_port: int = 6334,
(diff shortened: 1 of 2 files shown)


joein:
Hi @ainyava 

However, I don't think we'll be able to merge it:

Using both grpc and rest simultaneously is a valid case (and we can't assign the same port for two interfaces)
grpc interface does not implement some of the snapshots/shards/storage methods, and the rest one is used in that case

ainyava:
Oh I didn't know about that, how ever I think its kind of misleading that we cant use URL and host/port params at the same time while its not clear that how to configure both protocols separately

The real problem we're facing is:
We have Qdrant on Kubernetes and have access to it using Ingress (through port 443) so we need to configure separate endpoints (because our Ingress controller adds some annotation for gRPC) so we need to configure port and hostnames gor HTTP/gRPC separately.

I would appreciate some suggestions on how to resolve this in a clear manner, also If its needed to change the source code,
It would be a great pleasure for me to do so.

Thanks, Ainyava

PR #659, opened 2024-06-25
universal-query: Local mode

coszio:
Implementation of local mode for `query`.

~~No tests added yet~~

### DIFF
--- a/qdrant_client/local/async_qdrant_local.py
+++ b/qdrant_client/local/async_qdrant_local.py
@@ -40,7 +40,7 @@ from qdrant_client.http.models.models import RecommendExample
 from qdrant_client.local.local_collection import (
     LocalCollection,
     DEFAULT_VECTOR_NAME,
-    _ignore_mentioned_ids_filter,
+    ignore_mentioned_ids_filter,
 )
 
 META_INFO_FILENAME = "meta.json"
@@ -337,21 +337,16 @@ class AsyncQdrantLocal(AsyncQdrantBase):
     ) -> List[types.Prefetch]:
         if prefetch is None:
             return []
-        elif isinstance(prefetch, list) and len(prefetch) == 0:
+        if isinstance(prefetch, list) and len(prefetch) == 0:
             return []
-        else:
-            prefetches = []
-            if isinstance(prefetch, types.Prefetch):
-                prefetches = (
-                    prefetch.prefetch
-                    if isinstance(prefetch.prefetch, list)
-                    else [prefetch.prefetch]
-                )
-            elif isinstance(prefetch, Sequence):
-                prefetches = list(prefetch)
-            return [
-                self._resolve_prefetch_input(prefetch, collection_name) for prefetch in prefetches
-            ]
+        prefetches = []
+        if isinstance(prefetch, types.Prefetch):
+            prefetches = (
+                prefetch.prefetch if isinstance(prefetch.prefetch, list) else [prefetch.prefetch]
+            )
+        elif isinstance(prefetch, Sequence):
+            prefetches = list(prefetch)
+        return [self._resolve_prefetch_input(prefetch, collection_name) for prefetch in prefetches]
 
     def _resolve_prefetch_input(
         self, prefetch: types.Prefetch, collection_name: str
@@ -362,7 +357,7 @@ class AsyncQdrantLocal(AsyncQdrantBase):
             collection_name, prefetch.query, prefetch.using, prefetch.lookup_from
         )
         prefetch.query = query
-        prefetch.filter = _ignore_mentioned_ids_filter(prefetch.filter, list(mentioned_ids))
+        prefetch.filter = ignore_mentioned_ids_filter(prefetch.filter, list(mentioned_ids))
         prefetch.prefetch = self._resolve_prefetches_input(prefetch.prefetch, collection_name)
         return prefetch
 
@@ -387,7 +382,7 @@ class AsyncQdrantLocal(AsyncQdrantBase):
             (query, mentioned_ids) = self._resolve_query_input(
                 collection_name, query, using, lookup_from
             )
-            query_filter = _ignore_mentioned_ids_filter(query_filter, list(mentioned_ids))
+            query_filter = ignore_mentioned_ids_filter(query_filter, list(mentioned_ids))
         prefetch = self._resolve_prefetches_input(prefetch, collection_name)
         return collection.query_points(
             query=query,
--- a/qdrant_client/local/async_qdrant_local.py
+++ b/qdrant_client/local/async_qdrant_local.py
@@ -15,7 +15,20 @@ import logging
 import os
 import shutil
 from io import TextIOWrapper
-from typing import Any, Dict, Generator, Iterable, List, Mapping, Optional, Sequence, Tuple, Union
+from typing import (
+    Any,
+    Dict,
+    Generator,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+    get_args,
+    Set,
+)
 from uuid import uuid4
 import numpy as np
 import portalocker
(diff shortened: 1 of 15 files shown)


qdrant_client/local/async_qdrant_local.py
```diff
@@ -226,6 +243,117 @@ async def search_groups(
             with_lookup_collection=with_lookup_collection,
         )
 
+    def _resolve_query_input(
```
agourlay, line 246: could be tested in isolation :)

qdrant_client/local/local_collection.py
```diff
@@ -657,17 +658,212 @@ def query(
...
+        query_filter: Optional[types.Filter] = None,
+        score_threshold: Optional[float] = None,
+        with_payload: Union[bool, Sequence[str], types.PayloadSelector] = True,
+        with_vectors: Union[bool, Sequence[str]] = False,
+    ) -> List[types.ScoredPoint]:
+        if isinstance(query, models.FusionQuery):
+            # Fuse results
+            if query.fusion == models.Fusion.RRF:
+                # RRF: Reciprocal Rank Fusion
+                rrf = reciprocal_rank_fusion(responses=sources, limit=limit + offset)
```
agourlay, line 755:
```suggestion
                rrf_results = reciprocal_rank_fusion(responses=sources, limit=limit + offset)
```

qdrant_client/local/qdrant_local.py
```diff
@@ -244,6 +250,137 @@ def search_groups(
             with_lookup_collection=with_lookup_collection,
         )
 
+    def _resolve_query_input(
```
agourlay, line 253: could be tested in isolation :)

joein (changes requested):

qdrant_client/local/local_collection.py
```diff
@@ -1950,3 +2131,51 @@ def info(self) -> models.CollectionInfo:
...
+
+    return query_filter
+
+
+def record_to_scored_point(record: types.Record) -> types.ScoredPoint:
+    return types.ScoredPoint(
+        id=record.id,
+        payload=record.payload,
+        vector=record.vector,
+        score=record.score,
```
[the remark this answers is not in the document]
coszio: Copilot Go! 🙈

tests/fixtures/points.py
```diff
@@ -25,6 +25,23 @@ def random_vectors(
...
+        vector_sizes: Dict[str, int],
+) -> models.VectorStruct:
+    vectors = {}
+    for vector_name, vector_size in vector_sizes.items():
+        vec_count = random.randint(1, 10)
+        vectors[vector_name] = generate_random_multivector(vector_size, vec_count)
+    return vectors
+
+
+def generate_random_multivector(vec_size: int, vec_count) -> List[List[float]]:
```
joein, line 38:
```suggestion
def generate_random_multivector(vec_size: int, vec_count: int) -> List[List[float]]:
```

tests/congruence_tests/test_multivector_updates.py
joein, line 1: I am not sure if it is super useful, I think we might want to merge this test_updates, test_sparse_updates and test_multivector_updates into one, since the only difference is how we generate fixtures

qdrant_client/local/local_collection.py
```diff
@@ -658,16 +659,231 @@ def query_points(
...
         using: Optional[str] = None,
-        lookup_from_collection: Optional["LocalCollection"] = None,
-        lookup_from_vector_name: Optional[str] = None,
         **kwargs: Any,
     ) -> types.QueryResponse:
-        raise NotImplementedError()
+        # Assumes all vectors have been homogenized so that there are no ids in the inputs
+
+        scored_points = []
+        if prefetch is not None and isinstance(prefetch, list) and len(prefetch) > 0:
```
joein:
```suggestion
        if isinstance(prefetch, list) and len(prefetch) > 0:
```

qdrant_client/local/qdrant_local.py
```diff
@@ -25,8 +27,12 @@
 from qdrant_client.client_base import QdrantBase
 from qdrant_client.conversions import common_types as types
 from qdrant_client.http import models as rest_models
-from qdrant_client.http.models.models import RecommendExample
-from qdrant_client.local.local_collection import LocalCollection
+from qdrant_client.http.models.models import RecommendExample, Prefetch
+from qdrant_client.local.local_collection import (
+    LocalCollection,
+    DEFAULT_VECTOR_NAME,
+    _ignore_mentioned_ids_filter,
```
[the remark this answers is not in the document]
coszio: unprotected it

qdrant_client/local/local_collection.py
```diff
@@ -658,16 +659,231 @@ def query_points(
...
+            # It is a hybrid/re-scoring query
+            prefetches = prefetch if isinstance(prefetch, list) else [prefetch]
+            sources = []
+            for prefetch in prefetches:
+                sources.append(self._prefetch(prefetch, offset))
```
joein:
```suggestion
        if isinstance(prefetch, list) and len(prefetch) > 0:
            # It is a hybrid/re-scoring query
            sources = [
                self._prefetch(p, offset) for p in prefetch
            ]
```

qdrant_client/local/local_collection.py
```diff
@@ -658,16 +659,231 @@ def query_points(
...
+            prefetch.limit = prefetch.limit + offset
+        if prefetch.prefetch is not None and isinstance(prefetch.prefetch, list) and len(prefetch.prefetch) > 0:
+            # Recursive case: inner prefetches
+            prefetches = (
+                prefetch.prefetch if isinstance(prefetch.prefetch, list) else [prefetch.prefetch]
+            )
+
+            sources = []
+            for inner_prefetch in prefetches:
+                sources.append(self._prefetch(inner_prefetch, offset))
```
joein:
```suggestion
        prefetches = prefetch.prefetch
        if isinstance(prefetches, list) and len(prefetches) > 0:
            # It is a hybrid/re-scoring query
            sources = [
                self._prefetch(inner_prefetch, offset) for inner_prefetch in prefetches
            ]
```

tests/congruence_tests/test_query.py
```diff
@@ -0,0 +1,1094 @@
...
+    @classmethod
+    def no_query_no_prefetch(cls, client: QdrantBase) -> models.QueryResponse:
+        return client.query_points(collection_name=COLLECTION_NAME, limit=10)
+
+
+# ---- TESTS  ---- #
+
+
+def test_dense_query_lookup_from_another_collection():
+    fixture_points = generate_fixtures()
```
joein, line 525:
```suggestion
    fixture_points = generate_fixtures(10)
```

joein: one of the reasons to introduce this test was to check that we don't exclude lookup collection point ids from the current collection, if we have a lot of points in the original collection, we might not be able to track this issue

tests/congruence_tests/test_query.py
```diff
@@ -0,0 +1,1094 @@
...
+        remote_client,
+        searcher.dense_queries_parametrized,
+        search_params={"indexed_only": True},
+    )
+    compare_client_results(
+        local_client,
+        remote_client,
+        searcher.dense_queries_parametrized,
+        search_params={"quantization": {"ignore": True, "rescore": True, "oversampling": 2.0}},
+    )
```
joein, line 733: local mode does not work with search params
remote mode on the amount of points we have (1000) does not build an hnsw index

all the comparisons are actually equal, we can just leave one set of params in order to check the interface

tests/congruence_tests/test_query.py
coszio: Good point

coszio (commented):
Approved from my side ✅

tests/congruence_tests/test_query.py
joein: I am not sure how the test with "indexed_only": True actually passes, since remote client does not have any indexed points and should not return any points

joein: Apparently, it returns them

joein: The answer:
everything which is below indexing_threshold will still be returned even if `indexed_only: True`

joein (approved):

PR #666, opened 2024-06-28
fix: increase context search limit

joein:
ruff went crazy

the actual fix is increasing the limit for context query

```
-            limit=10,
+            limit=1000,
```

### DIFF
--- a/tests/congruence_tests/test_query.py
+++ b/tests/congruence_tests/test_query.py
@@ -419,12 +419,12 @@ class TestSimpleSearcher:
         )
 
     @classmethod
-    def dense_context_image(cls, client: QdrantBase) -> List[models.ScoredPoint]:
+    def dense_context_image(cls, client: QdrantBase, limit: int) -> List[models.ScoredPoint]:
         return client.query_points(
             collection_name=COLLECTION_NAME,
             query=models.ContextQuery(context=models.ContextPair(positive=11, negative=19)),
             with_payload=True,
-            limit=1000,
+            limit=limit,
             using="image",
         )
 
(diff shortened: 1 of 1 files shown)


tests/congruence_tests/test_query.py
```diff
@@ -414,14 +422,13 @@ def dense_many_discover(cls, client: QdrantBase) -> List[models.ScoredPoint]:
     def dense_context_image(cls, client: QdrantBase) -> List[models.ScoredPoint]:
         return client.query_points(
             collection_name=COLLECTION_NAME,
-            query=models.ContextQuery(
-                context=models.ContextPair(positive=11, negative=19)
-            ),
+            query=models.ContextQuery(context=models.ContextPair(positive=11, negative=19)),
             with_payload=True,
-            limit=10,
+            limit=1000,
```
agourlay, line 427: is there a smaller value that works? 
1000 seems overkill

tests/congruence_tests/test_query.py
```diff
@@ -725,10 +746,16 @@ def test_query_invalid_vector_type():
 
     vector_invalid_type = [1, 2, 3, 4]
     with pytest.raises(ValueError):
-        print(local_client.query_points(collection_name=COLLECTION_NAME, query=vector_invalid_type, using="text"))
+        print(
```
agourlay, line 749: can you remove this `print` please, I forgot about it :face_with_head_bandage:

tests/congruence_tests/test_query.py
joein: we can decrease the collection size, we need to check all the points, because there might be score overlap which will lead to a failure

agourlay (approved):

PR #682, opened 2024-07-08
fix: fix grpc conversion bugs for sparse and multi vectors

joein:
this pr fixes 2 bugs

1) search with sparse vectors via query with grpc
2) upsert multivec points with grpc


coszio (approved):
so, we were using the global instead of the local variable... `model` -> `vector`.

And now we are also converting from query_interface 👍

PR #700, opened 2024-07-22
Test migration of multivectors and sparse vectors

hh-space-invader:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/tests/test_migrate.py
+++ b/tests/test_migrate.py
@@ -8,6 +8,9 @@ from tests.congruence_tests.test_common import (
     compare_collections,
     generate_fixtures,
     generate_sparse_fixtures,
+    sparse_vectors_config,
+    generate_multivector_fixtures,
+    multi_vector_config,
     initialize_fixture_collection,
 )
 
@@ -35,6 +38,15 @@ def remote_client() -> QdrantClient:
     client.close()
 
 
+@pytest.fixture
+def second_remote_client() -> QdrantClient:
+    client = QdrantClient(port=6334)
+    delete_collections(client)
+    yield client
+    delete_collections(client)
+    client.close()
+
+
 def delete_collections(client: QdrantClient) -> None:
     collection_names = [collection.name for collection in client.get_collections().collections]
     for collection_name in collection_names:
@@ -47,7 +59,7 @@ def delete_collections(client: QdrantClient) -> None:
         ("local_client", "remote_client"),
         ("remote_client", "local_client"),
         ("local_client", "second_local_client"),
-        # remote - remote requires two launched Qdrant instances and is not tested
+        ("remote_client", "second_remote_client"),
     ],
 )
 def test_single_vector_collection(source_client, dest_client, request) -> None:
@@ -70,7 +82,7 @@ def test_single_vector_collection(source_client, dest_client, request) -> None:
     points = dense_points  # + sparse_points
     source_client.upload_points(collection_name, points, wait=True)
     source_client.migrate(dest_client)
-    dest_client.upload_points(collection_name, points, wait=True)
+    # dest_client.upload_points(collection_name, points, wait=True)
     compare_collections(
         source_client,
         dest_client,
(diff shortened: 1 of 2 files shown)


tests/test_migrate.py
```diff
@@ -35,6 +38,15 @@ def remote_client() -> QdrantClient:
     client.close()
 
 
+@pytest.fixture
+def second_remote_client() -> QdrantClient:
+    client = QdrantClient(port=6334)
+    delete_collections(client)
+    yield client
+    delete_collections(client)
+    client.close()
```
joein, line 47: we have only one running qdrant instance in the CI, thus, the second remote client will connect to the same database, which does not make sense

tests/test_migrate.py
```diff
@@ -70,7 +82,7 @@ def test_single_vector_collection(source_client, dest_client, request) -> None:
     points = dense_points  # + sparse_points
     source_client.upload_points(collection_name, points, wait=True)
     source_client.migrate(dest_client)
-    dest_client.upload_points(collection_name, points, wait=True)
+    # dest_client.upload_points(collection_name, points, wait=True)
```
joein, line 85: seems like a bug 🙈 
we need to remove it completely

tests/test_migrate.py
```diff
@@ -97,10 +109,48 @@ def test_multiple_vectors_collection(source_client, dest_client, request) -> Non
...
+    multi_vectors_points = generate_multivector_fixtures()
+    points = multi_vectors_points
+    source_client.upload_points(collection_name, points, wait=True)
+    source_client.migrate(dest_client)
+    compare_collections(
+        source_client,
+        dest_client,
+        num_vectors=VECTOR_NUMBER,
+        collection_name=collection_name,
+    )
```
joein, line 126: this test is about a different thing

in this test we have several dense vectors in a collection, "text", "image" and "code".
it's not about multi vectors

dense vectors have a type of "List[float]"
multi vectors have a type of "List[List[float]]"
So it should be a separate test

tests/test_migrate.py
```diff
@@ -97,10 +109,48 @@ def test_multiple_vectors_collection(source_client, dest_client, request) -> Non
...
+    """
+    Args:
+        source_client: fixture
+        dest_client: fixture
+        request: pytest internal object to get launch fixtures from parametrize
+    """
+    source_client: QdrantClient = request.getfixturevalue(source_client)
+    dest_client: QdrantClient = request.getfixturevalue(dest_client)
+    collection_name = "sparse_vector_collection"
+    vectors_config = sparse_vectors_config
```
joein, line 148: nit: you can just pass sparse_vectors_config
(same with `sparse_points` - `points` below

tests/test_migrate.py
joein: this test uses default fixtures, e.g. multi_vector_config, it produces the following vectors:
multi-text
multi-image
multi-code

we also allow users to have unnamed multi vectors (like in `test_single_vector_collection`), could you also add such a test please?

joein (changes requested):

tests/test_migrate.py
```diff
@@ -117,7 +114,70 @@ def test_multiple_vectors_collection(source_client, dest_client, request) -> Non
         ("local_client", "remote_client"),
         ("remote_client", "local_client"),
         ("local_client", "second_local_client"),
-        # remote - remote requires two launched Qdrant instances and is not tested
+    ],
+)
+def test_multivectors_collection(source_client, dest_client, request) -> None:
```
joein, line 119: one more test is required
`test_single_multivector_collection`

it should check migration of a collection with one unnamed multivector

joein:
There was a fix for collection comparison: #702 

In order the CI to be green, we need to rebase this branch

joein:
@hh-space-invader multivectors are only available as of qdrant v1.10, we should add a version check to multivectors related test

here is a link to a method which has a version check
https://github.com/qdrant/qdrant-client/blob/e74fa7607428a29a3ea70c3705722a5a170a9d15/tests/test_fastembed.py#L169

joein (approved):

PR #705, opened 2024-07-22
Random sample in local mode

coszio:
Implements random sampling in local mode and adds the required conversions


tests/congruence_tests/test_query.py
```diff
@@ -1294,3 +1308,18 @@ def test_query_group(prefer_grpc):
...
+
+    searcher = TestSimpleSearcher()
+
+    local_client = init_local()
+    init_client(local_client, fixture_points)
+
+    remote_client = init_remote(prefer_grpc=prefer_grpc)
+    init_client(remote_client, fixture_points)
+
+    compare_client_results(local_client, remote_client, searcher.random_query)
```
agourlay, line 1325: Why are the random implementations between local and server equivalent?
I would expect those to return different values.

qdrant_client/conversions/conversion.py
```diff
@@ -941,6 +941,13 @@ def convert_fusion(cls, model: grpc.Fusion) -> rest.Fusion:
 
         raise ValueError(f"invalid Fusion model: {model}")  # pragma: no cover
 
+    @classmethod
+    def convert_sample(cls, model: grpc.Sample) -> rest.Sample:
+        if model == grpc.Sample.RANDOM:
```
joein, line 946: same placeholder as below

qdrant_client/conversions/conversion.py
```diff
@@ -941,6 +941,13 @@ def convert_fusion(cls, model: grpc.Fusion) -> rest.Fusion:
...
+            return rest.Sample.RANDOM
+
+        raise ValueError(f"invalid Fusion model: {model}")  # pragma: no cover
```
joein, line 949:
```suggestion
        raise ValueError(f"invalid Sample model: {model}")  # pragma: no cover
```

tests/congruence_tests/test_query.py
joein: In this test we basically sample all the points available in the collection
In `random_query` we use `limit=100` and we generate 100 points in `fixture_points = generate_fixtures(100)`

qdrant_client/local/local_collection.py
```diff
@@ -879,6 +878,13 @@ def _query_collection(
                 with_vectors=with_vectors,
             )
             return [record_to_scored_point(record) for record in records[offset:]]
+        elif query == models.SampleQuery(sample=models.Sample.RANDOM):
```
joein, line 881: just nit: seems to be a little bit odd, maybe it's better to isinstance + ==models.Sample.RANDOM?

qdrant_client/local/local_collection.py
```diff
@@ -1767,6 +1773,49 @@ def _scroll_by_value(
...
+            payload_filter=query_filter,
+            ids_inv=self.ids_inv,
+        )
+        # in deleted: 1 - deleted, 0 - not deleted
+        # in payload_mask: 1 - accepted, 0 - rejected
+        # in mask: 1 - ok, 0 - rejected
+        mask = payload_mask & ~self.deleted
+
+        random_scores = np.random.rand(len(self.ids))
+        random_order = np.argsort(random_scores)
```
joein, line 1794: nit: `np.random.permutation(self.inv_ids)`

joein:
my comments are mostly nit, other than that looks fine

joein:
my comments are mostly nit, other than that looks good

tests/congruence_tests/test_query.py
coszio: We return all points postprocessed to be sorted by ID. Just to make sure we can return all of them

qdrant_client/proto/points.proto
```diff
@@ -508,6 +508,15 @@ enum Fusion {
     DBSF = 1; // Distribution-Based Score Fusion
 }
 
+/// Sample points from the collection
+///
+/// Available sampling methods:
+///
+/// * `random` - Random sampling
+enum Sample {
+    RANDOM = 0;
```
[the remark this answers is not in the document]
coszio: You are correct, I went along with the Fusion enum. I'll open a PR on core repo

qdrant_client/local/local_collection.py
coszio: but we need the internal ids to filter against the mask, not the external ones

joein (approved):

PR #712, opened 2024-07-31
Show url which is insecure on client connect

dhruv-anand-aintech:


joein:
hi @dhruv-anand-aintech , thank you for your interest!

We have a slightly different approach to the external contributions
A new branch should be created from `dev` and point to `dev`, also changes to core classes like `QdrantClient` requires a generation of the async client
it can be done with `./tools/generate_async_client.sh`

dhruv-anand-aintech:
Thanks, let me do that

PR #758, opened 2024-08-27
chore: Update clients

coszio:
Updates clients with the latest dev changes.

For grpc I could not install `grpcio==1.48.2`, so I updated it with `grpcio==1.49.1`

New stuff:
- facets
- distance matrix
- strict mode config
- misc


joein:
@coszio 

what was the problem with grpcio?
1.48.2 works fine for me

using 1.49.1 brings type checker issues :(

coszio:
I believe what we need is to pin the `grpcio-tools` version to 1.48.2, but not necessarily for `grpcio`. 

In my computer trying to build `grpcio==1.48.2` fails every time, by relaxing this one, I could generate correctly, as far as I can tell

tools/generate_grpc_client.sh
```diff
@@ -4,8 +4,8 @@ set -e
 
 PROJECT_ROOT="$(pwd)/$(dirname "$0")/../"
 
-pip install grpcio==1.48.2
-pip install grpcio-tools==1.48.2
+pip install "grpcio>=1.48.2,<2.0.0"
+pip install grpcio-tools==1.48.2 # the expected python version to use is 1.10.10
```
joein, line 8:
```suggestion
pip install grpcio-tools==1.48.2 # the expected python version to use is 3.10.10
```

joein (approved):

generall (changes requested):
Please generate grpc with original version of `grpcio`

Do not change the version.

generall (approved):

PR #773, opened 2024-09-13
Fix grpc payload index params

joein:
grpc client used to ignore such params as on_disk and is_tenant for parametrized payload indices, only rest client worked

Integer payload index params such as lookup and range used to be mandatory in rest, however, grpc could omit those params. 
This behaviour was leading to an exception, when one tried to get collection info via a rest client for a collection with payload indices created via grpc client


tests/test_qdrant_client.py
```diff
@@ -2134,13 +2134,21 @@ def test_create_payload_index(prefer_grpc):
...
 
-    client.create_payload_index(COLLECTION_NAME, "keyword", models.PayloadSchemaType.KEYWORD)
-    client.create_payload_index(COLLECTION_NAME, "integer", models.PayloadSchemaType.INTEGER)
-    client.create_payload_index(COLLECTION_NAME, "float", models.PayloadSchemaType.FLOAT)
-    client.create_payload_index(COLLECTION_NAME, "geo", models.PayloadSchemaType.GEO)
-    client.create_payload_index(COLLECTION_NAME, "text", models.PayloadSchemaType.TEXT)
-    client.create_payload_index(COLLECTION_NAME, "bool", models.PayloadSchemaType.BOOL)
-    client.create_payload_index(COLLECTION_NAME, "datetime", models.PayloadSchemaType.DATETIME)
+    client.create_payload_index(
+        COLLECTION_NAME, "keyword", models.PayloadSchemaType.KEYWORD, wait=True
```
coszio, line 2138: I assume all these `wait=true` are added now because of the new non-blocking payload index creation 👍

tests/test_qdrant_client.py
```diff
@@ -2152,36 +2160,55 @@ def test_create_payload_index(prefer_grpc):
...
+            wait=True,
         )
+        if prefer_grpc:
+            rest_client = QdrantClient()
+            _ = rest_client.get_collection(COLLECTION_NAME).payload_schema
+        payload_schema = client.get_collection(COLLECTION_NAME).payload_schema
+        assert payload_schema["integer_parametrized"].params.lookup is True
+        assert payload_schema["integer_parametrized"].params.range is False
+        assert payload_schema["integer_parametrized"].params.is_principal is False
+        assert payload_schema["integer_parametrized"].params.on_disk is True
```
coszio, line 2203: Just discussed this part in chat: 

This is being asserted against the results of rest client, because this is the kind of behavior that would fail before. So this is in the spirit of a regression test.

coszio (approved):

PR #782, opened 2024-09-19
Remove-np-ninf

icfly2:
### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

Numpy 2 dropped support for the np.INF and NINF and instead it recomends np.inf and - np.inf

WIP: I'm still looking for the contribution guide, inparticular the how to run the tests locally bit, as I'm not familiar with poetry. Tests pass in the runners but I'm pretty sure the current test suite doesn't cover the changed code.

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -615,7 +615,7 @@ class LocalCollection:
 
             score = scores[idx]
             # skip undefined scores from sparse vectors
-            if sparse_scoring and score == - np.inf:
+            if sparse_scoring and score == -np.inf:
                 continue
             point_id = self.ids_inv[idx]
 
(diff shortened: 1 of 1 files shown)


icfly2:
I can show that if you upgrade numpy to 2.x it fails the current test suite, but with the change, the client doesn't fail the tests. This requires the client without fastembed, as that is pinned to below 2 for numpy:

```

File: qdrant_client/local/local_collection.py [hunk 1]
```diff
@@ -615,7 +615,7 @@ def search(
 
             score = scores[idx]
             # skip undefined scores from sparse vectors
-            if sparse_scoring and score == np.NINF:
+            if sparse_scoring and score == - np.inf:
```
joein, line 618:
```suggestion
            if sparse_scoring and score == -np.inf:
```

joein:
Thank you for addressing this

joein (approved):

PR #795, opened 2024-09-28
Update README.md, fastembed the way to save it 

VinciGit00:
Screenshot 2024-09-28 alle 16 48 23

Now you have to write > pip install fastembed

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Local mode is useful for development, prototyping and testing.
 ## Fast Embeddings + Simpler API
 
 ```
-pip install qdrant-client[fastembed]
+pip install fastembed
 ```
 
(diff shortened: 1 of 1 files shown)


generall:
i don't think I understand this change. the document  section is about integration with qdrant

VinciGit00:
The command was updated

joein:
@VinciGit00 could you try installing it with `pip3 install "qdrant-client[fastembed]"` ? (with the quotes, since zsh require them)

VinciGit00:
Screenshot 2024-10-02 alle 13 52 23

joein:
that's correct, qdrant-client supports gpu as of v1.9.2

PR #799, opened 2024-10-03
Agnostic local inference

joein:
WIP:

- [x]  extend test set
- [ ] refactor inspectors
- [x] - fix type hints
- [x] - add docstrings 
- [x] - fix interface for inspection of several models


qdrant_client/qdrant_fastembed.py
```diff
@@ -819,3 +774,145 @@ def query_batch(
...
+            query = (
+                GrpcToRest.convert_point_id(query) if isinstance(query, grpc.PointId) else query
+            )
+            return models.NearestQuery(nearest=query)
+
+        if isinstance(query, models.Document):
+            model_name = query.model
+            if model_name is None:
+                raise ValueError(
+                    "`query_points` requires explicit model name specification for `Document`"
```
generall, line 812: We don't know if this method will only be called from `query_points`

qdrant_client/qdrant_fastembed.py
```diff
@@ -819,3 +774,145 @@ def query_batch(
...
+        elif isinstance(data, list):
+            # we don't want to iterate over a vector
+            if data and isinstance(data[0], float):
+                return data
+            return [self._embed_raw_data(value, is_query=is_query) for value in data]
+        return data
+
+    def _embed_document(
+        self, document: models.Document, is_query: bool = False
+    ) -> Union[List[float], models.SparseVector]:
```
generall, line 886: Late interaction can also return `List[List[float]]` I propose to define a type alias for this, otherwise too verbose

generall (approved):

Issue #829, opened 2024-10-29
Image embeddings support in local inference


joein:

joein:
#836

PR #836, opened 2024-10-30
Local inference image support

joein:
- [x] refactor and extend tests
- [x] define INFERENCE_OBJECT_TYPES as a constant in `embed`


joein:
#829

qdrant_client/qdrant_fastembed.py
```diff
@@ -998,3 +1049,28 @@ def _embed_document(self, document: models.Document, is_query: bool = False) ->
...
+
+        Raises:
+            ValueError: If model is not supported
+        """
+        model_name = image.model
+        if model_name in _IMAGE_EMBEDDING_MODELS:
+            embedding_model_inst = self._get_or_init_image_model(
+                model_name=model_name, **(image.options or {})
+            )
+            image_data = base64.b64decode(image.image)
```
generall, line 1070: It does seem like a lot of overhead and not really a user-friendly interface, but I understand why it is done like this (I assume, cause we want local remove work the same)

generall (approved):
Not for this PR, but we would need to discuss how to simplify image selection, as asking users to read image from file and encode it info base64 is not going to fly well

joein:
> Not for this PR, but we would need to discuss how to simplify image selection, as asking users to read image from file and encode it info base64 is not going to fly well

We can convert to base64 on our own, but we would either need PIL for that, or find some other ways 
We would also need an extended image model (because models.Image only accepts strings)

PR #880, opened 2025-01-14
update clients

joein:


generall (approved):

qdrant_client/proto/collections.proto
timvisee, line 1: Could you bump this proto file one more time?

I believe there's another (minor) change in there to add the resharding direction to the resharding info.

timvisee (changes requested):

timvisee (approved):
Thanks for the update!

All good, though it looks like some CI tests are still having a stroke.

PR #898, opened 2025-02-13
fix: use the same name for `check_compatibility` flag in args, warnings and tests

abionics:
Replace `check_version` with `check_compatibility` in warnings and tests to fit flag name

For example it fix this place:

https://github.com/qdrant/qdrant-client/blob/64311d3a19b644647f73f2ac143902f4203a3d77/qdrant_client/qdrant_remote.py#L210

https://github.com/qdrant/qdrant-client/blob/64311d3a19b644647f73f2ac143902f4203a3d77/qdrant_client/qdrant_remote.py#L219


joein:
Hey @abionics 
Thank you for pointing it out and fixing!

PR #903, opened 2025-02-17
Regen REST and gRPC clients

agourlay:
Regenerating clients to accommodate for changes we need in the serverless backend.


joein:
conversion will fail if space or multi strict configs are not set

generall:
> conversion will fail if sparse or multi strict configs are not set

:eye: :white_check_mark: 

Thanks!

joein (approved):

PR #922, opened 2025-03-11
speed ci

hh-space-invader:
improve: Updated vectors with smaller ones

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/tests/congruence_tests/test_facet.py
+++ b/tests/congruence_tests/test_facet.py
@@ -28,7 +28,7 @@ def all_facet_keys() -> list[str]:
 
 @pytest.fixture(scope="module")
 def fixture_points() -> list[models.PointStruct]:
-    return generate_fixtures()
+    return generate_fixtures(1000)
 
 
 @pytest.fixture(scope="module", autouse=True)


joein (changes requested):
This PR might decrease the amount of cases we cover in tests with payload filters, which is undesirable
I've created another PR which updates the amount of vectors only in those tests which do not run queries with payload filters 

Also, it seems that test which consumes the biggest amount of time is tests/embed_tests/test_local_inference

I think it makes sense to optimize it 
Most of the tests there takes more than 5 seconds to run
These tests were the most time consuming on my machine:
test_upsert 17.83
test_upload 38.70
test_propagate_options 21.31
test_inference_object 14.38
test_upload_mixed_batches_upload_points 34.44
test_upload_mixed_batches_upload_collection 33.15

I think it might worth it to add some caches to those tests
E.g., add cache to fastembed classes via mocking, so we would not need to compute embeddings for the same documents when possible (might not be possible when we're using `parallel` in some of the methods)
Moreover, we might even store some of the model instances in cache and reuse them if options we are initializing them with are the same

PR #938, opened 2025-03-26
Rest & grpc client generation fix

I8dNLo:
Now:
Rest client should not break environment by downgrading pydentic to version 1.10
GRPC client generation should execute in a separate python environment of python3.10 and should not also break environment

### DIFF
--- a/tools/generate_rest_client.sh
+++ b/tools/generate_rest_client.sh
@@ -1,23 +1,18 @@
 #!/usr/bin/env bash
 set -e
 
-# Define the project root relative to this script.
 PROJECT_ROOT="$(pwd)/$(dirname "$0")/../"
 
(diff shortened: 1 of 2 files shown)


tools/generate_rest_client.sh
```diff
@@ -1,35 +1,45 @@
...
 fi
 
-cd $(mktemp -d)
-
+cd "$TMP_PYDANTIC"
 git clone git@github.com:qdrant/pydantic_openapi_v3.git
 cd pydantic_openapi_v3
 
+export POETRY_VIRTUALENVS_PATH="$TMP_POETRY_ENV"
+poetry env use python3.10
```
generall, line 32: Why this script now assumes poetry?

PR #954, opened 2025-04-22
Fix precision error in search test

I8dNLo:


tests/congruence_tests/test_common.py
```diff
@@ -372,13 +372,21 @@ def init_client(
...
         vectors_config=vectors_config,
         sparse_vectors_config=sparse_vectors_config,
     )
+    if init_debug_collection:
+        initialize_fixture_collection(
+            client=client,
+            collection_name=collection_name + "_debug",
+            vectors_config=vectors_config,
+            sparse_vectors_config=sparse_vectors_config,
+        )
```
joein, line 389: it seems like all the changes in this file are unnecessary

tests/congruence_tests/test_search.py
```diff
@@ -144,27 +144,46 @@ def filter_search_text_single(
             limit=10,
         )
 
+    def load_query_vectors(self, query):
+        self.query_text = np.array(query.vector['text']).tolist()
+        self.query_image = np.array(query.vector['image']).tolist()
+        self.query_code = np.array(query.vector['code']).tolist()
+
```
joein, line 151: it's a pity, that we have to do it like this, but not sure how to make it better at the current state of the tests, either refactor more, or ditch this

tests/congruence_tests/test_search.py
```diff
@@ -144,27 +144,46 @@ def filter_search_text_single(
...
+    try:
+        compare_client_results(local_client, remote_client, searcher.simple_search_image)
+        compare_client_results(local_client, remote_client, searcher.simple_search_image_select_vector)
+    except AssertionError as e:
+        raise ValueError("Vector:", searcher.query_image) from e
+
+    try:
+        compare_client_results(local_client, remote_client, searcher.simple_search_code)
+    except AssertionError as e:
+        raise ValueError("Vector:", searcher.query_code) from e
```
joein, line 186: this seems to be not very useful, cause tests usually being run and fail in CI, and there we don't have an access to the data and checking queries does not make a lot of sense

tests/congruence_tests/test_search.py
```diff
@@ -205,25 +235,26 @@ def test_simple_opt_vectors_search():
...
-            print(f"\nFailed with filter {query_filter}")
-            raise e
+            raise ValueError("Vector:", searcher.query_text, "Filter:", query_filter) from e
 
 
-def test_single_vector():
-    fixture_points = generate_fixtures(num=200, vectors_sizes=text_vector_size)
 
+def test_single_vector():
+    fixture_points = generate_fixtures(num=201, vectors_sizes=text_vector_size)
```
joein, line 243: adding one fixture point to queries seems to be non obvious

tests/congruence_tests/test_search.py
```diff
@@ -205,25 +235,26 @@ def test_simple_opt_vectors_search():
...
+    searcher.query_text = np.array(query.vector).tolist()
 
     vectors_config = models.VectorParams(
         size=text_vector_size,
         distance=models.Distance.DOT,
     )
 
     local_client = init_local()
-    init_client(local_client, fixture_points, vectors_config=vectors_config)
+    init_client(local_client, fixture_points, vectors_config=vectors_config, init_debug_collection=True)
```
joein, line 254: were we meant to delete debug code?

tests/fixtures/points.py
```diff
@@ -11,43 +11,97 @@
...
+    differ by less than or equal to `thold`.
+
+    Parameters:
+        vectors (np.ndarray): Array of shape (N, D)
+        distance (str): One of {'text', 'image', 'code'}
+        thold (float): Threshold for minimal distance difference
+
+    Returns:
+        bool: True if any two distances are within threshold, else False
+    """
```
joein, line 30: relying on vector names seems to be off
is it possible to rely on distance type?

tests/fixtures/points.py
```diff
@@ -11,43 +11,97 @@
...
+        "too_close() was called. Make sure this is intended, as it may affect performance.",
+        category=UserWarning,
+        stacklevel=2
+    )
+
+    if len(vectors) < 3:
+        return False  # Need at least 2 distances to compare their difference
+
+    last_vec = vectors[-1]
+    other_vecs = vectors[:-1]
```
joein, line 41: this does not look right, query should be a parameter

tests/fixtures/points.py
```diff
@@ -11,43 +11,97 @@
...
+        diff = other_vecs - last_vec
+        distances = np.linalg.norm(diff, axis=1)
+    else:
+        # Cosine distance: 1 - cosine similarity
+        normed_last = last_vec / norm(last_vec)
+        normed_others = other_vecs / norm(other_vecs, axis=1, keepdims=True)
+        distances = 1 - np.dot(normed_others, normed_last)
+
+    diffs = np.abs(distances[:, None] - distances[None, :])
+    np.fill_diagonal(diffs, np.inf)
```
joein, line 58: can we reuse distances from local mode implementation?

joein (changes requested):

tests/congruence_tests/test_search.py
I8dNLo: That's a collection to debug it in future. 
We can save EXACTLY the breaking vector and reproduce any kind of situation of stochastic tests rent offline

I8dNLo: I would like to keep it

tests/congruence_tests/test_common.py
I8dNLo: I would, may be change the name of this collection, but it is the collection I write vectors if they break tests, so we could easily reproduce scenario

tests/congruence_tests/test_search.py
joein: We're running tests with pytest -x, so tests won't run if there's a failure and collections will be persisted

PR #966, opened 2025-04-24
fix: fix bm25 search on empty collection

joein:


coszio (approved):
Thanks for including the test

PR #993, opened 2025-05-19
fix: init_options should not crush with unpickleable objects

joein:


tests/test_qdrant_client.py
```diff
@@ -204,6 +205,14 @@ def test_client_init():
     assert client.init_options["prefix"] == "custom"
     assert client.init_options["metadata"] == {"some-rest-meta": "some-value"}
 
+    ssl_context = (ssl.create_default_context(),)
```
generall, line 208:
```suggestion
    ssl_context = ssl.create_default_context()
```

generall (approved):

PR #1075, opened 2025-09-19
fix: fix minor things

joein:


qdrant_client/async_qdrant_remote.py
```diff
@@ -123,6 +123,10 @@ def __init__(
                 limits = httpx.Limits(max_connections=None, max_keepalive_connections=0)
             elif self._pool_size is not None:
                 limits = httpx.Limits(max_connections=self._pool_size)
+        elif self._pool_size is not None:
+            raise ValueError(
+                f"`pool_size` and `limits` are mutually exclusive. `pool_size`: {pool_size}, `limit`: {limits}"
```
JojiiOfficial, line 128: I think we should test whether `max_connections` is set in `limits` because if a user only specifies `max_keepalive_connections` inside `limits`, this error gets thrown although correct.

joein: I think that's too complicated, if they specify `max_keepalive_connections` in `Limits`, they should also specify `max_connections` via `Limits`, rather than using `Limits` and `pool_size` separately

Issue #1080, opened 2025-09-29
Feature Request: Support `uuid.UUID` as ID type


RomainGoussault:
### Current behavior

Qdrant methods accept only string or integer representations of UUIDs as IDs. Passing a `uuid.UUID` object directly is not supported.

### Problem

In our codebase we prefer to use the `uuid.UUID` type rather than plain strings, since it makes the intent explicit and reduces accidental misuse. However, Qdrant requires converting every UUID to a string before passing it, which adds boilerplate and reduces clarity:

```
answer = await self.client.retrieve(
    collection_name=collection_name,
    ids=[str(point_id) for point_id in point_ids],
    ...
)
```

### Proposed behavior

Allow `uuid.UUID` objects directly in all methods that take or return IDs (retrieve, upsert, delete).
This would remove the need for repetitive str(uuid) conversions and make the API more Pythonic.

cirezd:
yes please, this is much more convenient

PR #1085, opened 2025-10-15
Allow uuids in interface

joein:
#1080 

This PR adds support for direct UUID usage.
Previously, in order to use uuid, users had to `import uuid`, create `UUID` objects and convert them to str.
E.g.:

```python
import uuid
points = [models.PointStruct(id=str(uuid.uuid4()), vector=...)]
```

and now with this pr it is possible to just do

```python
import uuid 
points = [models.PointStruct(id=uuid.uuid4(), vector=...)]
```

### DIFF
--- a/qdrant_client/embed/_inspection_cache.py
+++ b/qdrant_client/embed/_inspection_cache.py
@@ -862,7 +862,13 @@ DEFS = {
         "properties": {
             "has_id": {
                 "description": "ID-based filtering condition",
-                "items": {"anyOf": [{"type": "integer"}, {"type": "string"}]},
+                "items": {
+                    "anyOf": [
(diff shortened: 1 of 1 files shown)


qdrant_client/async_qdrant_client.py
```diff
@@ -1608,7 +1608,7 @@ async def update_vectors(
         wait: bool = True,
         ordering: Optional[types.WriteOrdering] = None,
         shard_key_selector: Optional[types.ShardKeySelector] = None,
-        update_filter: Optional[types.Payload] = None,
+        update_filter: Optional[types.Filter] = None,
```
tbung: Will probably create a conflict with #1090 , not that it matters.

PR #1110, opened 2025-11-13
Drop python3.9

joein:


tests/congruence_tests/test_query.py
```diff
@@ -1,4 +1,4 @@
-from typing import Tuple, Callable, Any, Union
+from typing import Tuple, Callable, Any
```
tbung: `Tuple` has been depracated in favor of bultin `tuple` since 3.9. This might be a good chance to drop those too. See also https://docs.python.org/3/library/typing.html#deprecated-aliases

tbung (commented):
I think you missed some, at least https://github.com/qdrant/qdrant-client/blob/693d1e48a094846edfb7a7fa791b6f091de69b6a/tests/congruence_tests/test_multivector_recommend_queries.py#L1 https://github.com/qdrant/qdrant-client/blob/693d1e48a094846edfb7a7fa791b6f091de69b6a/tests/congruence_tests/test_group_search.py#L1 https://github.com/qdrant/qdrant-client/blob/693d1e48a094846edfb7a7fa791b6f091de69b6a/tests/congruence_tests/test_recommendation.py#L1

Also, https://github.com/qdrant/pydantic_openapi_v3 probably should be updated to generate >3.9 type hints as well, if we care about that. That'd certainly help with finding anything we missed with grep or something.

tests/congruence_tests/test_query.py
joein: wondering how I missed all that...
thanks

PR #902, opened 2025-02-17
new: decode json responses with orjson

joein:

Issue #714, opened 2024-08-02
Huge CPU usage of qdrant client when deserializing payload


GDegrove:
## Current Behavior
When enabling qdrant async client for both GRPC and REST qdrant endpoint, we see a huge increase in CPU usage on our applications. We used a profiler (parca) to analyze the flame graph of our application, and we realized that most of the CPU time is passed in the payload deserialization. 
This makes our application require a lot of CPU when we use qdrant as a vector database in application such as recommendation engine. 

## Steps to Reproduce

1. Create a scroll call to a collection with 100 to 500 items (normal in recommendation use-case)
2. look at the CPU usage of the qdrant client
3.
4.

## Expected Behavior
The deserialization of the qdrant payload is CPU efficient, and does not impact our applications when using it

## Possible Solution
do not use JSON to deserialize, or use an optimized library such as orjson (or any other) to perform the de-serialization.

## Context (Environment)

python 3.12, using Async client. Perform scroll/recomend calls to qdrant.

## Detailed Description

implement a CPU optimized de serialization of the CPU payload
## Possible Implementation

using [orjson](https://github.com/ijl/orjson) as a json library instead of the default python JSON that is known to be inefficient

joein:
hi @GDegrove 

just to be sure, could you also provide the version of pydantic you're using?

GDegrove:
> hi @GDegrove
> 
> just to be sure, could you also provide the version of pydantic you're using?

We are using pydantic 2, for exactitude `version = "2.8.2"`

joein:
We'll try to look into it, thanks for pointing it out

arthur-st:
Pydantic doesn't use the `json` package from the standard library for handling de/serialization. Instead, it handles that internally.

joein:
Unfortunately, it is not possible to use a custom json tool in pydantic v2 (e.g. https://github.com/pydantic/pydantic/discussions/6388)
We've looked into the other places where we could replace the builtin json module with orjson, however, the results are not as promising at the moment (#902)
 We'll continue looking into the possible performance improvements

PR #1122, opened 2025-11-20
Load JSON using pydantic directly in HTTP client

Fogapod:
Partially addresses https://github.com/qdrant/qdrant-client/issues/714

There were 2 attempts at this: https://github.com/qdrant/qdrant-client/pull/756 and https://github.com/qdrant/qdrant-client/pull/902 but they add extra dependency.
Pydantic json parsing is pretty fast. They claim being faster than orjson. Even if they are not there is a benefit of not crossing pyo3 bridge twice.

I used `TypeAdapter` instead of hoops with `obj` property although I'm not sure why this extra layer is needed at all. I checked a few types this is called with and they are all `BaseModel` already so you could just `model_validate_json` on them. If there are some non-pydantic types i guess something like this could be done:
```python
    if issubclass(type_, BaseModel):
        return type_
```

My local pyright shows `8738 errors, 0 warnings, 0 informations` not sure what to think of that. I did run `poetry install`

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/http/api_client.py
+++ b/qdrant_client/http/api_client.py
@@ -4,7 +4,7 @@ from typing import Any, Awaitable, Callable, Dict, Generic, Type, TypeVar, overl
 from urllib.parse import urljoin
 
 from httpx import AsyncClient, Client, Request, Response
-from pydantic import ValidationError
+from pydantic import TypeAdapter, ValidationError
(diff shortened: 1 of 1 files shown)


Fogapod:
Looks like `TypeAdapter` was added in pydantic 2 so this won't work if v1 compatibility is still needed

PR #1149, opened 2026-02-08
Add if_not_exists parameter to create_collection

### LINKED ISSUE #1022: Create Collection if not exists
# Current Situtation:
To create a collection and check if it doesnt exist the User currently has to:

```
if not client.collection_exists():
    client.create_collection(
        collection_name="test_collection"
    )
```

# Future Call:

simply use 

`client.create_collection(collection_name ="test_collection", if_not_exists=True)`

This would be a QoL Change

veeceey:
## Summary
- Adds `if_not_exists` boolean parameter to `create_collection()` method
- When `if_not_exists=True`, only creates collection if it doesn't already exist
- Implements for both `QdrantClient` and `AsyncQdrantClient`
- Includes comprehensive tests for sync and async clients
- Maintains backward compatibility (default `if_not_exists=False`)

## Motivation
Currently users need to manually check if a collection exists before creating it:
```python
if not client.collection_exists("test_collection"):
    client.create_collection(
        collection_name="test_collection"
    )
```

With this change, users can simplify this to:
```python
client.create_collection(
    collection_name="test_collection",
    if_not_exists=True
)
```

## Test Plan
- [x] Added tests for sync client with `if_not_exists=True`
- [x] Added tests for sync client without parameter (default behavior)
- [x] Added tests for async client with `if_not_exists=True`
- [x] Added tests for async client without parameter (default behavior)
- [x] All tests pass locally

## Fixes
- Fixes #1022

### DIFF
--- a/tests/test_create_collection_if_not_exists.py
+++ b/tests/test_create_collection_if_not_exists.py
@@ -43,10 +43,11 @@ def test_create_collection_if_not_exists_false_sync(sync_client):
     """Test that create_collection with if_not_exists=False still creates collection"""
     collection_name = "test_collection_if_not_exists_false"
 
-    # Create collection without if_not_exists (default behavior)
+    # Create collection with explicit if_not_exists=False
     result = sync_client.create_collection(
         collection_name=collection_name,
         vectors_config=models.VectorParams(size=10, distance=models.Distance.COSINE),
+        if_not_exists=False,
     )
     assert result is True
     assert sync_client.collection_exists(collection_name)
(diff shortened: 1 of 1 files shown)


veeceey:
Thanks for the review, CodeRabbit! Good suggestion on adding a negative test case for `if_not_exists=False` on an existing collection. That would confirm the default behavior is preserved and document the expected failure mode. I'll add that test case and push an update.

veeceey:
friendly bump - let me know if there's any feedback!

PR #1156, opened 2026-02-19
Fix/validate nan query vectors

Goodnight77:
### description 
 
validate query vectors for nan values in grpc client before sending to the server
before : grpc client would return points with nan scores when given vector that contains nan, while local and http clients properly rejected it 
those nan values in query vectors would propagate through distance calculations, this can causes confusion for users who receives invalid results without clear error message 

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/qdrant_client/conversions/conversion.py
+++ b/qdrant_client/conversions/conversion.py
@@ -3727,11 +3727,13 @@ class RestToGrpc:
 
     @staticmethod
     def _validate_no_nan(vector: list[float]) -> None:
-        """validate that vector does not contain NaN values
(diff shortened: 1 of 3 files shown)


Goodnight77:
@joein

joein:
Hey @Goodnight77 

Thanks for the attempt to fix it, but I'd rather find some other solution than make a check on each and every vector in python client since it might quickly cause a performance degradation :(

Goodnight77:
@joein thnx for your return, that makes sense, i will drop this approach now and think of another solution that avoids the performance cost

Issue #348, opened 2023-10-24
Add raises to docstrings


joein:

Jai-Dhiman:
Hey, I'd love to contribute to the project, I'm a first time open-source contributer, is this still an issue?

Muhammad-Danyal353:
@joein 
Hello there, I am working on this issue: adding "raises <exception type>" to docstrings.

Muhammad-Danyal353:
@joein Opened a PR for this issue: https://github.com/qdrant/qdrant-client/pull/1185
This update adds explicit Raises sections to sync and async client docstrings to document expected exceptions. Feedback is welcome.

PR #1159, opened 2026-02-27
docs: add 'Raises' sections to QdrantClient docstrings

toroleapinc:
## Summary

Add `Raises` documentation sections to methods and properties in `QdrantClient` that explicitly raise exceptions.

## Changes

Added `Raises` sections to the following:

- **`QdrantClient` class docstring** (`__init__`): Documents `ValueError` raised when conflicting connection parameters are specified or when `cloud_inference` is used with local mode.
- **`grpc_collections` property**: Documents `NotImplementedError` when used with local/in-memory mode.
- **`grpc_points` property**: Documents `NotImplementedError` when used with local/in-memory mode.
- **`http` property**: Documents `NotImplementedError` when used with local/in-memory mode.
- **`update_collection` method**: Documents `ValueError` when both `optimizer_config` and `optimizers_config` are specified.

## Notes

This is a first pass focusing on the main `QdrantClient` class methods that directly raise exceptions. Additional methods in the remote/local backends and the async client can be documented in follow-up PRs.

Relates to #348

### DIFF
--- a/qdrant_client/qdrant_client.py
+++ b/qdrant_client/qdrant_client.py
@@ -78,6 +78,10 @@ class QdrantClient(QdrantFastembedMixin):
             inherited from `httpx` (default: 100)
         headers: Custom headers to send with every request.
         **kwargs: Additional arguments passed directly into REST client initialization
+
+    Raises:
+        ValueError: If more than one of ``location``, ``url``, ``host``, or ``path`` is specified,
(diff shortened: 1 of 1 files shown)


toroleapinc:
Rebased on master. Clean diff now — just docstring additions to `qdrant_client.py` matching the existing `raise` sites (issue #348). @joein when you have a minute.

PR #1160, opened 2026-02-28
perf: replace list.pop(0) with deque.popleft() in ModelEmbedder

giulio-leone:
## Problem

`_next_embed()` consumes computed embeddings from `_embed_storage` front-to-back via `list.pop(0)`, which is **O(n)** because every remaining element must be shifted.  For large embedding batches (hundreds/thousands of vectors), the consumption loop becomes **O(n²)**.

## Solution

Switch `_embed_storage` values from `list` to `collections.deque`, replacing `.pop(0)` with `.popleft()` for **O(1)** front removal.

## Changes

- `qdrant_client/embed/model_embedder.py`:
  - Import `deque` from `collections`
  - Type annotation: `dict[str, deque[NumericVector]]`
  - Wrap `embed()` result in `deque()`
  - `.pop(0)` → `.popleft()`

## Testing

- All 26 embed tests pass (2 consecutive runs)
- Schema parser, utils, inspectors all green

### DIFF
--- a/qdrant_client/embed/model_embedder.py
+++ b/qdrant_client/embed/model_embedder.py
@@ -1,5 +1,5 @@
 import os
-from collections import defaultdict
+from collections import defaultdict, deque
 from copy import deepcopy
 from multiprocessing import get_all_start_methods
 from typing import Iterable, Any, Type, get_args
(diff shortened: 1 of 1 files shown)


giulio-leone:
Hi! Gentle ping — this PR is rebased, CI passes, and ready for review. Happy to address any feedback. Thanks!

PR #1178, opened 2026-03-23
fix: offload sync Pydantic serialization to thread in async write path

### LINKED ISSUE #1175: Pydantic serialization invoked by async client can block main loop
Well-behaved async functions in Python never block: they use `await` to wait for I/O, or thread dispatch to move CPU-bound work out-of-thread (and use `await` to pass control of the main thread off to a different coroutine while waiting for that CPU-bound work to complete).

When using the qdrant Python client, I've had py-spy based instrumentation for detecting event loop blockage return stack traces like the following:

```
pydantic/main.py:BaseModel.model_dump_json:519
qdrant_client/http/api/points_api.py:to_json:20
qdrant_client/http/api/points_api.py:jsonable_encoder:35
qdrant_client/http/api/points_api.py:_PointsApi._build_for_upsert_points:509
qdrant_client/http/api/points_api.py:AsyncPointsApi.upsert_points:751
qdrant_client/async_qdrant_remote.py:AsyncQdrantRemote.upsert:1677
qdrant_client/async_qdrant_client.py:AsyncQdrantClient.upsert:1579
myapp_redacted/vectordb/adapter.py:QdrantVectorDBAdapter.create_points:119
```

...with upwards of a half second of CPU time in some cases. (The line numbers above are from qdrant_client version 1.15.1, but the problem has been verified to still be present in 1.17.1)

This is not great: anything CPU-bound (the JSON encoding, in this case) should be moved out-of-process from the async event loop when called from an async function.

---

#### Root cause

The async write path (`AsyncPointsApi.upsert_points`) calls the synchronous `_build_for_upsert_points`, which runs `jsonable_encoder` → `model_dump_json` on the entire `PointInsertOperations` model tree — including all embedding vectors and payloads — before the first `await`. This blocks the event loop for the entire duration of Pydantic serialization.

```python
# points_api.py — async method delegates to sync builder
async def upsert_points(self, ...):
    return await self._build_for_upsert_points(...)  # sync work happens here, before the await

def _build_for_upsert_points(self, ...):
    body = jsonable_encoder(point_insert_operations)  # <-- blocks event loop
    return self.api_client.request(..., content=body)
```

The same pattern affects all async write endpoints that use `jsonable_encoder` (search, scroll, etc.), but upsert is the most impactful because the payload includes high-dimensional float vectors.

#### Reproducer

The script below mocks out the HTTP transport to isolate serialization cost from network I/O, and instruments `_build_for_upsert_points` to measure the synchronous time spent on the event loop:

<details>
<summary>reproducer.py (click to expand)</summary>

```python
"""Reproducer: AsyncQdrantClient.upsert blocks the event loop with synchronous serialization.

Network I/O is mocked out — this measures only the serialization cost.
Tested against qdrant-client 1.15.1 and 1.17.1 with identical results.
"""

import asyncio
import random
import time
from unittest.mock import patch

import httpx
from qdrant_client import AsyncQdrantClient
from qdrant_client.http.api import points_api
from qdrant_client.models import PointStruct, SparseVector


def make_points(n: int, dim: int = 1536) -> list[PointStruct]:
    """Build n PointStructs with realistic payloads and embedding vectors."""
    rng = random.Random(42)
    return [
        PointStruct(
            id=f"chunk-{i}",
            vector={
                "dense": [rng.random() for _ in range(dim)],
                "sparse": SparseVector(
                    indices=list(range(50)),
                    values=[rng.random() for _ in range(50)],
                ),
            },
            payload={
                "org_id": "550e8400-e29b-41d4-a716-446655440000",
                "investigation_id": "6ba7b810-9dad-11d1-80b4-00c04fd430c8",
                "report_id": "6ba7b811-9dad-11d1-80b4-00c04fd430c8",
                "text": "Sample investigation report text. " * 10,
                "section": "analysis",
                "field_names": ["executive_summary", "analysis_details"],
                "entities": [
                    {"type_name": "IP Address", "value": f"192.168.1.{j}",
                     "type_hash": "abc123", "source": "direct",
                     "in_report_text": True, "invocation_count": 1}
                    for j in range(5)
                ],
                "is_split": False, "chunk_index": i, "total_chunks": n,
                "term_frequencies": {f"token_{k}": k + 1 for k in range(100)},
                "inserted_at": 1700000000, "report_created_at": 1699999000,
                "dense_model": "text-embedding-3-small", "dense_version": 3,
                "sparse_model": "bm25", "sparse_version": 1,
                "verdict": ["Malicious"], "confidence": "High",
            },
        )
        for i in range(n)
    ]


_original_build = points_api._PointsApi._build_for_upsert_points

def _timed_build(self, **kwargs):
    t0 = time.perf_counter()
    result = _original_build(self, **kwargs)
    _timed_build.last_elapsed = time.perf_counter() - t0
    return result
_timed_build.last_elapsed = 0.0

def _ok_response(*_args, **_kwargs):
    return httpx.Response(200, json={
        "result": {"operation_id": 1, "status": "completed"},
        "status": "ok", "time": 0.001,
    })


async def measure(n_points: int) -> tuple[float, float]:
    points = make_points(n_points)
    client = AsyncQdrantClient(url="http://localhost:6333", check_compatibility=False)
    points_api._PointsApi._build_for_upsert_points = _timed_build
    try:
        with patch.object(
            client._client.openapi_client.points_api.api_client._async_client,
            "send", side_effect=_ok_response,
        ):
            t0 = time.perf_counter()
            await client._client.upsert(collection_name="test", points=points, wait=False)
            wall = time.perf_counter() - t0
    finally:
        points_api._PointsApi._build_for_upsert_points = _original_build
    return wall, _timed_build.last_elapsed


async def main():
    await measure(5)  # warmup
    print(f"{'points':>7s}  {'wall':>10s}  {'sync_block':>10s}  {'pct_blocked':>11s}")
    print("-" * 48)
    for n in [10, 25, 50, 100, 200]:
        wall, sync_block = await measure(n)
        pct = (sync_block / wall * 100) if wall > 0 else 0
        print(f"{n:>7d}  {wall*1000:>9.1f}ms  {sync_block*1000:>9.1f}ms  {pct:>10.1f}%")

if __name__ == "__main__":
    asyncio.run(main())
```

</details>

#### Results (qdrant-client 1.17.1, Python 3.13, macOS)

```
 points        wall  sync_block  pct_blocked
------------------------------------------------
     10        1.6ms        1.3ms        83.8%
     25        3.6ms        3.2ms        90.4%
     50        6.4ms        6.1ms        95.6%
    100       13.2ms       12.8ms        96.9%
    200       26.5ms       26.0ms        98.3%
```

With `asyncio` debug mode enabled (`asyncio.run(main(), debug=True)`), the slow callback warning fires:

```
WARNING:asyncio:Executing <Task ...> took 0.166 seconds
```

The event loop is blocked for 83–98% of the wall time of an upsert call, scaling linearly with point count. With real-world payloads and larger batch sizes, this reaches 500ms+ as shown in the production stack trace above.

#### Impact

Any other coroutines sharing the event loop — HTTP request handlers, health checks, heartbeats, timeouts — are starved for the duration of serialization. This is particularly problematic in web servers (FastAPI/Litestar + uvicorn) and task workers where the event loop is shared.

#### Prior art

- #615 reported latency scaling with concurrency in the async client; the maintainer suggested trying gRPC, but the root cause (sync serialization blocking the loop) was never identified.
- #714 reported high CPU on deserialization; #902 addresses the read path with orjson but does not address the write path.
- #1167 (closed) identified another sync-in-async issue (`httpx.get` in `__init__`), confirming receptiveness to this class of fix.

#### Suggested fix

The minimal fix for the async write path: offload the synchronous serialization work to a thread in async API methods, e.g. by running `jsonable_encoder` via `asyncio.get_running_loop().run_in_executor()`.

Separately, making `jsonable_encoder` pluggable (so callers can supply a faster serializer like orjson or msgspec) would be a useful mitigation — composing well with the direction of #902 — but wouldn't fix the underlying issue: any amount of synchronous CPU work on the event loop is architecturally wrong, even if a faster serializer makes the symptom less severe.

charles-dyfis-net:
AsyncPointsApi methods for `upsert_points`, `batch_update`, and `update_vectors` call `_build_for_*` which runs `jsonable_encoder` (thus Pydantic's `model_dump_json`) synchronously before the first `await`. This blocks the event loop for the entire duration of Pydantic serialization, which scales with point count and vector dimensionality.

Move the sync `_build_for_*` call into `run_in_executor` for these three methods (skipping some others where body size is bounded and small). The `_build_for_*` methods return an unawaited coroutine object (from the async api_client.request), so the executor runs the serialization in a thread and we await the resulting coroutine back on the event loop.

Fixes #1175

### DIFF
--- a/qdrant_client/http/api/points_api.py
+++ b/qdrant_client/http/api/points_api.py
@@ -1,4 +1,5 @@
 # flake8: noqa E501
+import asyncio
 from typing import TYPE_CHECKING, Any, Dict, Set, TypeVar, Union
 
 from pydantic import BaseModel
@@ -562,13 +563,19 @@ class AsyncPointsApi(_PointsApi):
         """
         Apply a series of update operations for points, vectors and payloads
         """
-        return await self._build_for_batch_update(
+        # _build_for_batch_update is sync: it serializes the body with
+        # jsonable_encoder (CPU-bound, can be slow) then calls
+        # self.api_client.request() which merely creates a coroutine.
+        # Run the sync part in a thread, then await the resulting coroutine.
+        coro = await asyncio.to_thread(
+            self._build_for_batch_update,
             collection_name=collection_name,
             wait=wait,
             ordering=ordering,
             timeout=timeout,
             update_operations=update_operations,
         )
+        return await coro
 
     async def clear_payload(
         self,
@@ -778,13 +785,16 @@ class AsyncPointsApi(_PointsApi):
         """
         Update specified named vectors on points, keep unspecified vectors intact.
         """
-        return await self._build_for_update_vectors(
+        # See batch_update for rationale.
+        coro = await asyncio.to_thread(
+            self._build_for_update_vectors,
             collection_name=collection_name,
             wait=wait,
             ordering=ordering,
             timeout=timeout,
             update_vectors=update_vectors,
         )
+        return await coro
 
     async def upsert_points(
         self,
@@ -797,13 +807,16 @@ class AsyncPointsApi(_PointsApi):
         """
         Perform insert + updates on points. If point with given ID already exists - it will be overwritten.
         """
-        return await self._build_for_upsert_points(
+        # See batch_update for rationale.
+        coro = await asyncio.to_thread(
+            self._build_for_upsert_points,
             collection_name=collection_name,
             wait=wait,
             ordering=ordering,
             timeout=timeout,
             point_insert_operations=point_insert_operations,
         )
+        return await coro
 
 
 class SyncPointsApi(_PointsApi):


charles-dyfis-net:
Some contribution process notes:

- Troubleshooting was AI-assisted: I presented Claude with a stack trace from production performance troubleshooting tools -- quoted from in #1175 -- and interactively walked/talked through the relevant code with the model.
- I did not run the pre-commit hooks because they're tied specifically to Python 3.10, and I don't keep an interpreter that old handy. I _have_ run the pytest suite, and the only error was in unrelated code due to a version mismatch (my local qdrant server is somewhat behind).
- A Netlify error installing qdrant_sphinx_theme in the "deploy preview" does not appear to be relevant to the changes at hand.

charles-dyfis-net:
Going to revise this slightly to use `asyncio.to_thread()` in place of `run_in_executor`.

PR #1182, opened 2026-03-28
Close gRPC channels and HTTP clients in batch uploaders

### LINKED ISSUE #1172: Resource leak: gRPC channels and HTTP clients not closed in batch uploaders
## What's happening

When using `upload_collection()` or `upload_points()` with `parallel > 1`, each worker creates its own gRPC channel or HTTP client — but neither is ever closed after the upload finishes.

### gRPC uploader

In `qdrant_client/uploader/grpc_uploader.py`, `process_upload()` calls `get_channel()` to create a new gRPC channel, but there's no corresponding `channel.close()`:

```python
def process_upload(self, items):
    channel = get_channel(host=self._host, port=self._port, ...)  # opened here
    points_client = grpc.PointsStub(channel)
    for batch in items:
        yield upload_batch_grpc(...)
    # channel is never closed
```

### REST uploader

In `qdrant_client/uploader/rest_uploader.py`, `__init__()` creates a `SyncApis` client that's also never cleaned up:

```python
class RestBatchUploader(BaseUploader):
    def __init__(self, ...):
        self.openapi_client: SyncApis = SyncApis(host=uri, **kwargs)
        # no close() or __del__ anywhere
```

## Impact

Each parallel worker leaks a connection. If you're doing repeated uploads with `parallel=4`, that's 4 orphaned connections per call that pile up over time. In long-running services this leads to file descriptor exhaustion and connection pool issues.

## Steps to reproduce

```python
from qdrant_client.models import PointStruct, VectorParams, Distance

client = QdrantClient("http://localhost:6333")

points = [PointStruct(id=i, vector=[0.1]*4, payload={"x": i}) for i in range(1000)]

# Each call leaks gRPC/HTTP connections from worker processes
for _ in range(100):
    client.upload_points("test", points, parallel=4)
```

## Suggested fix

- `GrpcBatchUploader.process_upload()`: close the channel after the upload loop (e.g. in a `finally` block)
- `RestBatchUploader`: add a cleanup method and call `self.openapi_client.close()` when the worker is done

joaquinhuigomez:
## Summary
- Close gRPC channels and HTTP clients in batch uploaders after upload completes
- Prevents file descriptor leaks when using parallel > 1

Each parallel worker in `GrpcBatchUploader` and `RestBatchUploader` creates its own gRPC channel or HTTP client. These were never closed after the upload loop finished, leaking file descriptors. Wrapping the upload loops in `try/finally` ensures cleanup happens both on normal completion and on error.

Fixes #1172

### DIFF
--- a/qdrant_client/uploader/rest_uploader.py
+++ b/qdrant_client/uploader/rest_uploader.py
@@ -112,14 +112,17 @@ class RestBatchUploader(BaseUploader):
         return cls(uri=uri, collection_name=collection_name, max_retries=max_retries, **kwargs)
 
     def process(self, items: Iterable[Any]) -> Iterable[bool]:
-        for batch in items:
-            yield upload_batch(
-                self.openapi_client,
-                self.collection_name,
-                batch,
-                shard_key_selector=self._shard_key_selector,
-                max_retries=self.max_retries,
-                update_filter=self._update_filter,
-                update_mode=self._update_mode,
-                wait=self._wait,
-            )
+        try:
+            for batch in items:
+                yield upload_batch(
+                    self.openapi_client,
+                    self.collection_name,
+                    batch,
+                    shard_key_selector=self._shard_key_selector,
+                    max_retries=self.max_retries,
+                    update_filter=self._update_filter,
+                    update_mode=self._update_mode,
+                    wait=self._wait,
+                )
+        finally:
+            self.openapi_client.close()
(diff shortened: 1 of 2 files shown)


qdrant_client/uploader/rest_uploader.py
```diff
@@ -112,14 +112,17 @@ def start(
...
+                    self.collection_name,
+                    batch,
+                    shard_key_selector=self._shard_key_selector,
+                    max_retries=self.max_retries,
+                    update_filter=self._update_filter,
+                    update_mode=self._update_mode,
+                    wait=self._wait,
+                )
+        finally:
+            self.openapi_client.close()
```
joein, line 128: We instantiate the client in init, but close it in `process`, despite in the current code `process` is used only once per class instance, it does not seem to be correct to close the client here.

PR #1183, opened 2026-03-31
Fix: Prevent indefinite hang in batch_update with Exponential Backoff

RickyMoon800:
## Proactive Fix
Implemented an Exponential Backoff retry mechanism in the connection layer. 
This prevents the Python client from hanging indefinitely when a network connection is dropped during large batch operations.

Bounty Address (USDT-TRC20): `TPCeXK1ziCgwZM8JgypZzdKxBemV7Q2TJn`


RickyMoon800:
Closing due to incorrect file overwrite. Apologies for the noise, a corrected surgical patch will be provided later.

PR #1186, opened 2026-04-03
ci: add release package sanity check

### LINKED ISSUE #958: Add release sanity check to CI

morgenyu:
Adds a release package sanity check to CI for built distributions.

This new job:
- builds the package with `poetry build`
- creates a fresh virtual environment
- installs the built wheel
- verifies that `pytest` is not included as a runtime dependency
- imports installed non-test `qdrant_client` modules to catch packaging/runtime import issues

The smoke check removes repository paths from `sys.path` before importing so it validates the installed wheel rather than the local source tree.

Closes #958

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -15,9 +15,9 @@ jobs:
     name: Release package sanity test
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


joein:
Hey @morgenyu 

Thank you for attempting this!

Though, the PR was a bit complicated, we didn't want to have another python script for this, closing this in favour of #1207

Issue #1189, opened 2026-04-11
Bug: Local memory mode incorrectly computes values_count for dict payloads


litt1e-c:
### Describe the bug
There is a behavioral inconsistency between the Remote server (Rust engine) and the Local `:memory:` engine in the Python client when applying the `values_count` filter on dictionary (JSON object) payloads.

According to Qdrant semantics, a scalar or a JSON object/dict does not have array dimensions, so its `values_count` should evaluate to `1`. The Remote server handles this correctly. However, the Local `:memory:` engine incorrectly treats the `dict` as an array and returns the number of keys in the dictionary (`len(dict)`). 

### Root Cause Analysis
In the local engine implementation, the logic checks if a payload value is an array using something equivalent to `hasattr(value, "__len__") and not isinstance(value, str)`. Since a Python `dict` has a `__len__` attribute, it falls into this branch, and its length (number of keys) is returned instead of `1`.

### Reproducible Code
Here is a minimal reproducible example demonstrating the discrepancy. The remote server correctly passes the assertions, while the local mode fails.

```python
from qdrant_client import QdrantClient
from qdrant_client.models import (
    Distance,
    FieldCondition,
    Filter,
    PointStruct,
    ValuesCount,
    VectorParams,
)

HOST = "127.0.0.1"
PORT = 6333
GRPC_PORT = 6334

COLLECTION = "values_count_dict_repro"


def ids_from_scroll(client, flt):
    points, _ = client.scroll(
        collection_name=COLLECTION,
        scroll_filter=flt,
        limit=20,
        with_payload=True,
        with_vectors=False,
    )
    return sorted([p.id for p in points])


def main():
    remote = QdrantClient(host=HOST, port=PORT, grpc_port=GRPC_PORT, prefer_grpc=False)
    local = QdrantClient(":memory:")

    for client in (remote, local):
        client.recreate_collection(
            collection_name=COLLECTION,
            vectors_config=VectorParams(size=2, distance=Distance.COSINE),
        )
        client.upsert(
            collection_name=COLLECTION,
            wait=True,
            points=[
                # Scalar dict: should be treated as a single element per docs, count == 1
                PointStruct(id=1, vector=[0.1, 0.2], payload={"vc": {"a": 1, "b": 2}}),
                # Actual array: count == 2
                PointStruct(id=2, vector=[0.2, 0.3], payload={"vc": ["x", "y"]}),
                # Normal scalar: count == 1
[...]

PR #1191, opened 2026-04-13
fix: fix local mode values count

joein:
#1189

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -20,7 +20,7 @@ def get_value_counts(values: list[Any]) -> list[int]:
         for value in values:
             if value is None:
                 counts.append(0)
-            elif hasattr(value, "__len__") and not isinstance(value, str):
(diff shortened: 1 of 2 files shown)


qdrant_client/local/payload_filters.py
```diff
@@ -20,7 +20,7 @@ def get_value_counts(values: list[Any]) -> list[int]:
         for value in values:
             if value is None:
                 counts.append(0)
-            elif hasattr(value, "__len__") and not isinstance(value, str):
+            elif isinstance(value, list):
```
tbung, line 23: `collections.abc.Sequence` would probably be the generalized/more robust check here, but I guess it doesn't matter since we control what types go here

tbung (approved):

qdrant_client/local/payload_filters.py
joein: when we add a point to a collection we serialize it to json and then deserialize, so I guess there should not be any tough types and list should be enough

PR #1192, opened 2026-04-18
fix: treat dict payloads as scalar values in values_count filter

rahulkini31:
## Summary

- `get_value_counts` in `payload_filters.py` used `hasattr(value, "__len__")` to detect array-like values, but `dict` also has `__len__`, causing JSON object payloads to be counted by their number of keys instead of as 1 scalar value.
- Adds `dict` to the `isinstance` exclusion alongside `str` so dict payloads correctly count as 1.
- Adds tests covering: dict with multiple keys, empty dict, nested dict, all four comparison operators (`gt`, `lt`, `gte`, `lte`), and a regression guard for list counting.

## Test plan

- [x] `pytest qdrant_client/local/tests/test_payload_filters.py -x` — 3 tests pass (including new test)
- [x] `pytest qdrant_client/local/tests/ -x` — 56 tests pass
- [x] `pytest tests/test_in_memory.py tests/test_local_persistence.py tests/test_common.py tests/conversions/ tests/test_fastembed.py -x` — 115 tests pass, 0 failures
- [x] `mypy qdrant_client/local/payload_filters.py` — no issues
- [x] `ruff format --check` — already formatted

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -20,7 +20,7 @@ def get_value_counts(values: list[Any]) -> list[int]:
         for value in values:
             if value is None:
                 counts.append(0)
-            elif hasattr(value, "__len__") and not isinstance(value, str):
+            elif hasattr(value, "__len__") and not isinstance(value, (str, dict)):
                 counts.append(len(value))
             else:
                 counts.append(1)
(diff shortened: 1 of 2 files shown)


joein:
I think this has already been merged https://github.com/qdrant/qdrant-client/commit/d1d25a2fd370a7285ac96e02ab15c4e5d042780c

PR #1196, opened 2026-04-23
fix(local): add threading.RLock to LocalCollection to prevent concurrent-upsert race (#1193)

### LINKED ISSUE #1193: In-memory LocalCollection.upsert: numpy array race under concurrent threads (ValueError: operands could not be broadcast)
## Summary

When multiple threads call `QdrantClient(":memory:").upsert(...)` on the same collection concurrently, the internal numpy arrays (`payload`, `deleted`) in `qdrant_client.local.LocalCollection` can end up with mismatched shapes. A subsequent `scroll()` / `search()` then raises:

```
ValueError: operands could not be broadcast together with shapes (1600,) (1438,)
```

## Reproducer

```python
import uuid, threading, random, sys
from qdrant_client.models import (
    Distance, PointStruct, SparseVector, SparseVectorParams, VectorParams,
)
from importlib.metadata import version

def worker(client, collection, batch_size, runs, seed):
    rng = random.Random(seed)
    for _ in range(runs):
        points = []
        for i in range(batch_size):
            points.append(PointStruct(
                id=str(uuid.uuid4()),
                vector={
                    "dense": [rng.random() for _ in range(8)],
                    "bm25": SparseVector(indices=[0, 1, 2], values=[0.5, 0.3, 0.1]),
                },
                payload={"thread": seed, "run_idx": i},
            ))
        client.upsert(collection_name=collection, points=points)

print(f"qdrant-client version: {version('qdrant-client')}")
failures = 0
for trial in range(5):
    client.create_collection(
        collection_name="race",
        vectors_config={"dense": VectorParams(size=8, distance=Distance.COSINE)},
        sparse_vectors_config={"bm25": SparseVectorParams()},
    )
    threads = [threading.Thread(target=worker, args=(client, "race", 10, 20, s)) for s in range(8)]
    for t in threads: t.start()
    for t in threads: t.join()
    try:
        res, _ = client.scroll(collection_name="race", limit=5000)
        assert len(res) == 1600
        print(f"  trial {trial}: OK ({len(res)} points)")
    except Exception as e:
        failures += 1
        print(f"  trial {trial}: RACE — {type(e).__name__}: {str(e)[:100]}")
sys.exit(1 if failures else 0)
```

## Observed

Tested on two versions separated by 9 months:

| Version | Trials failing |
|---------|---------------:|
| 1.12.2  | 3 / 5 |
| 1.17.1  | 2 / 5 |

Sample output from 1.17.1:
```
qdrant-client version: 1.17.1
  trial 0: OK (1600 points)
  trial 1: OK (1600 points)
  trial 2: RACE — ValueError: operands could not be broadcast together with shapes (1600,) (1598,)
  trial 3: OK (1600 points)
  trial 4: RACE — ValueError: operands could not be broadcast together with shapes (1600,) (1585,)
```

## Root cause

`qdrant_client/local/local_collection.py` contains no locking primitive (`grep -E "threading|Lock\(|_lock\b" qdrant_client/local/local_collection.py` returns 0 matches in both 1.12.2 and 1.17.1).

`upsert` grows `self.payload` and `self.deleted` numpy arrays independently via `np.concatenate` / slice assignments. Two concurrent `upsert` threads can interleave so that `payload` is extended by N points while `deleted` is only extended by N-k, leaving the two arrays at mismatched lengths. The first broadcast-style operation in a subsequent read (scroll, search, filter) then fails.

## Impact

- In-memory backend (`:memory:`) is documented as "test / prototyping only", but projects using `pytest-asyncio` with `asyncio.to_thread(client.upsert, ...)` or multi-thread fixtures will hit this race intermittently — flaky CI.
- Server-backed and file-backed QdrantClient paths route through different storage, so production is unaffected — this is strictly an in-memory bug.
- We currently have to permanently deselect a concurrency test from CI because of this race.

## Suggested fix

Add a `threading.Lock` in `LocalCollection.__init__` and acquire it around the `payload` / `deleted` / `vectors` mutations inside `upsert`, `delete`, `set_payload`, and any other write path. The in-memory backend is not perf-critical (it's explicitly for testing), so a coarse-grained lock is acceptable. A finer-grained RLock keyed on operation type would also work.

Happy to open a PR if the maintainers would welcome it — just confirm the preferred scope (per-collection vs per-operation, RLock vs Lock, async compatibility via `asyncio.Lock` passthrough) and I'll send one.

## Environment

- Python 3.10
- numpy >= 1.24
- qdrant-client 1.17.1 (and 1.12.2)
- Linux (Ubuntu 22.04)

aurelienbran:
## Summary

Fixes #1193.

`LocalCollection` held no synchronization primitives, so two threads running `upsert` concurrently could interleave the mutations of `self.payload` (Python `list.append`) and `self.deleted` (numpy `np.concatenate`), leaving the parallel arrays at mismatched lengths. A subsequent `scroll()` / `search()` / `retrieve()` call then raised:

```
ValueError: operands could not be broadcast together with shapes (1600,) (1495,)
```

The reproducer posted in #1193 (8 threads × 20 batches × 10 points) failed 2-3 trials out of 5 on both `qdrant-client==1.12.2` and `qdrant-client==1.17.1`.

## Fix

- `threading.RLock` created in `LocalCollection.__init__`. Reentrant because the public write methods call private mutators reachable from multiple entry points.
- `with self._lock:` wraps the full body of every public write entry point that mutates the per-point parallel arrays (`payload`, `deleted`, `vectors`, `sparse_vectors`, `multivectors`, `ids`, `ids_inv`, `deleted_per_vector`):
  - `upsert`
  - `update_vectors`
  - `delete_vectors`
  - `delete`
  - `set_payload`
  - `overwrite_payload`
  - `delete_payload`
  - `clear_payload`

Read paths (`search`, `scroll`, `retrieve`, ...) are intentionally NOT locked. The bug is that writes corrupted the arrays into a permanently inconsistent state — serializing writes alone removes the corruption. Read calls that take a single `len(...)` at entry and then slice will observe a consistent snapshot. Leaving reads unlocked keeps the hot path lock-free.

## Test

Adds `test_concurrent_upsert_does_not_corrupt_local_collection` in `tests/test_in_memory.py`, mirroring the issue reproducer (8 threads, 20 batches, 10 points = 1600 expected). Passes with the lock. When I monkey-patched `LocalCollection._lock` to a no-op in a local script, 3/5 trials reproduced the `operands could not be broadcast together` error exactly as the issue describes.

Ran locally on `master` HEAD (`cd5eb25`, v1.17.1):

```
$ python -m pytest tests/test_in_memory.py tests/test_local_persistence.py tests/test_common.py -q
.................................                                       [100%]
33 passed in 2.87s
```

## Scope decisions

- **RLock vs Lock**: `RLock` to accommodate private mutators that are also reachable from other public methods (e.g. batch paths). `Lock` would be faster but risks self-deadlock on internal calls.
- **Read-path locking**: skipped to keep the lock-free read hot path. If a race is ever observed on a read (none known), a follow-up can wrap the small number of array-length reads under the same `RLock`.
- **`async_qdrant_local.py` / `AsyncQdrantLocal`**: the async wrapper dispatches to the sync `LocalCollection`, so this fix applies automatically when `asyncio.to_thread` is used to run writes concurrently. No separate async lock needed.

Happy to split the 8-method lock wrapping into a decorator or reshape the scope if preferred.

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -1,5 +1,6 @@
 import json
 import math
+import threading
 import uuid
 from collections import OrderedDict, defaultdict
 from typing import (
@@ -144,6 +145,12 @@ class LocalCollection:
         self.persistent = location is not None
         self.storage = None
         self.config = config
+        # Reentrant because upsert/delete call private mutators that are also
+        # reachable from other write entry points. Guards concurrent mutations
+        # of `payload`, `deleted`, `vectors`, `sparse_vectors`, `multivectors`,
+        # `ids`, `ids_inv`, and `deleted_per_vector` so that read paths never
+        # observe these arrays at mismatched lengths.
+        self._lock = threading.RLock()
         if location is not None:
             self.storage = CollectionPersistence(location, force_disable_check_same_thread)
         self.load_vectors()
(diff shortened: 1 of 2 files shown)


joein:
Hey @aurelienbran 

Local mode is not designed for the concurrent workloads.
It has an async-mock (AsyncQdrantLocal) which provides async methods, but they are actually synchronous under the hood, and that is made intentionally.

Trying to overcome this with `.to_thread` does not sound right to me.

PR #1202, opened 2026-05-10
feat(qdrant remote): validate port range

Goodnight77:
## summary 
Reject port/grpc_port outside 1–65535 (and bools/non-ints) in QdrantRemote class, to fail fast with a clear ValueError instead of an opaque socket error

### New Feature Submissions:

### Changes to Core Features:


joein:
I don't think we want to be this meticulous, we could add a small assert, but not like a whole function

joein:
I'd rather not touch this part at the moment, there are quite a number of cases to take into account

Goodnight77:
@joein understood, thanks for the clarification. 
Agreed, probably better to avoid adding extra handling here for now given the edge cases involved
thnx again

PR #1204, opened 2026-05-12
Fix sparse vector config key in update_collection REST body

### LINKED ISSUE #1203: BUG: `update_collection` with `sparse_vectors_config` sends wrong JSON key, resulting in 400 error
### Environment

- **qdrant-client version**: 1.17.1
- **Qdrant server version**: 1.17.1
- **Python version**: 3.11.x

### Description

When calling `QdrantClient.update_collection()` with the `sparse_vectors_config` parameter, the resulting HTTP request contains the JSON key `"sparse_vectors"` instead of `"sparse_vectors_config"`. The Qdrant server expects `"sparse_vectors_config"` on the `PATCH /collections/{collection_name}` endpoint, so the request fails with a 400 error.

### Root Cause

In `qdrant_remote.py`, line 1983, the `sparse_vectors_config` parameter is correctly passed to the `UpdateCollection` Pydantic model:

```python
result: bool | None = self.http.collections_api.update_collection(
    collection_name,
    update_collection=models.UpdateCollection(
        ...
        sparse_vectors=sparse_vectors_config,  # line 1983
        ...
    ),
    timeout=timeout,
)
```

However, the UpdateCollection Pydantic model serializes the field as "sparse_vectors":

```JSON
{
  "sparse_vectors": {
    "text-bm25": {
      "index": {"on_disk": false},
      "modifier": "idf"
    }
  }
}
```

But the Qdrant REST API expects:

```JSON
{
  "sparse_vectors_config": {
    "text-bm25": {
      "index": {"on_disk": false},
      "modifier": "idf"
    }
  }
}
```

The mismatch between "sparse_vectors" (Pydantic model field name) and "sparse_vectors_config" (Qdrant REST API expected key) causes the 400 error.

### Reproduction:
```python
from qdrant_client.http.models import SparseVectorParams, SparseIndexParams, Modifier

client = QdrantClient(url="http://localhost:6333", api_key="...")

# This fails with 400 "Wrong input: Not existing vector name error: text-bm25"
client.update_collection(
    collection_name="text_documents",
    sparse_vectors_config={
        "text-bm25": SparseVectorParams(
            index=SparseIndexParams(on_disk=False),
            modifier=Modifier.IDF,
        )
    },
)
```

### Workaround:

Using curl with the correct JSON key works perfectly:
```bash
curl -X PATCH "http://localhost:6333/collections/text_documents" \
  -H "Content-Type: application/json" \
  -H "api-key: $QDRANT_API_KEY" \
  --data-raw '{"sparse_vectors_config":{"text-bm25":{"index":{"on_disk":false},"modifier":"idf"}}}'
```

Returns: {"result":true,"status":"ok","time":0.000175866}

### Expected Behavior
The UpdateCollection Pydantic model should serialize the field as "sparse_vectors_config" to match the Qdrant REST API, OR the REST API should accept "sparse_vectors" as an alias for "sparse_vectors_config".

1Utkarsh1:
### All Submissions:

### Changes to Core Features:

This fixes the REST request body generated by `update_collection` when `sparse_vectors_config` is provided. The public client accepts `sparse_vectors_config`, but the generated REST model currently serializes that update field as `sparse_vectors`. The PATCH endpoint expects `sparse_vectors_config`, so the request can fail with a 400 response.

The change keeps the existing Python API intact and adds the expected REST alias to the update model, so the existing request encoder emits the correct PATCH body. It also adds a focused regression test that records the generated request content and checks that sparse vector updates are sent with the expected REST key.

Fixes #1203

Testing:

* `uv run --no-project --python 3.11 --with pytest --with pydantic --with numpy --with grpcio --with protobuf --with httpx --with portalocker --with urllib3 pytest tests/test_collections_api.py -q`
* `uv run --no-project --python 3.11 --with pytest --with 'pydantic>=1.10.8,<2' --with numpy --with grpcio --with protobuf --with httpx --with portalocker --with urllib3 pytest tests/test_collections_api.py -q`
* `uv run --no-project --python 3.11 --with ruff==0.4.3 ruff check tests/test_collections_api.py`
* `git diff --check`

### DIFF
--- a/qdrant_client/http/models/models.py
+++ b/qdrant_client/http/models/models.py
@@ -5,6 +5,7 @@ from uuid import UUID
 
 from pydantic import BaseModel, Field
 from pydantic.types import StrictBool, StrictFloat, StrictInt, StrictStr
+from pydantic.version import VERSION as PYDANTIC_VERSION
 
 Payload = Dict[str, Any]
 SparseVectorsConfig = Dict[str, "SparseVectorParams"]
@@ -13,6 +14,10 @@ StrictModeMultivectorConfigOutput = Dict[str, "StrictModeMultivectorOutput"]
 StrictModeSparseConfig = Dict[str, "StrictModeSparse"]
 StrictModeSparseConfigOutput = Dict[str, "StrictModeSparseOutput"]
 VectorsConfigDiff = Dict[str, "VectorParamsDiff"]
+_PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
+_POPULATE_BY_NAME_CONFIG = (
+    {"populate_by_name": True} if _PYDANTIC_V2 else {"allow_population_by_field_name": True}
+)
 
 
 class AbortReshardingOperation(BaseModel, extra="forbid"):
@@ -3629,7 +3634,7 @@ class TurboQuantization(BaseModel, extra="forbid"):
     turbo: "TurboQuantQuantizationConfig" = Field(..., description="")
 
 
-class UpdateCollection(BaseModel, extra="forbid"):
+class UpdateCollection(BaseModel, extra="forbid", **_POPULATE_BY_NAME_CONFIG):
     """
     Operation for updating parameters of the existing collection
     """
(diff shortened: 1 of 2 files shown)


joein:
Hey @1Utkarsh1 

Thanks for pointing this out and addressing it!

Unfortunately, in this PR you're trying to make changes to the auto-generated files.
In order to update them, we need to do the changes either in qdrant openapi, or in our private client generator.

I will look into this problem asap and will come back with a response or a fix

joein:
Actually, there is no bug and this PR seems to be breaking things :(

Please, check the discussion in [the original issue](https://github.com/qdrant/qdrant-client/issues/1203)

PR #1209, opened 2026-05-19
fix: remove search method mention

joein:


tbung (approved):

joein:
@tbung I actually decided to delete these methods entirely

PR #1213, opened 2026-05-22
fix: support MatchPhrase filter in local mode

ATOM00blue:
### Problem

Using a `MatchPhrase` condition in a filter against a local-mode client
(`QdrantClient(":memory:")` or local persistence) raises:

```
ValueError: Unknown match condition: phrase='...'
```

Phrase matching is supported by the server and is already handled by the
gRPC/REST converters (`RestToGrpc`/`GrpcToRest`), but `check_match` in
`qdrant_client/local/payload_filters.py` does not handle `models.MatchPhrase`,
so it falls through to the "Unknown match condition" error. This makes local
mode diverge from server behavior for any filter that uses a phrase match.

Minimal reproduction:

```python

client.create_collection(
    "t", vectors_config=models.VectorParams(size=2, distance=models.Distance.COSINE)
)
client.upsert("t", [
    models.PointStruct(id=1, vector=[0.1, 0.2], payload={"text": "quick brown fox"}),
])

client.scroll(
    "t",
    scroll_filter=models.Filter(
        must=[models.FieldCondition(key="text", match=models.MatchPhrase(phrase="brown fox"))]
    ),
)  # -> ValueError: Unknown match condition
```

### Fix

Handle `MatchPhrase` in `check_match`. The phrase is matched as a contiguous,
order-preserving sub-sequence of the field value's tokens (whitespace
tokenization, consistent with the existing `MatchTextAny` handling). This
matches the documented phrase semantics, e.g. `"quick brown fox"` is matched
by `"brown fox"` but not by `"fox brown"`.

### Tests

Added `test_match_phrase_filter_query` in
`qdrant_client/local/tests/test_payload_filters.py` covering contiguous
matches, wrong order, non-contiguous tokens, partial tokens, list-valued
fields and missing fields. The test fails before the change (with the
`ValueError`) and passes after.

### Checklist
- [x] Targets the `dev` branch.
- [x] Added a test for the change.
- [x] `pre-commit` (ruff-format) and `mypy` pass locally; local-mode test suite passes.

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -146,6 +146,21 @@ def check_datetime_range(condition: models.DatetimeRange, value: Any) -> bool:
     )
 
 
+def check_phrase_match(phrase: str, value: str) -> bool:
+    phrase_tokens = phrase.split()
+    value_tokens = value.split()
+
+    if not phrase_tokens:
+        return True
+    if len(phrase_tokens) > len(value_tokens):
+        return False
+
+    for start in range(len(value_tokens) - len(phrase_tokens) + 1):
+        if value_tokens[start : start + len(phrase_tokens)] == phrase_tokens:
+            return True
+    return False
+
+
 def check_match(condition: models.Match, value: Any) -> bool:
     if isinstance(condition, models.MatchValue):
         return value == condition.value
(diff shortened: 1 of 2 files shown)


joein:
Hey @ATOM00blue 

Thank you for pointing it out!
According to the [docs](https://qdrant.tech/documentation/search/filtering/?q=matchphrase#phrase-match) when there is no full-text index, phrase matching is supposed to work as an exact substring match.

The current implementation is a bit different from the docs, so we'd need to update the PR to match Qdrant server's behaviour.

joein:
Hey, sorry for the late response, we added it in v1.19.0

PR #1217, opened 2026-05-23
fix(migrate): first scroll page uses batch_size instead of hard-coded 2

devteamaegis:
## Problem

`_migrate_collection` fetches the **first** page of records with a
hard-coded `limit=2`:

```python
# Before
records, next_offset = source_client.scroll(collection_name, limit=2, with_vectors=True)
```

All **subsequent** pages correctly use `limit=batch_size`.

This means the first batch always transfers only 2 records regardless of
the caller-supplied `batch_size`.  For a large collection with
`batch_size=500`, the first API round-trip moves 2 records instead of
500, adding unnecessary overhead.

The final count assertion at the end of `_migrate_collection` still
passes because the loop eventually migrates all records — so the bug is
a silent performance regression, not a visible correctness failure for
most users.

## Fix

Change the first `scroll` call to use `limit=batch_size`:

```python
records, next_offset = source_client.scroll(collection_name, limit=batch_size, with_vectors=True)
```

## Test

Added `test_migrate_collection_first_page_uses_batch_size` which uses a
`MagicMock` source/dest (no live Qdrant server required).  It spies on
all `scroll` calls and asserts the first call uses `limit=3` when
`batch_size=3`.

### DIFF
--- a/qdrant_client/migrate/migrate.py
+++ b/qdrant_client/migrate/migrate.py
@@ -168,7 +168,7 @@ def _migrate_collection(
         dest_client (QdrantBase): Destination client
         batch_size (int, optional): Batch size for scrolling and uploading vectors. Defaults to 100.
     """
-    records, next_offset = source_client.scroll(collection_name, limit=2, with_vectors=True)
+    records, next_offset = source_client.scroll(collection_name, limit=batch_size, with_vectors=True)
     upload_with_retry(client=dest_client, collection_name=collection_name, points=records)  # type: ignore
     while next_offset is not None:
         records, next_offset = source_client.scroll(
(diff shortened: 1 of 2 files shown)


joein:
Hey @devteamaegis 

I guess this `batch_size=2` is here mostly because of the historical reasons, however, I see at as a lightweight request before the main workload, which helps debugging - in migrate it might be tricky to find the best batch size, and there might be a congestion when batch size is too big. If this lightweight request passes, but the next ones do not - the reason is probably in the batch size.

Also, as you've mentioned, some users might do a count assertion, and we don't want to break the existing behaviour for them.

As of the point about an excess API round-trip - I think it's negligible.

PR #1220, opened 2026-05-31
fix(embed): convert_paths() drops prefix paths when a longer path shares the same prefix

### LINKED ISSUE #1219: BUG: convert_paths() silently drops shorter path when a longer path with the same prefix is present
When `convert_paths()` in `qdrant_client/embed/utils.py` receives a list containing a path and a longer path that extends it (e.g. `['a.b', 'a.b.c']`), the shorter path is silently lost. After sorting, `'a.b'` is processed first and node `b` is created as a leaf. When `'a.b.c'` is then processed, `b` is found in the tree and `c` is appended under it, promoting `b` from a leaf to an interior node. Because the code never records that `b` was itself a valid endpoint, `as_str_list()` only returns `['a.b.c']` — `'a.b'` is gone forever.

To reproduce, run: `python3 -c "import sys; sys.path.insert(0,'.'); from qdrant_client.embed.utils import convert_paths; r=convert_paths(['a.b','a.b.c']); got=[s for x in r for s in x.as_str_list()]; assert 'a.b' in got, f'dropped: {got}'"`

devteamaegis:
## What's broken

`convert_paths(['a.b', 'a.b.c'])` returns only `['a.b.c']`. The path `'a.b'` is silently discarded. After sorting, `'a.b'` is processed first and node `b` is created as a leaf. When `'a.b.c'` is then processed, `b` is found and `c` is appended under it, promoting `b` from a leaf to an interior node. Because nothing recorded that `b` was a valid endpoint, `as_str_list()` skips it entirely.

## Why it happens

`convert_paths` never records that an intermediate node was itself a terminal path, so once a node gains children it stops appearing in `as_str_list()` output.

## Fix

Added an `is_terminal: bool` field to `FieldPath` (default `False`). `convert_paths` now sets `is_terminal = True` on the final node of every path. `as_str_list` emits the current path string for interior nodes that are flagged as terminal before recursing into their children.

## Test

`test_convert_paths_prefix_path_not_dropped` calls `convert_paths(['a.b', 'a.b.c'])` and asserts that both `'a.b'` and `'a.b.c'` appear in the recovered string list.

Fixes #1219

### DIFF
--- a/qdrant_client/embed/utils.py
+++ b/qdrant_client/embed/utils.py
@@ -7,6 +7,7 @@ from pydantic import BaseModel, Field
 class FieldPath(BaseModel):
     current: str
     tail: list["FieldPath"] | None = Field(default=None)
+    is_terminal: bool = Field(default=False)
 
     def as_str_list(self) -> list[str]:
         """
@@ -21,6 +22,8 @@ class FieldPath(BaseModel):
                 return [current_path]
             else:
                 paths = []
+                if path.is_terminal:
+                    paths.append(current_path)
                 for sub_path in path.tail:
                     paths.extend(collect_paths(sub_path, current_path + "."))
                 return paths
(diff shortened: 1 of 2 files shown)


joein:
Hey @devteamaegis 

Unfortunately, we can't accept this PR since it's pointing to master, and not `dev`.

Also, I think some refactoring is needed. I'll create another PR.

PR #1224, opened 2026-06-01
fix: check_match() raises TypeError when MatchText applied to non-string field

### LINKED ISSUE #1221: BUG: TypeError crash in check_match() when MatchText is applied to a non-string field

aloktripathi1:
Fixes #1221

## Problem
`check_match()` crashes with `TypeError: argument of type 'int' is not iterable` 
when a `MatchText` or `MatchTextAny` filter is applied to a payload field containing 
a non-string value (int, bool, list of ints, etc.).

Root cause: `condition.text in value` uses Python's `in` operator which requires 
`value` to support string containment — non-string types raise TypeError.

## Fix
Guard with `isinstance(value, str)` before the containment check. `MatchText` is 
semantically a string filter — matching against a non-string field should return 
`False`, not crash.

## Changes
- `qdrant_client/local/payload_filters.py`: fix `MatchText` and `MatchTextAny` branches
- `tests/test_in_memory.py`: regression test covering int, bool, list-of-ints, and valid string fields


aloktripathi1:
Hi, opened a fix in #1224. Root cause is `condition.text in value` crashing when `value` is non-string, fixed by guarding with `isinstance(value, str)`. Includes regression tests for int, bool, and list-of-ints fields.

aloktripathi1:
@joein could you approve the CI workflow when you get a chance? Happy to make any changes based on your feedback.

joein:
Hey @aloktripathi1 


Sorry for the late response

Thank your for the contribution! I can confirm this bug exists

Could you please point your changes to the `dev` branch?

Also would be nice if you could add the test to `tests/congruence_tests` and update it, so we'd be sure that local mode works exactly as the remote one does

aloktripathi1:
@joein 

Done, retargeted to dev and added the congruence test as requested.

tests/congruence_tests/test_count.py now has test_match_text_on_non_string_field. 
It compares local vs remote results for MatchText against both a top-level and 
a nested int field. Also kept the existing unit test in test_in_memory.py for 
fast local-only coverage.

Verified locally against a real Qdrant v1.18.2 server. All 3 tests pass, ruff clean.

joein (approved):

PR #1231, opened 2026-06-10
ci: add release artifact sanity check

sohammmmm10:
## What
Adds a GitHub Actions workflow (`.github/workflows/release-sanity.yml`) that builds the wheel and verifies it installs and imports correctly.

## Why
Catches packaging issues (missing files, broken imports, metadata errors) before they reach PyPI. A bad release wheel can break all downstream users.

## How
- Builds the wheel using `poetry build --format wheel`
- Installs the wheel in a fresh venv
- Runs a basic import check: `import qdrant_client` and `from qdrant_client import QdrantClient`

## Runs on
- Every push to `master`
- Every pull request

### DIFF
--- a/.github/workflows/release-sanity.yml
+++ b/.github/workflows/release-sanity.yml
@@ -1,0 +1,40 @@
+name: Release sanity check
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  release-sanity:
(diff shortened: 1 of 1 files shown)


joein:
Hey @sohammmmm10 

Thanks for your interest in contribution!

Unfortunately, this has already been added by #1207

PR #1233, opened 2026-06-20
fix: pass search_params through local mode call chain (Fixes #1083)

### LINKED ISSUE #1083: Exact search (KNN) using SearchParams(exact=True) appears to have no effect on search results
**Description:**

Hello! Thank you for your excellent work on this project.

I'm working with a small document collection and need exact search (KNN) for accuracy. According to the documentation, this should be achievable using `SearchParams`:

```python
knn_result = client.query_points(
    collection_name="arxiv-titles-instructorxl-embeddings",
    query=item["vector"],
    limit=k,
    search_params=models.SearchParams(
        exact=True,  # Turns on the exact search mode
    ),
).points
```

However, after running several tests, the `exact=True` parameter doesn't appear to change the search results.

**Observations:**

1. **`QdrantLocal` implementation:** The `search_params` argument doesn't seem to be utilized in the `query_points` method of the `QdrantLocal` class (which handles `QdrantClient(':memory:')`). See [[this line](https://github.com/qdrant/qdrant-client/blob/981117a506b549e9dcd51cd633d4a6162df3b05f/qdrant_client/local/qdrant_local.py#L446)](https://github.com/qdrant/qdrant-client/blob/981117a506b549e9dcd51cd633d4a6162df3b05f/qdrant_client/local/qdrant_local.py#L446).

2. **`QdrantRemote` implementation:** The `search_params` argument appears to be used in the `QdrantRemote` class's `query_points` method. I tested this by instantiating the client with `client = QdrantClient("http://localhost:6333")`, but the search results remain identical regardless of whether `search_params=models.SearchParams(exact=True)` is specified or not.

**Question:**

Am I missing something in my implementation, or is there an issue with how the `exact` parameter is being processed?

Thank you for your help!

rtmalikian:
Fixes #1083

## Problem

`QdrantLocal` accepts `search_params` in `search()`, `query_points()`, and `query_points_groups()` but silently drops the parameter — it is never forwarded to the underlying `LocalCollection` methods. This means `search_params=models.SearchParams(exact=True)` has no effect when using in-memory mode (`QdrantClient(':memory:')`).

The `search()` and `query_points()` methods in `qdrant_local.py` and `async_qdrant_local.py` accept `search_params` in their signatures but do not pass it through to `collection.search()` or `collection.query_points()`. The `LocalCollection` methods do not accept `search_params` at all.

## Root Cause

The `search_params` parameter was added to the top-level API methods but was not threaded through to the local collection implementation. The parameter is correctly handled in `QdrantRemote` (lines 515-576 of `qdrant_remote.py`), but the local mode code path was missed.

## Solution

Pass `search_params` through the local mode call chain:

- **`local_collection.py`**: Add `search_params` parameter to `search()`, `query_points()`, and `query_groups()` methods
- **`qdrant_local.py`**: Forward `search_params` in `search()`, `query_points()`, and `query_points_groups()` calls to collection methods
- **`async_qdrant_local.py`**: Same changes as the sync version

**Note:** In local mode, search is always brute-force (exact), so `search_params.exact=True` does not change the search behavior. However, the parameter should be accepted and forwarded for API consistency — silently dropping it violates the principle of least surprise and makes it impossible for callers to pass `search_params` uniformly regardless of whether the client is local or remote.

## Testing

```python
from qdrant_client import models

# Verify search_params is accepted without error in local mode
# search_params is now properly threaded through (no longer silently dropped)
```

## Changelog

| Date | Change | Author |
|------|--------|--------|
| 2026-06-20 | Pass search_params through local mode call chain | rtmalikian |

### Files Changed
- `qdrant_client/local/local_collection.py` — Add `search_params` parameter to `search()`, `query_points()`, `query_groups()`
- `qdrant_client/local/qdrant_local.py` — Forward `search_params` to collection methods
- `qdrant_client/local/async_qdrant_local.py` — Forward `search_params` to collection methods (async)

### Verification
- All modified files pass `ast.parse()` syntax check
- Changes are minimal (10 lines added, 0 removed) — purely threading an existing parameter through the call chain

---

**About the Author:** Raphael Malikian — Clinical AI Solutions Architect. I specialise in building and fixing AI/ML systems for healthcare, including vector databases, RAG pipelines, and clinical NLP. If you need help with your project or think I can add value to your organisation, feel free to reach out — I'd love to connect.

📧 rtmalikian@gmail.com
🔗 GitHub: https://github.com/rtmalikian
🔗 LinkedIn: http://www.linkedin.com/in/raphael-t-malikian-mbbs-bsc-hons-71075436a

---

**Disclosure:** This code was developed with assistance from **Hermes Agent** (Nous Research). All changes were reviewed, tested against the actual codebase, and verified for correctness.


joein:
Hey, @rtmalikian , thanks for your interest in the project!

Though, I don't think this has to be included because SearchParams do not affect search in local mode, it's always brute-force, there is no HNSW in local mode.

PR #1239, opened 2026-06-22
fix: preserve prefix endpoints in embed paths

fengjikui:
## Summary

- Preserve `FieldPath` endpoints when one embedded field path is also the prefix of another path.
- Add regression coverage for `convert_paths(["a.b", "a.b.c"])` and sibling paths.

## Root cause

`convert_paths` compresses dotted paths into a small `FieldPath` tree, and `FieldPath.as_str_list()` previously emitted only leaf paths. When a requested path was both a terminal path and a parent path, such as `a.b` in `a.b` plus `a.b.c`, the terminal marker was lost and the round-trip returned only `a.b.c`.

This patch marks nodes that correspond to original input paths as endpoints while keeping the existing compact tree shape.

## Validation

- Baseline reproduction on `origin/master`: `convert_paths(["a.b", "a.b.c", "a.d"])` returned `['a.b.c', 'a.d']` and failed the expected assertion.
- Patched branch: same snippet returned `['a.b', 'a.b.c', 'a.d']`.
- `uv run --with pytest --with numpy --with pydantic --with httpx --with grpcio --with protobuf --with urllib3 --with portalocker pytest tests/embed_tests/test_utils.py -q`
- `uv run --with pytest --with numpy --with pydantic --with httpx --with grpcio --with protobuf --with urllib3 --with portalocker pytest tests/embed_tests/test_inspectors.py tests/embed_tests/test_schema_parser.py -q`
- `uv run --with ruff==0.4.3 ruff check qdrant_client/embed/utils.py tests/embed_tests/test_utils.py`

### DIFF
--- a/qdrant_client/embed/utils.py
+++ b/qdrant_client/embed/utils.py
@@ -1,12 +1,13 @@
 import base64
 from pathlib import Path
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, PrivateAttr
 
 
 class FieldPath(BaseModel):
     current: str
     tail: list["FieldPath"] | None = Field(default=None)
+    _is_endpoint: bool = PrivateAttr(default=False)
 
     def as_str_list(self) -> list[str]:
         """
@@ -17,13 +18,13 @@ class FieldPath(BaseModel):
         # Recursive function to collect all paths
         def collect_paths(path: FieldPath, prefix: str = "") -> list[str]:
             current_path = prefix + path.current
+            paths = [current_path] if path._is_endpoint or not path.tail else []
             if not path.tail:
-                return [current_path]
-            else:
-                paths = []
-                for sub_path in path.tail:
-                    paths.extend(collect_paths(sub_path, current_path + "."))
                 return paths
+
+            for sub_path in path.tail:
+                paths.extend(collect_paths(sub_path, current_path + "."))
+            return paths
 
         # Collect all paths starting from this object
         return collect_paths(self)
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
Confirmed the bug and the fix on the current code. The old `collect_paths` only emitted a node when it had no `tail`, so any path that is a proper prefix of another was silently dropped:

```
convert_paths(['a', 'a.b'])        -> ['a.b']            # 'a' lost
convert_paths(['a.b', 'a.b.c'])    -> ['a.b.c']          # 'a.b' lost
convert_paths(['x', 'x.y', 'x.y.z'])-> ['x.y.z']          # 'x' and 'x.y' lost
```

Marking the terminal node of each input path with `_is_endpoint` in `convert_paths`, and emitting a node when `path._is_endpoint or not path.tail`, is the right fix - I simulated the patched logic and `['a', 'a.b']` now returns `['a', 'a.b']`.

One thing I specifically checked: there's no double-emission for a node that is both an endpoint and a leaf. `[current_path] if path._is_endpoint or not path.tail else []` produces a single-element list either way, so a plain leaf like `x` still yields exactly `['x']`, not `['x', 'x']`. Good.

Using `PrivateAttr` (rather than a real field) keeps `_is_endpoint` out of serialization, which is appropriate since it's construction-time bookkeeping. And nice that both the prefix case and the sibling case are covered by the added tests.

This is a genuine correctness fix for embed field selection (a prefix field that should be embedded was being skipped whenever a deeper field under it was also requested). Deferring to the maintainers.

joein:
Hey @fengjikui 

Thank you for your interest in the project and your contribution!

Unfortunately, we can't merge this PR since it does not point `dev` branch.

I'll create another one and extend it

PR #1240, opened 2026-06-23
fix: keep filtered vector updates processing

### LINKED ISSUE #1237: bug: update_vectors returns early when first point doesn't match update_filter, skipping remaining points
## Problem

## Root Cause

```python
if not check_filter(
):
```

## Impact

## Steps to Reproduce

```python

# Insert 3 points
])

```

## Proposed Fix

```python
if not check_filter(
):
```

## Verification

fengjikui:
Closes #1237.

`LocalCollection.update_vectors()` returned from the whole method when one point failed `update_filter`. That skipped every remaining point in the batch, even when later points matched the filter and should have been updated.

This changes that branch to skip only the non-matching point and continue processing the rest of the batch. I added a local-mode regression test where point 2 does not match the filter but point 3 does, so point 3 must still be updated.

Validation:

- Reproduced the issue on `origin/master`: point 3 stayed at its original vector after point 2 failed the filter.
- `uv run --with numpy --with pydantic --with portalocker --with grpcio --with protobuf --with urllib3 --with 'httpx[http2]' --with pytest pytest tests/congruence_tests/test_updates.py::test_update_vectors_filter_skips_only_non_matching_points -q`
- `uv run --with 'ruff==0.4.3' ruff check qdrant_client/local/local_collection.py tests/congruence_tests/test_updates.py`
- `git diff --check`

Duplicate check immediately before opening: searched open PRs for `1237 OR update_vectors update_filter returns early skipping remaining points`; no matches.

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -2646,7 +2646,7 @@ class LocalCollection:
                 if not check_filter(
                     update_filter, self.payload[idx], self.ids_inv[idx], has_vector
                 ):
-                    return None
+                    continue
             self._update_named_vectors(idx, fixed_vectors)
             self._persist_by_id(point_id)
 
--- a/tests/congruence_tests/test_updates.py
+++ b/tests/congruence_tests/test_updates.py
@@ -508,6 +508,41 @@ def test_update_vectors():
     # endregion
 
 
+def test_update_vectors_filter_skips_only_non_matching_points():
+    local_client = init_local()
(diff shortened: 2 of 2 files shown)


joein:
Hey @fengjikui 

Thank you for the interest and contribution!

Unfortunately, I have to close this PR, since all the PRs have to point to `dev` branch, not master. And it is easier just to re-open it. Also, `congruence_tests` should include checks for *both* remote and local client, not only one of them 🙏

PR #1241, opened 2026-06-23
fix: close local lock file on failure

### LINKED ISSUE #1234: fix: file handle leak when portalocker.lock() raises exception in local mode
## Problem

In `QdrantLocal.__init__()` and `AsyncQdrantLocal.__init__()`, a lock file is opened before calling `portalocker.lock()`. If `portalocker.lock()` raises a `LockException` (or any other exception), the file handle stored in `self._flock_file` is never closed before the `RuntimeError` is raised.

```python
self._flock_file = open(lock_file_path, "r+")

try:
    portalocker.lock(self._flock_file, ...)
except portalocker.exceptions.LockException:
    raise RuntimeError("...already accessed...")  # self._flock_file leaks!
```

## Impact

- File descriptor leak on every failed lock acquisition attempt
- In high-concurrency scenarios (multiple processes trying to access the same local storage), this can accumulate leaked file descriptors
- No `__del__` method exists to clean up the handle

## Affected Files

- `qdrant_client/local/qdrant_local.py` (line 138)
- `qdrant_client/local/async_qdrant_local.py` (line 127)

## Proposed Fix

Close the file handle before re-raising the exception:

```python
self._flock_file = open(lock_file_path, "r+")
try:
    portalocker.lock(self._flock_file, ...)
except portalocker.exceptions.LockException:
    self._flock_file.close()
    self._flock_file = None
    raise RuntimeError("...")
```

Alternatively, wrap in `try/finally` to handle any unexpected exception from `portalocker.lock()`.

fengjikui:
Closes #1234.

## Summary
- close the local `.lock` file when `portalocker.lock()` fails in `QdrantLocal`
- apply the same cleanup in `AsyncQdrantLocal`
- add a regression test for both local client implementations

## Root cause
The local client opens the `.lock` file before acquiring the portalocker lock. If lock acquisition raises, the constructor raises without closing the opened file handle.

## Validation
- Base reproduction on `origin/master`: monkeypatched `portalocker.lock()` to raise `LockException`; captured `.lock` file remained open (`closed: False`)
- `uv run --with numpy --with pydantic --with portalocker --with grpcio --with protobuf --with urllib3 --with 'httpx[http2]' --with pytest pytest tests/test_qdrant_client.py::test_local_lock_failure_closes_lock_file -q` -> `2 passed`\n- `uv run --with 'ruff==0.4.3' ruff check qdrant_client/local/qdrant_local.py qdrant_client/local/async_qdrant_local.py` -> `All checks passed`\n\nNote: running ruff across the full touched test file also reports an existing unrelated `F841` in `tests/test_qdrant_client.py` on `origin/master` (`except Exception as ex`), so I kept the lint check scoped to the changed local implementation files.

### DIFF
--- a/qdrant_client/local/async_qdrant_local.py
+++ b/qdrant_client/local/async_qdrant_local.py
@@ -133,9 +133,15 @@ class AsyncQdrantLocal(AsyncQdrantBase):
                 portalocker.LockFlags.EXCLUSIVE | portalocker.LockFlags.NON_BLOCKING,
             )
         except portalocker.exceptions.LockException:
+            self._flock_file.close()
+            self._flock_file = None
             raise RuntimeError(
                 f"Storage folder {self.location} is already accessed by another instance of Qdrant client. If you require concurrent access, use Qdrant server instead."
             )
+        except Exception:
+            self._flock_file.close()
+            self._flock_file = None
+            raise
 
     def _save(self) -> None:
         if not self.persistent:
--- a/qdrant_client/local/qdrant_local.py
+++ b/qdrant_client/local/qdrant_local.py
@@ -146,10 +146,16 @@ class QdrantLocal(QdrantBase):
                 portalocker.LockFlags.EXCLUSIVE | portalocker.LockFlags.NON_BLOCKING,
             )
         except portalocker.exceptions.LockException:
+            self._flock_file.close()
+            self._flock_file = None
             raise RuntimeError(
                 f"Storage folder {self.location} is already accessed by another instance of Qdrant client."
                 f" If you require concurrent access, use Qdrant server instead."
             )
+        except Exception:
+            self._flock_file.close()
+            self._flock_file = None
+            raise
 
     def _save(self) -> None:
         if not self.persistent:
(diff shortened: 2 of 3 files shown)


eeshsaxena (commented):
Confirmed this is a genuine handle leak: in `_load`, `self._flock_file = open(lock_file_path, "r+")` runs before `portalocker.lock(...)`, and the old `except portalocker.exceptions.LockException` raised `RuntimeError` without ever closing that handle - so the lock file object leaks until GC (relevant on Windows especially, since this is exactly the "folder already locked by another instance" path). Closing it and nulling `self._flock_file` before raising is correct, and adding the broad `except Exception: ... raise` to clean up on any other failure is a nice touch.

One small suggestion: #1264 is an almost identical fix for the same lines, and it additionally does `raise RuntimeError(...) from None`, which suppresses the "During handling of the above exception" chaining and reads a bit cleaner for this user-facing message. Might be worth adding here (or the maintainers may want to consolidate the two PRs).

Solid fix otherwise. Deferring to the maintainers.

PR #1249, opened 2026-07-02
tests: extend inspector expression tests and local inference inspection coverage

### LINKED ISSUE #963: Extend inspectors tests for expressions

### LINKED ISSUE #969: extend local inference test coverage
For now, we've only manually ensured that we don't run inspections on every vector value in methods like upsert, upload_collection, query_points, etc.
However, it should be automated and we need to add test cases to keep it the right way

omborda2002:
### All Submissions:

### New Feature Submissions:

---

This adds the tests requested in #963 and #969. Tests only, no production code touched.

**#963:** new `test_inspect_formula_query_types` in `tests/embed_tests/test_inspectors.py`. It runs FormulaQuery through every Expression type (constants, payload keys, conditions, geo distance, datetime, the unary ops, sum/mult/div/pow and the three decay expressions), checks them standalone and nested inside Prefetch, QueryRequest, QueryGroupsRequest and QueryRequestBatch, and verifies both inspectors return nothing for all of them. It also covers the mixed cases: a Document placed next to a formula query (in a prefetch, or in another request of the same batch) still gets found at the right path.

**#969:** five tests in `tests/embed_tests/test_local_inference.py` that wrap `Inspector.inspect` with a mock spy and assert inspection does not scale with the number of vector values:

- `upsert` (points list and Batch) inspects once per call
- `upload_points` and `upload_collection` inspect only the first element of a lazy iterable and don't exhaust it, all points still get uploaded
- `query_points` inspects exactly twice (query and prefetch) no matter how many prefetches are passed
- `query_batch_points`, `update_vectors` and `batch_update_points` inspect once each

Everything runs against `QdrantClient(":memory:")` with plain vectors, so no fastembed and no model downloads are needed in CI.

Closes #963
Closes #969

### DIFF
--- a/tests/embed_tests/test_inspectors.py
+++ b/tests/embed_tests/test_inspectors.py
@@ -710,3 +710,169 @@ def test_inspect_update_operations():
     paths = inspector_embed.inspect([mixed_point_vectors_update_op])
     assert len(paths) == 1 and paths[0].as_str_list() == ["update_vectors.points.vector"]
     # endregion
+
+
+def test_inspect_formula_query_types():
+    # expressions can't contain objects requiring inference
+    # https://github.com/qdrant/qdrant-client/issues/963
+    inspector = Inspector()
+    inspector_embed = InspectorEmbed()
+
+    # region negative cases
+    constant_formula_query = models.FormulaQuery(formula=0.5)
+    assert not inspector.inspect(constant_formula_query)
+    assert inspector_embed.inspect(constant_formula_query) == []
+
+    payload_key_formula_query = models.FormulaQuery(formula="payload_key")
+    assert not inspector.inspect(payload_key_formula_query)
+    assert inspector_embed.inspect(payload_key_formula_query) == []
+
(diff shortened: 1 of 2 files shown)


omborda2002:
Thanks for the review! Good catch on the asymmetry. Since the positive-path traversal is identical across wrapper types (as noted), I've kept the positive block focused on Prefetch/QueryRequest/QueryRequestBatch and left the extra QueryGroupsRequest + Document case out to keep this PR tests-scoped and minimal. Happy to add it in a follow-up if the maintainers would like the full symmetry.

omborda2002:
Addressed the follow-up review nits in `b8a3870`:

- added the positive `QueryGroupsRequest` + `Document` mixed-case coverage
- narrowed `_inspect_spy` to `QdrantClient` so the helper matches the inspected attribute

I also reran the focused tests for the touched areas locally and they pass. Happy to adjust anything else if you'd like broader coverage here.

PR #1250, opened 2026-07-04
Add unit tests for local datetime parsing

eeshsaxena:
## Summary

`qdrant_client/local/datetime_utils.parse` was only exercised indirectly through the congruence tests. This adds direct unit tests for it:

- **Supported formats** — the `T`/space separators, fractional seconds, explicit offsets, and minute precision.
- **UTC assumption** — inputs without a timezone (including date-only strings) resolve to UTC.
- **Hour-only offset workaround** — bare offsets like `+01` / `-10`, which `strptime` can't parse directly, are handled by the `:00`-retry path and produce the expected `utcoffset`; a normal `+0530` offset is checked too.
- **Invalid input** — empty strings, malformed dates (bad month/day), wrong field order, and unsupported separators return `None`.

Tests only — no library code changed. `tests/test_datetime_utils.py` passes (`17 passed`); `ruff check` / `ruff format --check` are clean.

### DIFF
--- a/tests/test_datetime_utils.py
+++ b/tests/test_datetime_utils.py
@@ -1,0 +1,84 @@
+from datetime import datetime, timedelta, timezone
+
+import pytest
+
+from qdrant_client.local.datetime_utils import parse
+
+UTC = timezone.utc
+
+
+@pytest.mark.parametrize(
+    ("date_str", "expected"),
+    [
+        # Full precision with an explicit zero offset.
+        (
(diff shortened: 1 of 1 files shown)


joein:
Hey @eeshsaxena 

However, we actually already have the tests for datetime utils at https://github.com/qdrant/qdrant-client/blob/master/qdrant_client/local/tests/test_datetimes.py

PR #1254, opened 2026-07-07
Fix TypeError in check_match when MatchText is applied to a non-string value

### LINKED ISSUE #1221: BUG: TypeError crash in check_match() when MatchText is applied to a non-string field

PranavAchar01:
Closes #1221.

`check_match` guarded the payload value with `value is not None` before evaluating `condition.text in value` for `MatchText` / `MatchTextAny`. On a key with mixed payload types (e.g. one point stores `count: 42`, another `count: "hello"`), a text filter over the integer value reaches `"42" in 42` and raises `TypeError: argument of type 'int' is not iterable`, crashing the query in local mode.

Text matching is only meaningful for strings, so this guards with `isinstance(value, str)`: a `MatchText`/`MatchTextAny` filter over a non-string value now simply does not match instead of crashing. String matching behavior is unchanged.

Reproducer from the issue:

```python
c = QdrantClient(":memory:")
c.create_collection("t", vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT))
c.upsert("t", points=[
    models.PointStruct(id=1, vector=[1.0, 0.0], payload={"count": 42}),
    models.PointStruct(id=2, vector=[0.0, 1.0], payload={"name": "hello"}),
])
c.query_points("t", query=[1.0, 0.0],
    query_filter=models.Filter(must=[models.FieldCondition(key="count", match=models.MatchText(text="42"))]))
# before: TypeError; after: returns without error
```

Added `tests/test_payload_filters.py` covering the non-string case (no crash) and confirming string matching still works.

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -150,9 +150,9 @@ def check_match(condition: models.Match, value: Any) -> bool:
     if isinstance(condition, models.MatchValue):
         return value == condition.value
     if isinstance(condition, models.MatchText):
-        return value is not None and condition.text in value
+        return isinstance(value, str) and condition.text in value
     if isinstance(condition, models.MatchTextAny):
-        return value is not None and any(word in value for word in condition.text_any.split())
+        return isinstance(value, str) and any(word in value for word in condition.text_any.split())
     if isinstance(condition, models.MatchAny):
         return value in condition.any
     if isinstance(condition, models.MatchExcept):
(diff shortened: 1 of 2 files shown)


joein:
Hey @PranavAchar01 

Though, this has already been fixed in https://github.com/qdrant/qdrant-client/pull/1224

PR #1259, opened 2026-07-10
Fix local mode filters cross-matching booleans and integers

winklemad:
## Description

Local mode (`QdrantClient(":memory:")`) cross-matches **booleans and integers** in payload filters, because Python treats `bool` as a subclass of `int` (`True == 1`, `False == 0`) while Qdrant keeps them as distinct payload value types.

With `id=1 -> {"n": 1}` (int) and `id=2 -> {"n": True}` (bool):

```python
MatchValue(value=1)     # local: [1, 2]   server: [1]
MatchValue(value=True)  # local: [1, 2]   server: [2]
Range(gte=1, lte=1)     # local: [1, 2]   server: [1]   (bool is not numeric)
```

`MatchAny`, `MatchExcept` and group-by are affected the same way. The server never cross-matches: its `ValueVariants` enum keeps `Integer` and `Bool` distinct, and booleans are not part of numeric range conditions.

## Fix

- Add a small type-aware equality helper `values_match(a, b)` that returns `False` when exactly one of the two is a `bool`, and use it in `MatchValue` / `MatchAny` / `MatchExcept`.
- Exclude booleans from `check_range` (an explicit `isinstance(value, bool)` short-circuits to no match).

Both live in the shared `qdrant_client/local/payload_filters.py`, so the fix covers sync and async local mode (the async engine reuses the same collection).

## Test

Adds `test_bool_and_int_filters_do_not_cross_match` in `tests/test_in_memory.py` covering `MatchValue` (1 / True / 0 / False), `MatchAny`, and `Range` — each now returns only the correctly-typed points, matching the server. Verified against `dev`; the full `tests/test_in_memory.py` suite passes and `ruff format --line-length 99` (0.4.3) is clean.


eeshsaxena (commented):
This looks correct and the root cause is right - `bool` subclassing `int` (so `True == 1` and `True in [1]`) is exactly what makes the old `value == condition.value` cross-match. I reproduced all the affected paths on `main`:

```
check_match(MatchValue(value=1), True)   -> True   # should be False
check_match(MatchValue(value=True), 1)   -> True   # should be False
check_match(MatchAny(any=[1, 2]), True)  -> True   # should be False
check_range(Range(gt=0), True)           -> True   # should be False
```

and confirmed the real matches still pass (`MatchValue(1)` vs `1`, `MatchValue(True)` vs `True`).

A couple of things I checked that are worth noting for confidence:

- **Array payloads are covered.** `check_condition` flattens list-valued payloads and calls `check_match(condition.match, v)` per element (line 199), so routing everything through `values_match` also fixes the array case (`[True]` no longer matching `MatchValue(1)`), not just scalar payloads.
- **Range via the interface wrapper is covered** too, since `check_range_interface` delegates to the patched `check_range`.
- `MatchExcept` flips correctly: `not any(values_match(...))` means a `True` payload is no longer wrongly excluded by an `except_=[1]`.

One small suggestion: the added test covers `MatchValue`, but since the same `values_match` helper now backs `MatchAny`/`MatchExcept` and the separate `check_range` guard, a couple of extra asserts would lock in those paths against regressions - e.g. `MatchAny(any=[1])` vs `True`, `MatchExcept(except_=[1])` vs `True` (should keep the point), and a `Range` vs `True`. Entirely optional.

Nice, well-scoped fix. Deferring to the maintainers.

winklemad:
Thanks @eeshsaxena — appreciate the careful review and reproducing the paths. Added the isolated asserts in a270fb9: `MatchAny(any=[1])` vs the `True` payload now returns only the int point, and `Range(gt=0)` — which `True` would satisfy numerically if bools weren't excluded — returns only the int point too, so the single-value `MatchAny` path and the `check_range` guard are each locked in. `MatchExcept(except=[1])` keeping the `True` point was already asserted. Full `tests/test_in_memory.py` passes and ruff 0.4.3 is clean.

joein:
Hey @winklemad 

Thanks for pointing it out and fixing it!

Thank you @eeshsaxena for the review 👍 

I'll take a look at it

One problem is that we need to compare local mode implementation against the remote mode.
For that we have congruence_tests, and the test should be added there instead of test in memory.

I'll address it

joein (approved):

PR #1260, opened 2026-07-10
Preserve proto3 field presence for falsy values in REST/gRPC conversion

winklemad:
No linked issue — found by reading the REST↔gRPC conversion code.

## Summary

Three conversions in `qdrant_client/conversions/conversion.py` lose information for values that are *legitimately falsy* (`0`, `0.0`, `False`, or an unset optional). In each case a falsy-but-present value ends up indistinguishable from "missing", which changes the meaning of the request/response.

## Bugs

**1. `shard_key` / `order_value` of `0` dropped (REST → gRPC)**

`convert_scored_point` and `convert_record` guarded these with a plain truthiness check:

```python
shard_key=cls.convert_shard_key(model.shard_key) if model.shard_key else None,
order_value=cls.convert_order_value(model.order_value) if model.order_value else None,
```

An integer shard key of `0` or an order value of `0` / `0.0` is a valid value, but `if model.shard_key` is `False` for it, so the field never made it into the gRPC message. Switched to `is not None`.

**2. `IntegerIndexParams.range` / `.lookup` unset → `False` instead of `None` (gRPC → REST)**

```python
range=model.range,
lookup=model.lookup,
```

Both are `optional bool` fields, so reading them without a `HasField` guard returns the proto default `False` when they were never set — turning "unspecified" into an explicit `False` and silently overriding the server-side default. Every sibling field in the same struct (`is_principal`, `on_disk`, `enable_hnsw`) already guards with `HasField`; `range`/`lookup` now match.

**3. `SparseVectorParams.index` fabricated when unset (gRPC → REST)**

```python
index=(
    cls.convert_sparse_index_config(model.index)
    if model.HasField("index") is not None
    else None
),
```

`HasField(...)` returns a `bool`, and `bool is not None` is always `True`, so the guard never short-circuits. When no index was set, `model.index` is a default-constructed empty message and an empty `SparseIndexParams(full_scan_threshold=None, on_disk=None, datatype=None)` was produced instead of `None`. Dropped the dead `is not None`.

## Tests

Added three focused tests in `tests/conversions/test_validate_conversions.py`:

- `test_convert_point_falsy_shard_key_and_order_value` — `0` shard key / order value survive both `ScoredPoint` and `Record`, and `None` still maps to an unset field.
- `test_convert_integer_index_params_optional_flags` — unset `range`/`lookup` → `None`, explicit `True`/`False` preserved.
- `test_convert_sparse_vector_params_without_index` — no index → `None`, set index still converts.

Full `tests/conversions/test_validate_conversions.py` passes (24 tests), and `ruff format --line-length 99` (0.4.3) is clean.

### DIFF
--- a/qdrant_client/conversions/conversion.py
+++ b/qdrant_client/conversions/conversion.py
@@ -1594,8 +1594,8 @@ class GrpcToRest:
     ) -> rest.IntegerIndexParams:
         return rest.IntegerIndexParams(
             type=rest.IntegerIndexType.INTEGER,
-            range=model.range,
-            lookup=model.lookup,
+            range=model.range if model.HasField("range") else None,
+            lookup=model.lookup if model.HasField("lookup") else None,
             is_principal=model.is_principal if model.HasField("is_principal") else None,
             on_disk=model.on_disk if model.HasField("on_disk") else None,
             enable_hnsw=model.enable_hnsw if model.HasField("enable_hnsw") else None,
@@ -2155,9 +2155,7 @@ class GrpcToRest:
     ) -> rest.SparseVectorParams:
         return rest.SparseVectorParams(
             index=(
-                cls.convert_sparse_index_config(model.index)
-                if model.HasField("index") is not None
-                else None
+                cls.convert_sparse_index_config(model.index) if model.HasField("index") else None
             ),
             modifier=(
                 cls.convert_modifier(model.modifier) if model.HasField("modifier") else None
@@ -3092,8 +3090,14 @@ class RestToGrpc:
                 else None
             ),
             version=model.version,
-            shard_key=cls.convert_shard_key(model.shard_key) if model.shard_key else None,
-            order_value=cls.convert_order_value(model.order_value) if model.order_value else None,
+            shard_key=(
+                cls.convert_shard_key(model.shard_key) if model.shard_key is not None else None
+            ),
+            order_value=(
+                cls.convert_order_value(model.order_value)
+                if model.order_value is not None
+                else None
+            ),
         )
 
     @classmethod
@@ -3558,8 +3562,14 @@ class RestToGrpc:
                 if model.vector is not None
                 else None
             ),
-            shard_key=cls.convert_shard_key(model.shard_key) if model.shard_key else None,
-            order_value=cls.convert_order_value(model.order_value) if model.order_value else None,
+            shard_key=(
+                cls.convert_shard_key(model.shard_key) if model.shard_key is not None else None
+            ),
+            order_value=(
+                cls.convert_order_value(model.order_value)
+                if model.order_value is not None
+                else None
+            ),
         )
 
     @classmethod
--- a/tests/conversions/fixtures.py
+++ b/tests/conversions/fixtures.py
@@ -515,6 +515,45 @@ scored_point_multivector = grpc.ScoredPoint(
     version=12,
     order_value=order_value_float,
 )
+scored_point_shard_key_order_value_falsy_value = grpc.ScoredPoint(
+    id=point_id,
+    payload=payload,
+    score=0.99,
+    vectors=single_vector_output,
+    version=12,
+    shard_key=grpc.ShardKey(number=0),
+    order_value=grpc.OrderValue(int=3),
+)
+retrieved_point_shard_key_order_value_falsy_value = grpc.RetrievedPoint(
+    id=point_id,
+    payload=payload,
+    vectors=single_vector_output,
+    shard_key=grpc.ShardKey(number=0),
+    order_value=grpc.OrderValue(int=3),
+)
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Ran all three of these against `master` rather than reading the diff, and they check out. Then went looking for the same pattern elsewhere in the file and found eight more instances the PR does not touch, on the write path.

**Verification of what is here.**

```
######## master ########
scored: HasField(shard_key)=False HasField(order_value)=False
record: HasField(shard_key)=False HasField(order_value)=False
unset IntegerIndexParams -> range=False lookup=False is_principal=True
SparseVectorParams()     -> index=SparseIndexParams(full_scan_threshold=None, on_disk=None, datatype=None)

######## this branch ########
scored: HasField(shard_key)=True  HasField(order_value)=True
record: HasField(shard_key)=True  HasField(order_value)=True
unset IntegerIndexParams -> range=None lookup=None is_principal=True
SparseVectorParams()     -> index=None
```

The round trip argument is the cleanest way to see why 2 and 3 matter. `grpc -> rest -> grpc` on `master`:

```
IntegerIndexParams(is_principal=True)  roundtrip equal: False
IntegerIndexParams()                   roundtrip equal: False
SparseVectorParams()                   roundtrip equal: False
```

on this branch:

```
IntegerIndexParams(is_principal=True)  roundtrip equal: True
IntegerIndexParams(range=True, lookup=False) roundtrip equal: True
IntegerIndexParams()                   roundtrip equal: True
SparseVectorParams()                   roundtrip equal: True
SparseVectorParams(index=...on_disk=True)    roundtrip equal: True
```

So these are not cosmetic, `master` genuinely cannot round trip an unset `range`/`lookup` or an absent sparse index. `HasField("range")` not raising also confirms both are declared `optional` in the proto, which was the thing I most expected to be wrong.

The `HasField("index") is not None` line is a nice find, `HasField` returns a plain bool so that guard was unconditionally true and every `SparseVectorParams` without an index came back carrying a fabricated all-`None` `SparseIndexParams`.

All three new tests fail on `master` with only `conversion.py` reverted, so they are not just encoding the new behavior:

```
FAILED test_convert_point_falsy_shard_key_and_order_value
FAILED test_convert_integer_index_params_optional_flags
FAILED test_convert_sparse_vector_params_without_index
```

Full `tests/conversions/test_validate_conversions.py` is 24 passed on the branch. Worth noting the fix also covers `shard_key=""`, since `ShardKey = Union[StrictInt, StrictStr]` and the empty string is equally falsy. I checked, it survives on the branch and is dropped on `master`. Might be worth an assertion, it is free.

**The gap.** `RestToGrpc.convert_points_update_operation` has the identical truthiness guard in eight places:

```
4598:  if model.upsert.shard_key
4622:  if model.delete.shard_key
4643:  if model.set_payload.shard_key
4667:  if model.overwrite_payload.shard_key
4691:  if model.delete_payload.shard_key
4705:  if model.clear_payload.shard_key
4719:  if model.update_vectors.shard_key
4748:  if model.delete_vectors.shard_key
```

Still reproduces on this branch:

```
UpsertOperation          shard_key=0 survives? False    shard_key=7 survives? True
DeleteOperation          shard_key=0 survives? False    shard_key=7 survives? True
ClearPayloadOperation    shard_key=0 survives? False    shard_key=7 survives? True
```

This one bothers me more than the two the PR does fix. `convert_scored_point` and `convert_record` are response-shaped, so a dropped `shard_key` mostly costs the caller some metadata. `convert_points_update_operation` feeds `batch_update_points`, so an integer shard key of `0` silently loses its routing and the write lands wherever the server puts an unspecified selector. Silent misrouting on a write is a worse failure mode than a missing field on a read.

The correct idiom is already in the file, about 470 lines above, in `convert_query_request`:

```python
shard_key_selector=(
    cls.convert_shard_key_selector(model.shard_key) if model.shard_key is not None else None
),
```

so those eight are just stragglers that never got updated. Same one-token change you already made twice. (`ShardKeySelector` also admits `List[ShardKey]`, so an empty list is falsy at those sites too, though that input is degenerate enough that I would not worry about it.)

I would rather see them in this PR than in a follow-up, since the PR is already "audit the falsy guards in `conversion.py`" and leaving half the audit undone makes the next person assume it was finished. Everything already in the diff looks right to me.

winklemad:
Thanks for the detailed verification and for tracing the same pattern through the write path. I reproduced all eight `convert_points_update_operation` cases locally and confirmed that `shard_key=0` is dropped by the truthiness guards. Since #1279 now covers those eight sites with regression coverage for both `0` and `""`, I’ll keep this PR focused on the three conversion fixes already here to avoid duplicating or conflicting with that work.

joein:
Hey @winklemad 
Thanks for pointing this out!
I'll check whether the bugs exist and then review the PR.
Thanks @ErenAta16 for the review

joein (approved):

PR #1261, opened 2026-07-10
ci: add package sanity check and stop shipping test files (#958)

### LINKED ISSUE #958: Add release sanity check to CI

hassanzafarr:
Closes #958.

The published wheel/sdist currently ship test modules that live inside the `qdrant_client` package. Building `1.18.0` locally and inspecting the wheel, I see:

- `qdrant_client/local/tests/*` (7 files)
- `qdrant_client/hybrid/test_reranking.py`

Three of these `import pytest` at module level, which is what makes `pytest` an effective import-time dependency of the package and causes the broken releases mentioned in the issue.

Root cause: `packages = [{include = "qdrant_client"}]` includes the whole tree, and there's no `exclude`, so anything named like a test that happens to live under `qdrant_client/` gets packaged.

### Changes

1. **Stop shipping test-only code** — added an `exclude` to `[tool.poetry]` in `pyproject.toml` for `qdrant_client/**/tests` and `qdrant_client/**/test_*.py`. This only affects the built artifacts; the files stay in the source tree and are still collected/run by `pytest`, so nothing changes for local dev or CI test runs.

2. **Guard against regressions** — the issue asks for a CI check, so I added:
   - `tools/check_package_contents.py` — stdlib-only, builds nothing itself; it inspects the wheel + sdist and fails if any shipped module is a test file or imports `pytest`/`_pytest`.
   - `.github/workflows/package-sanity.yml` — runs `poetry build` and then the check on push/PR.

A check on its own would have gone red immediately because the leak is live today, so both parts are needed together.

### Testing

- `poetry build`, then ran the checker against the resulting `dist/`:
  - Against the **current** (leaky) build → fails, listing all 7 test files + the 3 pytest imports.
  - Against the build **with the exclude** → passes.
- Confirmed the excluded files remain in the source tree and that real modules (e.g. `qdrant_client/hybrid/fusion.py`, the `qdrant_client/local` implementation) still ship — 91 Python modules in the wheel, only the test files removed.
- Interestingly the checker also flagged `qdrant_client/hybrid/test_reranking.py`, which sits outside `local/tests`; the broader exclude pattern covers it too.

Happy to adjust the exclude patterns or the workflow triggers to match your preferences.

> Note: @glenlouis8 mentioned interest in the issue a while back — apologies for any overlap, happy to defer or collaborate if they already have work in progress.

### DIFF
--- a/.github/workflows/package-sanity.yml
+++ b/.github/workflows/package-sanity.yml
@@ -1,0 +1,27 @@
+name: package-sanity
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    name: Check built package contents
+
+    steps:
+      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2.7.0
+
+      - name: Set up Python
+        uses: actions/setup-python@e9aba2c848f5ebd159c070c61ea2c4e2b122355e # v2.3.4
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip poetry
+
+      - name: Build package
+        run: poetry build
+
+      - name: Check package contents
+        run: python tools/check_package_contents.py dist
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,13 @@ authors = ["Andrey Vasnetsov <andrey@qdrant.tech>"]
 packages = [
     {include = "qdrant_client"}
 ]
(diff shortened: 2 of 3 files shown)


joein:
Hey @hassanzafarr 

Though, the issue you're trying to solve has already been resolved in #1207 

Also, I think these tests are too much and we don't need them.

However, I find excluding test files from the distribution valuable!

Unfortunately, I can't accept this PR because all the PRs should point `dev` branch like mentioned in the PR template.

I'll close this and recreate a new PR with excluding the test files from the distribution and pointing `dev` branch, I'll set you as a co-author.

PR #1263, opened 2026-07-12
feat(upload): optional progress callback for upload_points and upload_collection

### LINKED ISSUE #1262: Add optional progress callback to upload_points / upload_collection
## Problem
Users uploading large collections (millions of points) currently have no visibility into upload progress. They cannot log progress, estimate time remaining, or implement per-batch abort behavior without instrumenting internal uploader classes.

## Current behavior
```python
client.upload_points(
    collection_name=my_collection,
    points=itertools.islice(my_points, 1_000_000),
    batch_size=64,
)
# call returns only after the entire stream is exhausted
```

There is no way to observe progress or cancel mid-upload.

## Proposed solution
Add an optional `progress_callback` keyword argument to `upload_points` and `upload_collection` (sync and async). Default `None` preserves existing behavior.

When provided, the callback is invoked from the main process once per successfully-uploaded batch with a small `UploadProgress` dataclass:

```python
from qdrant_client.uploader.uploader import UploadProgress

def on_progress(p: UploadProgress):
    print(f"uploaded {p.total_uploaded} / {p.total} ({p.batch_count} batches)")

client.upload_points(
    collection_name="my_collection",
    points=points,
    batch_size=64,
    progress_callback=on_progress,
)
```

`UploadProgress` would carry:
- `total_uploaded: int` - cumulative count of points accepted by the server
- `batch_count: int` - how many batches completed

## Alternatives considered
- tqdm wrapper in user-land: rejected, adds dependency burden on the user.
- Return a progress object from the call: rejected, breaks the existing `None` return signature.
- Context manager: rejected, less flexible and harder to compose.

## Acceptance criteria
- Callback fires once per successfully-uploaded batch.
- `total_uploaded` matches the number of points accepted by the server.
- `progress_callback=None` (default) preserves existing behavior with no measurable overhead.
- Works in local mode (`:memory:`), remote mode (REST and gRPC), sync, async, and `parallel > 1`.
- All public method signatures only gain an optional keyword argument.

## Backward compatibility
Additive only. No existing call patterns change. The async client files are generated from the sync source via `tools/generate_async_client.sh`; the generator output is updated to mirror the new parameter.

## Scope
Small, focused. Out of scope: tqdm integration, abort signal wiring, per-batch error aggregation callbacks.

Harsh23Kashyap:
Fixes #1262.

Adds an optional `progress_callback` keyword argument to `QdrantClient.upload_points` and `upload_collection` (and to the underlying `QdrantRemote._upload_collection`). When provided, the callback is invoked from the main process once per successfully-uploaded batch with a small `UploadProgress` dataclass that carries `total_uploaded` (cumulative points accepted) and `batch_count`. When `None` (default) the existing behavior is unchanged.

### Why

Users uploading large collections have no visibility into upload progress today. `BaseUploader.process()` already yields a `bool` per batch internally, but the public path discards it. This is the smallest surface that exposes it without changing the uploader interface.

### Usage

```python
from qdrant_client.uploader.uploader import UploadProgress

def on_progress(p: UploadProgress) -> None:
    print(f"uploaded {p.total_uploaded} points ({p.batch_count} batches)")

client.upload_points(
    collection_name="my_collection",
    points=points,
    batch_size=64,
    progress_callback=on_progress,
)
```

### Implementation notes

- `UploadProgress` is a `frozen=True, slots=True` dataclass with two `int` fields, importable from `qdrant_client.uploader.uploader`.
- `_upload_collection` materializes batch sizes once from the source iterator and zips them with the worker's results, so the callback is invoked from the main process regardless of `parallel`.
- The callback fires once per successful batch; failed batches (`result == False`) are silently skipped. Order is unordered in `parallel > 1` mode (matches existing behavior).
- Local mode (`QdrantLocal`) accepts the keyword argument for API consistency but does not invoke the callback (it doesn't go through `_upload_collection`).
- `BaseUploader` is intentionally left unchanged so the autogenerated async mirror does not need to be regenerated as part of this PR.

### Files changed

- `qdrant_client/uploader/uploader.py` — `UploadProgress` dataclass.
- `qdrant_client/qdrant_remote.py` — `_upload_collection`, `upload_points`, `upload_collection` accept `progress_callback`.
- `qdrant_client/qdrant_client.py` — public facade forwards `progress_callback` to remote.
- `tests/test_upload_progress.py` — seven tests covering dataclass invariants, default-`None` signature, sync path, failure handling, parallel path, local-mode signature acceptance, and public facade signature.

### Verification

- `pytest tests/test_upload_progress.py tests/test_in_memory.py` — 11 passed locally (Python 3.11).
- `ruff check` against touched files reports no new violations (two pre-existing F541 warnings remain on master around `qdrant_remote.py:90-91`).
- Smoke import: `from qdrant_client.uploader.uploader import UploadProgress` and `QdrantRemote._upload_collection(... progress_callback=...)` accepted via `inspect.signature`.

### Out of scope (intentional)

- `tqdm` integration, abort signal wiring, per-batch error aggregation callbacks. These are reasonable follow-ups but were not requested and would expand the surface meaningfully.

### Commits

1. `feat(uploader): add UploadProgress dataclass`
2. `feat(remote): thread progress_callback through _upload_collection`
3. `feat(client): expose progress_callback on QdrantClient uploaders`
4. `test(client): cover progress_callback across sync, parallel, and local modes`

### DIFF
--- a/qdrant_client/qdrant_client.py
+++ b/qdrant_client/qdrant_client.py
@@ -20,6 +20,7 @@ from qdrant_client.local.qdrant_local import QdrantLocal
 from qdrant_client.migrate import migrate
 from qdrant_client.qdrant_fastembed import QdrantFastembedMixin
 from qdrant_client.qdrant_remote import QdrantRemote
+from qdrant_client.uploader.uploader import UploadProgress
 
 
 class QdrantClient(QdrantFastembedMixin):
@@ -1875,6 +1876,7 @@ class QdrantClient(QdrantFastembedMixin):
         shard_key_selector: types.ShardKeySelector | None = None,
         update_filter: types.Filter | None = None,
         update_mode: types.UpdateMode | None = None,
+        progress_callback: Callable[[UploadProgress], None] | None = None,
         **kwargs: Any,
     ) -> None:
         """Upload points to the collection
@@ -1900,6 +1902,10 @@ class QdrantClient(QdrantFastembedMixin):
                 This parameter overwrites shard keys written in the records.
             update_filter: If specified, only points that match this filter will be updated, others will be inserted
             update_mode: Allows to alter default upsert behavior, instead of inserting a point if it does not exist, or updating it if it does, can be set to insert-only or update-only strategies.
+            progress_callback: Optional callable invoked from the main process once per
+                successfully-uploaded batch with a `UploadProgress` snapshot carrying
+                `total_uploaded` (cumulative points) and `batch_count`. When `None`
+                (default), no callback runs. Invoked once per batch regardless of `parallel`.
 
         """
 
@@ -1935,6 +1941,7 @@ class QdrantClient(QdrantFastembedMixin):
             shard_key_selector=shard_key_selector,
             update_filter=update_filter,
             update_mode=update_mode,
+            progress_callback=progress_callback,
         )
 
     def upload_collection(
@@ -1951,6 +1958,7 @@ class QdrantClient(QdrantFastembedMixin):
         shard_key_selector: types.ShardKeySelector | None = None,
         update_filter: types.Filter | None = None,
         update_mode: types.UpdateMode | None = None,
+        progress_callback: Callable[[UploadProgress], None] | None = None,
         **kwargs: Any,
     ) -> None:
         """Upload vectors and payload to the collection.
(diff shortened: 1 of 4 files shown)


Harsh23Kashyap:
Can someone take a look when you have a chance?

generall:
As mentioned in issue discussion https://github.com/qdrant/qdrant-client/issues/1262#issuecomment-4951747216
I don't think this is a desired feature.

PR #1264, opened 2026-07-13
fix: prevent file handle leak when portalocker.lock() fails in local mode

### LINKED ISSUE #1234: fix: file handle leak when portalocker.lock() raises exception in local mode

saitejabandaru-in:
Resolves #1234.

### Description
In `QdrantLocal` and `AsyncQdrantLocal`, when starting the local storage client, a flock lock file is opened. If the directory is already locked (or another error occurs during `portalocker.lock()`), a `LockException` or other exception is raised, but the opened `self._flock_file` file descriptor is left open/leaked.

### Changes
- Added `except` blocks wrapping `portalocker.lock()` in both sync and async local client initializers to explicitly call `self._flock_file.close()` and clear the reference before re-raising the exceptions.
- Verified that existing persistence unit tests still pass successfully.

### DIFF
--- a/qdrant_client/local/async_qdrant_local.py
+++ b/qdrant_client/local/async_qdrant_local.py
@@ -137,7 +137,7 @@ class AsyncQdrantLocal(AsyncQdrantBase):
             self._flock_file = None
             raise RuntimeError(
                 f"Storage folder {self.location} is already accessed by another instance of Qdrant client. If you require concurrent access, use Qdrant server instead."
-            )
+            ) from None
         except Exception:
             self._flock_file.close()
             self._flock_file = None
--- a/qdrant_client/local/async_qdrant_local.py
+++ b/qdrant_client/local/async_qdrant_local.py
@@ -133,9 +133,15 @@ class AsyncQdrantLocal(AsyncQdrantBase):
                 portalocker.LockFlags.EXCLUSIVE | portalocker.LockFlags.NON_BLOCKING,
             )
         except portalocker.exceptions.LockException:
+            self._flock_file.close()
+            self._flock_file = None
             raise RuntimeError(
                 f"Storage folder {self.location} is already accessed by another instance of Qdrant client. If you require concurrent access, use Qdrant server instead."
             )
+        except Exception:
+            self._flock_file.close()
+            self._flock_file = None
+            raise
 
     def _save(self) -> None:
         if not self.persistent:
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
The leak is real. In `_load`, `self._flock_file = open(lock_file_path, "r+")` happens before `portalocker.lock(...)`, and on `LockException` the old code raised `RuntimeError` without closing that handle - so the file object leaks until GC (and on Windows the underlying handle can linger, which matters precisely in the "another instance already holds the lock" scenario this path is about). Closing it and setting `self._flock_file = None` before raising is the right fix.

Two things I liked:
- The broad `except Exception: ... raise` also covers non-`LockException` failures (e.g. an `OSError` from `portalocker.lock`), so the handle is cleaned up on any failure, not just the expected one. Ordering is correct too (the specific `LockException` is caught first).
- `raise RuntimeError(...) from None` suppresses the confusing "during handling of the above exception" chain, which reads better for this user-facing message.

Heads-up: #1241 proposes the same fix for the same lines. The main difference I can see is that this PR adds `from None` (which I think is the nicer traceback), so if the maintainers consolidate, this looks like the more complete of the two.

One optional nit: `self._flock_file.close()` inside the handler could itself theoretically raise, masking the original error; wrapping it in a small `try/except` or `contextlib.suppress(Exception)` would make the cleanup fully defensive. Minor. Deferring to the maintainers.

saitejabandaru-in:
Thanks for the review, @eeshsaxena and @coderabbitai! I have applied the fix to satisfy the Ruff B904 rule by using `raise ... from None` and pushed the changes.

PR #1270, opened 2026-07-17
test: add unit tests for local distance functions

eeshsaxena:
## Description

`qdrant_client/local/distances.py` had no dedicated test module. This adds `tests/test_distances.py` covering the pure distance/scoring helpers:

- `dot_product`, `euclidean_distance`, `manhattan_distance` against hand-computed values
- `distance_to_order` mapping each `Distance` to bigger/smaller-is-better
- `fast_sigmoid` mapping into (-1, 1), passing NaN/inf through unchanged
- `scaled_fast_sigmoid` shifting into (0, 1)

(`cosine_similarity` normalises its inputs in place, so each helper is exercised with fresh arrays.) No source changes.

## Verification

`pytest tests/test_distances.py` - 7 passed. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/tests/test_distances.py
+++ b/tests/test_distances.py
@@ -1,0 +1,62 @@
+import numpy as np
+import pytest
+
+from qdrant_client.http import models
+from qdrant_client.local.distances import (
+    DistanceOrder,
+    distance_to_order,
+    dot_product,
+    euclidean_distance,
+    fast_sigmoid,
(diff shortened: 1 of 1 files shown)


joein:
Hey @eeshsaxena 

Though, I don't think we need these exact tests :(

PR #1273, opened 2026-07-18
fix(local): continue update_vectors batch when a point fails update_filter

### LINKED ISSUE #1237: bug: update_vectors returns early when first point doesn't match update_filter, skipping remaining points

Harsh23Kashyap:
## Summary

Fix #1237: `LocalCollection.update_vectors` exited the batch loop the moment a point failed the `update_filter` check, so every point after it was silently skipped. Changed the early return to `continue` so each point is evaluated independently. The async mirror (`AsyncQdrantLocal.update_vectors`) delegates to this collection, so the fix applies there too.

## Changes

* **`qdrant_client/local/local_collection.py`** — one-line fix in `LocalCollection.update_vectors`: `return None` -> `continue` after `check_filter`.
* **`tests/test_local_update_vectors_filter.py`** (new) — regression coverage:
  * the reproducer from issue #1237 (matching point after a non-matching point)
  * all-points-match boundary
  * no-points-match boundary
  * async mirror via `AsyncQdrantClient(":memory:")`

## Verification

```
$ .ossvenv/bin/python -m pytest tests/test_local_update_vectors_filter.py -v
============================== 4 passed in 0.65s ==============================

$ .ossvenv/bin/python -m ruff check tests/test_local_update_vectors_filter.py qdrant_client/local/local_collection.py
All checks passed!
```

Mutation check: with the source change reverted to `return None`, the two "later matching point" tests fail (sync and async), confirming the guard is real.

The in-memory backend normalizes vectors on every write under a COSINE config, so the expected vectors in the assertions are the post-normalization form. Inputs are chosen to be directionally distinct so per-point assertions stay unambiguous.

## Backward compatibility

The function is annotated `-> None`; the previous `return None` was only legal because no caller consumed the return value. No public API change.

## Risk

Low. Single-line change inside a single loop, fully covered by a deterministic local-only test. No new dependencies, no protocol changes.

Fixes #1237

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -2646,7 +2646,7 @@ class LocalCollection:
                 if not check_filter(
                     update_filter, self.payload[idx], self.ids_inv[idx], has_vector
                 ):
-                    return None
+                    continue
             self._update_named_vectors(idx, fixed_vectors)
             self._persist_by_id(point_id)
 
--- a/tests/test_local_update_vectors_filter.py
+++ b/tests/test_local_update_vectors_filter.py
@@ -1,0 +1,166 @@
+"""Regression tests for issue #1237.
+
+``LocalCollection.update_vectors`` previously did ``return None`` when a point
+failed the ``update_filter`` check, which exited the entire batch loop and
+silently skipped every remaining point. The fix is to ``continue`` so each
+point is evaluated independently.
+
+Covers:
+
+* The reproducer from issue #1237 — a point that matches ``update_filter`` after
+  a non-matching point is still updated.
+* All-matching and none-matching boundaries.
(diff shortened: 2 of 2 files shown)


Harsh23Kashyap:
First-time follow-up. `continue update_vectors batch when a point fails update_filter` — the fix lets a single point failure (404 etc.) be skipped without aborting the rest of the batch. Netlify deploy preview is green, no review threads yet (CodeRabbit just summarized). Standing by for a maintainer pass.

joein:
Hey, thanks for the contribution!

However, this has already been fixed in #1244

PR #1276, opened 2026-07-21
fix: fix local mode set payload shared nested object #1271

joein:
#1271  #1272


ErenAta16:
Thanks for picking this up and finishing it off, joein, and for the co-author credit. The deepcopy on the shared nested payload is exactly the right fix, and folding it in with the #1271 repro closes the loop cleanly. Glad it landed.

PR #1279, opened 2026-07-22
fix: do not drop falsy shard keys in convert_points_update_operation

ErenAta16:
### What's broken

All eight branches of `RestToGrpc.convert_points_update_operation` gate the shard key on truthiness:

```python
shard_key_selector = (
    cls.convert_shard_key_selector(model.upsert.shard_key)
    if model.upsert.shard_key
    else None
)
```

`ShardKey = Union[StrictInt, StrictStr]`, so `0` and `""` are valid shard keys that evaluate to `False`. They are silently replaced with `None`, and the operation gets routed by the default rules instead of to the requested shard. Nothing raises, so the write just lands somewhere else.

Repro against `dev`, converting each operation type with `shard_key=0`:

```
shard_key = 0

LOST UpsertOperation            shard_key_selector present=False keys=[]
LOST DeleteOperation            shard_key_selector present=False keys=[]
LOST SetPayloadOperation        shard_key_selector present=False keys=[]
LOST OverwritePayloadOperation  shard_key_selector present=False keys=[]
LOST DeletePayloadOperation     shard_key_selector present=False keys=[]
LOST ClearPayloadOperation      shard_key_selector present=False keys=[]
LOST UpdateVectorsOperation     shard_key_selector present=False keys=[]
LOST DeleteVectorsOperation     shard_key_selector present=False keys=[]

8/8 operations silently dropped the shard key
control, shard_key=1 -> [number: 1]
```

With this patch:

```
OK   UpsertOperation            shard_key_selector present=True keys=[number: 0]
OK   DeleteOperation            shard_key_selector present=True keys=[number: 0]
OK   SetPayloadOperation        shard_key_selector present=True keys=[number: 0]
OK   OverwritePayloadOperation  shard_key_selector present=True keys=[number: 0]
OK   DeletePayloadOperation     shard_key_selector present=True keys=[number: 0]
OK   ClearPayloadOperation      shard_key_selector present=True keys=[number: 0]
OK   UpdateVectorsOperation     shard_key_selector present=True keys=[number: 0]
OK   DeleteVectorsOperation     shard_key_selector present=True keys=[number: 0]

0/8 operations silently dropped the shard key
```

### Why it matters

This is on the write path. `QdrantRemote.batch_update_points` (and the async variant) call `RestToGrpc.convert_update_operation` for every operation when the client is constructed with `prefer_grpc=True`. A custom-sharded collection keyed on integers loses shard `0` on every batch update, and there is no signal that it happened.

### The fix

Switch the eight guards to `is not None`. The field is declared `Optional[ShardKeySelector]` with a `None` default, so `None` is the only value that should mean "not set". This is the idiom the rest of the file already uses, for example in `convert_query_request`:

```python
shard_key_selector=(
    cls.convert_shard_key_selector(model.shard_key)
    if model.shard_key is not None
    else None
),
```

so this only brings the update path in line with the read path. No behaviour changes for `None` or for any truthy key.

### Testing

`test_convert_points_update_operation_falsy_shard_key` builds each of the eight REST operation types with `shard_key=0` and `shard_key=""`, converts to gRPC, and asserts the selector survives and round-trips back to the original value.

```
$ pytest tests/conversions/test_validate_conversions.py -q
......................                                                   [100%]
22 passed in 1.16s
```

Reverting only the `conversion.py` hunk makes the new test fail on the first operation and leaves the other 21 tests passing, so nothing else in the file depended on the old behaviour. `ruff format --line-length=99` reports no changes.

### DIFF
--- a/qdrant_client/conversions/conversion.py
+++ b/qdrant_client/conversions/conversion.py
@@ -4595,7 +4595,7 @@ class RestToGrpc:
         if isinstance(model, rest.UpsertOperation):
             shard_key_selector = (
                 cls.convert_shard_key_selector(model.upsert.shard_key)
-                if model.upsert.shard_key
+                if model.upsert.shard_key is not None
                 else None
             )
             update_filter = (
@@ -4619,7 +4619,7 @@ class RestToGrpc:
         elif isinstance(model, rest.DeleteOperation):
             shard_key_selector = (
                 cls.convert_shard_key_selector(model.delete.shard_key)
-                if model.delete.shard_key
+                if model.delete.shard_key is not None
                 else None
             )
             points_selector = cls.convert_points_selector(model.delete)
@@ -4640,7 +4640,7 @@ class RestToGrpc:
 
             shard_key_selector = (
                 cls.convert_shard_key_selector(model.set_payload.shard_key)
-                if model.set_payload.shard_key
+                if model.set_payload.shard_key is not None
                 else None
             )
 
@@ -4664,7 +4664,7 @@ class RestToGrpc:
 
             shard_key_selector = (
                 cls.convert_shard_key_selector(model.overwrite_payload.shard_key)
-                if model.overwrite_payload.shard_key
+                if model.overwrite_payload.shard_key is not None
                 else None
             )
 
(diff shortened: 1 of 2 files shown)


joein:
Hey @ErenAta16 

Thanks for pointing it out!

I'll review the PR

joein:
looks good to me!

let's wait for the green ci

joein (approved):

PR #1284, opened 2026-07-26
fix(remote): propagate gRPC timeout fallback to the protobuf field

### LINKED ISSUE #948: Unexpected Timeout Behavior in AsyncQdrantClient: timeout Parameter Not Enforcing Specified Duration
I'm encountering an issue with the timeout parameter in the AsyncQdrantClient. When I set the timeout to 1 second, the client does not enforce this duration as expected.  

```
from qdrant_client import AsyncQdrantClient

client = AsyncQdrantClient(url="http://localhost:6333", timeout=1)
```
### Issue Description:

Despite setting timeout=1.0, operations occasionally take longer than one second to fail or complete. This behavior suggests that the timeout is not being strictly enforced.​

### Expected Behavior:

All client operations should fail if they exceed the specified timeout duration of 1 second.​

### Additional Information:

Qdrant Client Version: 1.11
Python Version: 3.12.9
Operating System: Ubuntu 20.04.6 LTS

I would appreciate any insights into why the timeout is not functioning as intended and whether there are recommended workarounds or upcoming fixes.​

Thank you for your assistance!

Harsh23Kashyap:
## Summary

Fix the gRPC timeout-fallback bug: when a method is called without a per-call `timeout`, the gRPC transport deadline was set from `self._timeout` (the global constructor value), but the protobuf `timeout` field on the gRPC request stayed as `None`. The Qdrant server uses the protobuf field for its own internal operation budget, so a user-set global timeout was effectively ignored server-side.

Repro from the issue:

```
AsyncQdrantClient(timeout=120)  # 120s gRPC deadline
                              # but server still enforces its own ~60s default
```

## Root cause

Per `@yudelevi` on #948, every gRPC call site in `qdrant_remote.py` / `async_qdrant_remote.py` had the pattern:

```python
self.grpc_points.X(
    grpc.Y(..., timeout=timeout, ...),  # protobuf field, may be None
    timeout=timeout if timeout is not None else self._timeout,  # gRPC deadline, correct
)
```

The first `timeout=` is the protobuf field, which the server uses for its internal operation budget. When the per-call `timeout` was `None`, the field stayed unset and the server fell back to its default (~60s).

## Fix

Across every gRPC call site (read methods, write methods, batches, single-line gRPC calls, the `UpdateCollectionClusterSetupRequest` single-line path, etc.), introduce a local:

```python
effective_timeout = timeout if timeout is not None else self._timeout
```

and use it for both the protobuf field and the gRPC deadline. Methods that take no per-call `timeout` parameter (e.g. `collection_exists`, `get_collection`, `list_shard_keys`) use `effective_timeout = self._timeout` directly.

Total: 174 insertions / 111 deletions across `qdrant_remote.py` and `async_qdrant_remote.py`. The diff is mechanical and reviewable site by site.

## What's NOT in this PR

The HTTP path has the same `timeout=timeout` pattern in calls like `self.http.search_api.query_points(...)`. But `httpx` treats `None` as "no timeout", which is a different bug worth its own discussion. Left out of scope.

## Test plan

`tests/test_grpc_timeout_fallback.py` exercises the protobuf field and the gRPC deadline without a real network call by monkey-patching the gRPC stub pool (the property round-robins across the pool, so all pool entries are patched). Covers:

- global `timeout=42` -> both proto field and gRPC deadline = 42
- per-call `timeout=7` overrides global `timeout=42` -> both = 7
- neither set -> both = `QdrantRemote.DEFAULT_GRPC_TIMEOUT` (5)
- a write method (`upsert`) that originally ignored the per-call timeout
- a no-timeout-param method (`collection_exists`) that uses `self._timeout` directly
- all of the above for both `QdrantClient` and `AsyncQdrantClient`

```
$ .ossvenv/bin/pytest tests/test_grpc_timeout_fallback.py -v
============================== 11 passed in 0.43s ==============================
```

Existing `tests/test_tracing.py`, `tests/test_common.py`, `tests/test_in_memory.py` still pass (41/41).

`mypy qdrant_client/qdrant_remote.py qdrant_client/async_qdrant_remote.py` is clean. `ruff check tests/test_grpc_timeout_fallback.py` is clean. The 3 pre-existing F541 warnings in `qdrant_remote.py` / `async_qdrant_remote.py` are not introduced by this PR.

## Backward compatibility

This is a behavior change, but only in the direction of "user-set timeout is now respected". The default behavior (no global, no per-call) still uses `DEFAULT_GRPC_TIMEOUT = 5` for both the proto field and the gRPC deadline — same as before, just made consistent. Callers that were relying on the per-call `timeout` to be silently ignored by write methods (e.g. `upsert`) will see the per-call value now flow through to the protobuf field; the gRPC deadline was already using `self._timeout` for those.

Fixes #948

### DIFF
--- a/qdrant_client/async_qdrant_remote.py
+++ b/qdrant_client/async_qdrant_remote.py
@@ -45,6 +45,13 @@ class AsyncQdrantRemote(AsyncQdrantBase):
     DEFAULT_GRPC_TIMEOUT = 5
     DEFAULT_GRPC_POOL_SIZE = 3
 
+    def _effective_timeout(self, timeout: int | None) -> int:
+        # Single source of truth for the per-call / global fallback rule.
+        # Every gRPC call site in this file uses this so a future
+        # method that takes a `timeout=` arg cannot accidentally bypass
+        # `self._timeout` again. See issue #948.
+        return timeout if timeout is not None else self._timeout
+
     def __init__(
         self,
         url: str | None = None,
@@ -430,7 +437,7 @@ class AsyncQdrantRemote(AsyncQdrantBase):
         **kwargs: Any,
     ) -> types.QueryResponse:
         if self._prefer_grpc:
-            effective_timeout = timeout if timeout is not None else self._timeout
+            effective_timeout = self._effective_timeout(timeout)
             if query is not None:
                 query = RestToGrpc.convert_query(query)
             if isinstance(prefetch, models.Prefetch):
@@ -518,7 +525,7 @@ class AsyncQdrantRemote(AsyncQdrantBase):
         **kwargs: Any,
     ) -> list[types.QueryResponse]:
         if self._prefer_grpc:
-            effective_timeout = timeout if timeout is not None else self._timeout
+            effective_timeout = self._effective_timeout(timeout)
             requests = [
                 RestToGrpc.convert_query_request(r, collection_name)
                 if isinstance(r, models.QueryRequest)
@@ -585,7 +592,7 @@ class AsyncQdrantRemote(AsyncQdrantBase):
         **kwargs: Any,
     ) -> types.GroupsResult:
         if self._prefer_grpc:
-            effective_timeout = timeout if timeout is not None else self._timeout
+            effective_timeout = self._effective_timeout(timeout)
             if query is not None:
                 query = RestToGrpc.convert_query(query)
             if isinstance(prefetch, models.Prefetch):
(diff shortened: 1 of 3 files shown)


Harsh23Kashyap:
Re: the two follow-ups from the CodeRabbit review on #1284 — both addressed in f1fbc5e4:

1. **Single source of truth.** Added `self._effective_timeout(timeout: int | None) -> int` to both `QdrantRemote` and `AsyncQdrantRemote`. Replaced the ~58 inline `effective_timeout = ...` expressions with calls to it. Methods that take a per-call `timeout` use `self._effective_timeout(timeout)`; methods that don't (e.g. `collection_exists`, `list_shard_keys`) use `self._effective_timeout(None)` so the rule lives in exactly one place.

2. **5 missed write methods fixed.** `create_collection`, `create_payload_index`, `delete_payload_index`, `create_vector_name`, `delete_vector_name` in both sync and async (10 sites total) were each passing the raw per-call `timeout` to the protobuf field while using `self._timeout` for the gRPC deadline — the exact bug pattern the patch was meant to eliminate. Each now uses `self._effective_timeout(timeout)` for both.

Tests extended to cover the 5 newly-fixed methods. 16/16 pass; mypy clean; ruff check clean on the new test file.

PR #914, opened 2025-03-04
WIP: new: Support for too many requests

tellet-q:
Add support for following responses from Qdrant:

Rate limiter hit for REST API: https://github.com/qdrant/qdrant/pull/5917
Rate limiter hit for gRPC: https://github.com/qdrant/qdrant/pull/6072

Ensure rest and grpc uploaders continue execution after a pause if the above received.

### All Submissions:

* [ ] Contributions should target the `dev` branch. Did you create your branch from `dev`?
* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### New Feature Submissions:

1. [ ] Does your submission pass tests?
2. [ ] Have you installed `pre-commit` with `pip3 install pre-commit` and set up hooks with `pre-commit install`?

PR #1291, opened 2026-07-29
fix: escape url path params in the rest api client

### LINKED ISSUE #807: asking if collection exists and collection name with "/"
Asking for to the client for a collection name with includes "/" in the name raises UnexpectedResponse. Can this be validated in the client site before calling the remote qdrant server ?

vaibzde:
Closes #807

### Problem

Path params are interpolated into the request path as-is in `ApiClient.request`, so a value containing a reserved character silently changes the path of the request instead of being sent as a single segment:

```python
client = QdrantClient(url="http://localhost:6333")
client.collection_exists(collection_name="example/collection1")
```

builds `GET /collections/example/collection1/exists`, which is not an existing route, so the call fails with `UnexpectedResponse` instead of reporting on the collection.

### Change

`quote_path_params()` percent-encodes every path param before the url is formatted, in both `ApiClient` and `AsyncApiClient`. All path params in the generated api (`collection_name`, `shard_id`, `snapshot_name`, `vector_name`, `peer_id`, `id`, `field_name`) are single segments, so escaping them is safe. Names made of unreserved characters are unchanged, so this does not affect existing requests.

### Tests

Added `tests/test_api_client.py`, which asserts the built request url without needing a running Qdrant:

* a plain name is not escaped (both clients)
* a name with `/` stays one path segment (both clients)
* a name with a space is escaped
* non-string path params such as `shard_id=1` still work

I checked that the two escaping cases fail without the change and pass with it.

### How I verified

```
$ pytest tests/test_api_client.py -q
6 passed

$ pytest tests/test_in_memory.py tests/test_tracing.py tests/conversions -q
40 passed
```

I did not run the integration test suite locally, since it needs a running Qdrant instance — relying on CI for that.

### One thing worth your call

`qdrant_client/http/` is generated, and `generate_rest_client.sh` replaces the whole directory, so this would be overwritten on the next regeneration. I put the change here because `api_client.py` has been patched directly in this repo before (#890, #996, #914). If you would rather have it in the `pydantic_openapi_v3` template, I am happy to open a PR there instead and drop this one.

I also assumed encoding is preferable to raising a client-side `ValueError` on such names, since it does not add new error semantics — but the issue mentioned validation as an option too, so let me know if you prefer that.

### DIFF
--- a/qdrant_client/http/api_client.py
+++ b/qdrant_client/http/api_client.py
@@ -1,7 +1,7 @@
 from asyncio import get_event_loop
 from functools import lru_cache
 from typing import Any, Awaitable, Callable, Dict, Generic, Type, TypeVar, overload
-from urllib.parse import urljoin
+from urllib.parse import quote, urljoin
 
 from httpx import AsyncClient, Client, Request, Response
 from pydantic import ValidationError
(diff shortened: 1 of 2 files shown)


vaibzde:
Rebased-free bump on this one — still happy to switch to a client-side `ValueError` instead of the encoding fix if that's the preferred shape, or to move this into `pydantic_openapi_v3` if you'd rather it land in the generator template. Let me know either way.

PR #1303, opened 2026-08-01
fix: require values_count bounds to hold for the same count

### LINKED ISSUE #1292: BUG: local mode values_count matches when bounds are satisfied by different values

r0h1tb:
Fixes #1292

## Problem

`check_values_count` evaluated each bound (`lt`, `lte`, `gt`, `gte`) independently across all of a point's value counts. A point therefore matched when *one* value satisfied one bound and a *different* value satisfied another, even though no single count was in range.

The server ANDs all four bounds against a single count (`ValuesCount::check_count` in `lib/segment/src/types.rs`), so local mode diverged.

## Reproduction

```python
payload = {"nested": [{"field": [1]}, {"field": list(range(10))}]}   # counts: [1, 10]
flt = models.Filter(must=[models.FieldCondition(
    key="nested[].field", values_count=models.ValuesCount(gt=2, lt=9))])
```

Local mode returned the point, because `10 > 2` and `1 < 9`. Neither `1` nor `10` is inside `(2, 9)`, and the server rejects it.

## Fix

A point matches when at least one count satisfies *every* bound:

```python
return any(
    (condition.lt is None or count < condition.lt)
    and (condition.lte is None or count <= condition.lte)
    and (condition.gt is None or count > condition.gt)
    and (condition.gte is None or count >= condition.gte)
    for count in get_value_counts(values)
)
```

`get_value_counts` never returns an empty list (`[]` yields `[0]`), so `any()` doesn't introduce an empty-sequence edge case.

## Why this wasn't caught

It only reproduces on multi-value paths. With a single count the old logic is accidentally equivalent to the correct one, and every case in `tests/congruence_tests/test_values_count.py` uses a top-level key, which yields exactly one count.

## Tests

Adds `tests/test_local_payload_filters.py` — unit coverage for `get_value_counts` and `check_values_count`, including multi-value cases and the `gte=10, lte=1` shape where two different counts each satisfy one bound. These need no server, so they cover paths the congruence tests can't reach.

Both regression tests fail on `master` and pass with this change; the other 23 cases pass either way. `tests/test_in_memory.py` and `tests/test_local_persistence.py` still pass (34 total).

Also verified end to end through `QdrantClient(":memory:")` with the payload above: the out-of-range point is no longer returned, and a point whose count is genuinely inside the range still matches.

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -31,17 +31,16 @@ def check_values_count(condition: models.ValuesCount, values: list[Any] | None) 
     if values is None:
         return False
 
-    counts = get_value_counts(values)
-
-    if condition.lt is not None and all(count >= condition.lt for count in counts):
-        return False
-    if condition.lte is not None and all(count > condition.lte for count in counts):
-        return False
-    if condition.gt is not None and all(count <= condition.gt for count in counts):
-        return False
-    if condition.gte is not None and all(count < condition.gte for count in counts):
-        return False
-    return True
+    # All bounds have to hold for the *same* count, matching the server's
+    # ValuesCount::check_count. Evaluating each bound separately would let two
+    # different values combine to satisfy a range that neither one is inside.
+    return any(
+        (condition.lt is None or count < condition.lt)
+        and (condition.lte is None or count <= condition.lte)
+        and (condition.gt is None or count > condition.gt)
+        and (condition.gte is None or count >= condition.gte)
+        for count in get_value_counts(values)
+    )
 
 
 def check_geo_radius(condition: models.GeoRadius, values: Any) -> bool:
--- a/tests/test_local_payload_filters.py
+++ b/tests/test_local_payload_filters.py
@@ -1,0 +1,98 @@
+"""Unit tests for local-mode payload filter helpers.
+
+These run without a server, so they can cover the multi-value paths that the
+congruence tests in ``tests/congruence_tests`` don't reach.
+"""
+
+import pytest
+
+from qdrant_client import models
+from qdrant_client.local.payload_filters import check_values_count, get_value_counts
+
+
+class TestGetValueCounts:
+    @pytest.mark.parametrize(
+        ("values", "expected"),
+        [
+            ([], [0]),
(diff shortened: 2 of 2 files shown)


joein:
Hey @r0h1tb 

Thanks for the contribution!

Unfortunately, our contribution guides require all the created branches to point to the `dev` branch, otherwise it is tricky to merge it into dev properly due to our release pipeline.

Also, there is already an attempt to fix it by the author of the original issue.

Please, make sure that there is no work going on the issue before creating a PR.

Closing this for now due to branch mismatch

r0h1tb:
Thanks for taking the time to explain, and apologies for the noise — both points are fair.

You're right on the duplicate: #1293 was opened by @nazsats minutes after they filed #1292, three days before mine. I should have checked the reporter's own PRs first and didn't. And the `dev` base is the first line of the PR template, so that one is on me too.

No need to reconsider this — #1293 is the right PR to land, and I won't resubmit against `dev`.

PR #1304, opened 2026-08-01
feat(client): add get_collection_names() shortcut (#1302)

### LINKED ISSUE #1302: feat(client): add `get_collection_names()` shortcut method to all 6 client classes
# feat(client): add `get_collection_names()` shortcut method to all 6 client classes

## Problem

`get_collections()` returns a `types.CollectionsResponse` whose `.collections` is a list of `CollectionDescription` objects. The common case — "I just want the names" — requires a list comprehension:

```python
names = [c.name for c in client.get_collections().collections]
```

This works, but it is verbose for what should be a one-liner, and the `CollectionDescription` model is more than most callers want from a "list of names" query.

## Motivation

Three concrete use cases the workaround complicates:

1. **Cleanup scripts.** A teardown script that drops a list of test collections: `for name in [c.name for c in client.get_collections().collections]: client.delete_collection(name)`. A direct `for name in client.get_collection_names():` is the natural form.

2. **Conditional logic on which collections exist.** "Does this index live in its own collection or is it a payload field of a bigger one?" — the script wants `["posts", "comments", "users"]` to render a status table, not a list of `CollectionDescription` objects.

3. **Logging and dashboards.** A health endpoint that reports "managed collections: 12, names: [...]" wants the names directly, not a model dump.

Verified: `grep -n "def get_collection_names" qdrant_client/*.py qdrant_client/local/*.py` returns nothing. No open or closed issue mentions a public `get_collection_names()` method on the client.

## Current behavior

```python
>>> client = QdrantClient(":memory:")
>>> client.create_collection("a", vectors_config=VectorParams(size=4, distance=Distance.COSINE))
>>> client.create_collection("b", vectors_config=VectorParams(size=4, distance=Distance.COSINE))
>>> client.get_collection_names()
AttributeError: 'QdrantClient' object has no attribute 'get_collection_names'
>>> # workaround:
>>> [c.name for c in client.get_collections().collections]
['a', 'b']
```

## Expected behavior

```python
>>> client = QdrantClient(":memory:")
>>> client.create_collection("a", vectors_config=VectorParams(size=4, distance=Distance.COSINE))
>>> client.create_collection("b", vectors_config=VectorParams(size=4, distance=Distance.COSINE))
>>> client.get_collection_names()
['a', 'b']
>>> client.get_collection_names() == sorted(client.get_collection_names())
True
```

Async equivalent (the method is `async` and must be awaited):

```python
>>> client = AsyncQdrantClient(":memory:")
>>> names = await client.get_collection_names()
>>> names
['a', 'b']
```

Works for `:memory:` local mode, `path=` local mode, and remote mode (gRPC and REST). The remote path delegates to the existing `get_collections()` and extracts `.name` from each `CollectionDescription`. The local path returns `list(self.collections.keys())` (no `get_collections()` round-trip, no model construction).

## Proposed solution

Add `get_collection_names() -> list[str]` to each of the 6 classes:

- `QdrantRemote.get_collection_names()` — sync. `return [c.name for c in self.get_collections().collections]`. 1 line.
- `AsyncQdrantRemote.get_collection_names()` — async. `return [c.name for c in (await self.get_collections()).collections]`. 1 line.
- `QdrantLocal.get_collection_names()` — sync. `return list(self.collections.keys())` after the same `closed` guard `get_collections()` uses. 2 lines.
- `AsyncQdrantLocal.get_collection_names()` — sync (no I/O). Same as `QdrantLocal.get_collection_names()`. 2 lines.
- `QdrantClient.get_collection_names()` — sync. Delegates to `self._client.get_collection_names()`. 1 line.
- `AsyncQdrantClient.get_collection_names()` — async. Awaits the inner result if it's a coroutine (remote mode), returns the list directly (local mode, sync). Detected via `inspect.iscoroutine`. ~5 lines.

The order matches `get_collections()`'s order (server-side for remote, insertion order for local — Python dicts preserve insertion order since 3.7). Callers that want a sorted list use `sorted(client.get_collection_names())`.

## Alternatives considered

- **Return a `set[str]` instead of `list[str]`.** Rejected. The natural return type for a "list of things" is a list; sets lose insertion order and have surprising semantics for the "I want them in the order the server returned" use case.
- **Accept a `prefix: str` filter parameter.** Rejected for this PR. Users who need a filter do `[n for n in client.get_collection_names() if n.startswith("user_")]`. Adding a `prefix` parameter is a separate concern; if it lands later, it's a small additive change.
- **Add `__contains__` to support `"name" in client`.** Different shape; deferred. `"name" in client.collection_names` is closer to the "list of names" semantic anyway, and the user can build it from `get_collection_names()` if they want it.
- **Return a `tuple[str, ...]` instead of `list[str]`.** Rejected. Lists are the conventional return type for variable-length string sequences in Python, and the user-side `for name in client.get_collection_names():` works with either, but lists are what `dict.keys()` and `sorted()` produce.

## Scope

In scope:
- 6 small methods, one per class.
- 1 line in each of `client_generator.py` and `remote_generator.py` (add `get_collection_names` to `exclude_methods`).
- 1 sed step in `generate_async_client.sh` (AST-based, same pattern as #1294 / #1296) to re-inject the async `get_collection_names` into the facade and remote files after regen.
- 1 test file with ~20 tests (sync + async × local + remote × empty / single / multiple × regen survival).

Out of scope:
- A `prefix` filter parameter.
- `__contains__` and `__iter__` dunder methods.
- Returning sorted output by default.

## Acceptance criteria

- `QdrantClient(":memory:").get_collection_names()` returns `[]` for an empty database and a `list[str]` of all created collection names once any are added.
- `QdrantClient(path=...).get_collection_names()` works the same.
- `QdrantClient("http://localhost:6333", prefer_grpc=False).get_collection_names()` returns the same names as a direct `get_collections()` call.
- `QdrantClient("localhost:6334", prefer_grpc=True).get_collection_names()` works the same over gRPC.
- `await AsyncQdrantClient(...).get_collection_names()` returns a `list[str]`.
- For local mode, the order matches insertion order.
- The remote path does not add a network round-trip beyond the existing `get_collections()` call.
- The local path does not go through `get_collections()` (it reads `self.collections.keys()` directly).
- The `async_qdrant_client.py` and `async_qdrant_remote.py` changes survive `bash tools/generate_async_client.sh` without diff. The async version is re-injected by the AST-based sed step, same pattern as #1294 / #1296.
- 20+ tests in `tests/test_get_collection_names.py`, all pass alongside the existing suite.
- mypy clean on the 6 modified files.

## Backward compatibility

Fully additive. No existing public API change. The return type `list[str]` is new but does not shadow any existing method.

## Risks

- The async files are autogenerated, so the sed step in `generate_async_client.sh` is the only fragile part. Same AST-based pattern as #1294 / #1296 (fails loudly with non-zero exit if `close()` cannot be found).
- The local `get_collection_names()` does not check `closed` itself; it relies on the `closed` guard in the existing `get_collections()` path. For local mode, since we read `self.collections.keys()` directly, we need to add a small `closed` guard at the top of the local method (mirroring the one in `QdrantLocal.get_collections`).
- Low risk of name collision: `get_collection_names` is a new method name, distinct from the existing `get_collections` (plural) and `get_collection` (singular) methods.

## Notes for the implementer

- The test file follows the pattern in `tests/test_server_info.py` (pytest, no live server, mock the inner client for remote tests).
- The `inspect.iscoroutine` pattern from #1294 / #1296 is used in the async facade to handle both sync (local) and async (remote) inner clients.
- The local `get_collection_names()` reads `self.collections.keys()` directly (no `get_collections()` round-trip, no `types.CollectionsResponse` model construction). This is a deliberate optimization: the local `self.collections` is already a `dict[str, LocalCollection]`, so the list of keys is a single line.

Anai-Guo:
## Problem

Closes #1302.

`get_collections()` returns a `types.CollectionsResponse` whose `.collections` is a list of `CollectionDescription` objects. The common case — "I just want the names" — requires a list comprehension every time:

```python
names = [c.name for c in client.get_collections().collections]
```

This is verbose for what should be a one-liner (cleanup scripts, "which collections exist" checks, health dashboards), and `CollectionDescription` is more than a "list of names" query needs.

## Change

Add `get_collection_names()` as a thin convenience wrapper that returns `list[str]`, following the exact convention of the existing `collection_exists()` helper:

- abstract `QdrantBase` (`client_base.py`)
- concrete `QdrantRemote` and `QdrantLocal` — each derives the names from its own `get_collections()`, so gRPC/REST/local behave identically for free
- the `QdrantClient` facade delegates to `self._client.get_collection_names(...)`

The async variants (`AsyncQdrantBase`, `AsyncQdrantRemote`, `AsyncQdrantLocal`, `AsyncQdrantClient`) are **regenerated** via `tools/generate_async_client.sh` — not hand-edited — so `tests/async-client-consistency-check.sh` stays green.

## Test

`tests/congruence_tests/test_collections.py::test_get_collection_names` asserts local and remote return the same names, and that each equals the `.collections` name comprehension. Verified locally in `:memory:` mode for both the sync and async clients.

### DIFF
--- a/qdrant_client/async_client_base.py
+++ b/qdrant_client/async_client_base.py
@@ -224,6 +224,9 @@ class AsyncQdrantBase:
     async def get_collections(self, **kwargs: Any) -> types.CollectionsResponse:
         raise NotImplementedError()
 
+    async def get_collection_names(self, **kwargs: Any) -> list[str]:
+        raise NotImplementedError()
+
     async def get_collection(self, collection_name: str, **kwargs: Any) -> types.CollectionInfo:
         raise NotImplementedError()
 
--- a/qdrant_client/async_qdrant_client.py
+++ b/qdrant_client/async_qdrant_client.py
@@ -1517,6 +1517,15 @@ class AsyncQdrantClient(AsyncQdrantFastembedMixin):
         assert len(kwargs) == 0, f"Unknown arguments: {list(kwargs.keys())}"
         return await self._client.get_collections(**kwargs)
 
+    async def get_collection_names(self, **kwargs: Any) -> list[str]:
+        """Get names of all existing collections
+
+        Returns:
+            List of the collection names
+        """
+        assert len(kwargs) == 0, f"Unknown arguments: {list(kwargs.keys())}"
+        return await self._client.get_collection_names(**kwargs)
+
     async def get_collection(self, collection_name: str, **kwargs: Any) -> types.CollectionInfo:
         """Get detailed information about specified existing collection
 
--- a/qdrant_client/async_qdrant_remote.py
+++ b/qdrant_client/async_qdrant_remote.py
@@ -1712,6 +1712,11 @@ class AsyncQdrantRemote(AsyncQdrantBase):
         assert result is not None, "Get collections returned None"
         return result
 
+    async def get_collection_names(self, **kwargs: Any) -> list[str]:
+        return [
+            collection.name for collection in (await self.get_collections(**kwargs)).collections
+        ]
+
     async def get_collection(self, collection_name: str, **kwargs: Any) -> types.CollectionInfo:
         if self._prefer_grpc:
             return GrpcToRest.convert_collection_info(
--- a/qdrant_client/client_base.py
+++ b/qdrant_client/client_base.py
@@ -231,6 +231,9 @@ class QdrantBase:
     def get_collections(self, **kwargs: Any) -> types.CollectionsResponse:
         raise NotImplementedError()
 
+    def get_collection_names(self, **kwargs: Any) -> list[str]:
+        raise NotImplementedError()
+
     def get_collection(self, collection_name: str, **kwargs: Any) -> types.CollectionInfo:
         raise NotImplementedError()
 
--- a/qdrant_client/local/async_qdrant_local.py
+++ b/qdrant_client/local/async_qdrant_local.py
@@ -662,6 +662,11 @@ class AsyncQdrantLocal(AsyncQdrantBase):
             ]
         )
 
+    async def get_collection_names(self, **kwargs: Any) -> list[str]:
+        return [
+            collection.name for collection in (await self.get_collections(**kwargs)).collections
+        ]
+
     async def get_collection(self, collection_name: str, **kwargs: Any) -> types.CollectionInfo:
         collection = self._get_collection(collection_name)
         return collection.info()
--- a/qdrant_client/local/qdrant_local.py
+++ b/qdrant_client/local/qdrant_local.py
@@ -719,6 +719,9 @@ class QdrantLocal(QdrantBase):
             ]
         )
 
+    def get_collection_names(self, **kwargs: Any) -> list[str]:
+        return [collection.name for collection in self.get_collections(**kwargs).collections]
+
     def get_collection(self, collection_name: str, **kwargs: Any) -> types.CollectionInfo:
         collection = self._get_collection(collection_name)
         return collection.info()
--- a/qdrant_client/qdrant_client.py
+++ b/qdrant_client/qdrant_client.py
@@ -1576,6 +1576,16 @@ class QdrantClient(QdrantFastembedMixin):
 
         return self._client.get_collections(**kwargs)
 
+    def get_collection_names(self, **kwargs: Any) -> list[str]:
+        """Get names of all existing collections
+
+        Returns:
+            List of the collection names
+        """
+        assert len(kwargs) == 0, f"Unknown arguments: {list(kwargs.keys())}"
+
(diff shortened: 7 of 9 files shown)


joein:
Hey @Anai-Guo 

Thanks for the interest in the project!

Unfortunately, I don't think we'd like to add this method to the client in order to avoid bloating up the class with too many methods + it will make python client to provide more methods than the clients in other languages

Also, all the PRs should point to the `dev` branch instead of `master`, otherwise it's a bit tricky to bring the commit to dev branch

Closing it due to the above reasons 🙏

Anai-Guo:
Understood — keeping the client surface aligned across languages is a good reason, thanks for saying so plainly rather than letting it sit. Noted on `dev` as the base branch for anything I send in future.

PR #1308, opened 2026-08-03
fix: support FieldCondition.is_empty/is_null in local mode

sainikhiljuluri:
### All Submissions:

### Changes to Core Features:

---

## The problem

`FieldCondition` carries `is_empty` and `is_null` — the shorthand syntax for
`IsEmptyCondition` / `IsNullCondition`, described in the generated models as
"alternative syntax for `is_empty: 'field_name'`". In local mode they are ignored
**silently**, and the failure is total rather than partial:

```python
# ... 5 points, only some of which have a "reports" value

flt = models.Filter(must=[models.FieldCondition(key="reports", is_empty=True)])
client.scroll("c", scroll_filter=flt)      # -> []  (server: the 3 empty ones)

flt = models.Filter(must_not=[models.FieldCondition(key="reports", is_empty=True)])
client.scroll("c", scroll_filter=flt)      # -> every point in the collection
```

So `must` matches nothing and `must_not` matches everything — with no warning and
no `NotImplementedError`. It reaches `scroll`, `count`, `query_points`, `facet`
and `delete(filter=...)`, which means a `delete` guarded by `must_not=[...]` can
clear a whole local collection while doing almost nothing against a server.

## Root cause

`check_condition()`'s `FieldCondition` branch only ever inspected `match`,
`range`, `geo_bounding_box`, `geo_radius`, `values_count` and `geo_polygon`. A
condition carrying `is_empty=`/`is_null=` matches none of those `if`s, falls out
of the `elif` body and hits the trailing `return False` at the end of the
function — evaluating to `False` for every point.

## The fix

Two small helpers plus a branch in `check_condition`:

- a value is **empty** when it is `null` or an empty array; a key holding no
  value counts as empty but **not** null
- a value is **null** when it is `null` or an array containing a `null`
- for a key resolving to several values (`a[].b`), any one of them satisfying the
  condition is a match — so one point can satisfy both `is_empty=True` and
  `is_empty=False`

A condition that also carries `values_count` is left to the existing
`values_count` branch, so nothing about its behaviour changes.

Values are read with `flat=False`, like the two neighbouring `IsEmptyCondition`
and `IsNullCondition` branches. That is load-bearing rather than stylistic:
`{"field": []}` flattens to `None`, which would otherwise collapse the
empty-array case into the no-value case.

## How this was tested

The semantics above were **established by running the queries against a real
`qdrant/qdrant:dev` (1.18.3-dev) container, not by reading core** — an earlier
revision of this PR was written from the Rust source and got the `values_count`
interaction wrong in a way only the live server revealed.

- `tests/congruence_tests/test_is_empty_is_null.py` (new) — 16 points and 19
  filters diffed between local mode and a live server, covering a flat key, a
  nested key and an array-traversing key, both directions of each condition,
  `must_not`, and both verbose conditions. This follows the same pattern as
  #1190 / #1191 / #1224 / #1259.
- `qdrant_client/local/tests/test_payload_filters.py` — 4 new unit tests
  (72 → 76 collected in that directory).
- Full congruence suite against `qdrant/qdrant:dev`: **284 passed**, matching a
  clean `dev` baseline.
- `ruff-format --line-length=99` clean; `mypy` clean on `qdrant_client/local`.

## Risk and compatibility

No public API, signature, default or return-shape change — this is local-mode
internals, and the async local client picks it up unchanged. The only behaviour
that changes is for conditions that are currently broken. In principle someone
could be relying on `must_not=[FieldCondition(is_empty=True)]` matching
everything, but that is the bug rather than a contract.

## Two things worth your judgement

**1. Payload indexes.** Local mode does not model them (it warns as much), so
where the server's answer depends on whether a field is indexed, local mode can
only mirror one path. This PR mirrors the unindexed payload-scan path, which is
what the surrounding local-mode branches already do and what the congruence
fixtures exercise.

That matters in one visible place: on an **unindexed** field the server does not
treat `is_null` and `IsNullCondition` as interchangeable — the verbose condition
tests the values a key resolves to, so an array holding a null is not itself
null, while the shorthand looks inside it:

```
payload {"field": [None, 1]}
  FieldCondition(key="field", is_null=True)   -> matches
  IsNullCondition(key="field")                -> does not match
```

With a keyword index on the field, both match. `is_empty` and
`IsEmptyCondition` agree either way. Since the models call the two forms
alternative syntax, this looks like it may be a core inconsistency rather than
intended — happy to flip the client whenever core settles it, and the congruence
test means you would hear about it either way.

**2. Deliberately out of scope.** `check_condition` evaluates `match` / `range` /
geo *before* `values_count`, whereas the server prefers `values_count`. That
predates this PR and is unchanged by it, so I have not touched it. Glad to send a
separate PR if you want it fixed.

---

Found by comparing local mode against server behaviour rather than from a
reported issue, so there is no linked issue. AI-assisted: an agent wrote the
patch and tests, and every behavioural claim here was verified against a running
Qdrant container.

### DIFF
--- a/qdrant_client/local/payload_filters.py
+++ b/qdrant_client/local/payload_filters.py
@@ -45,6 +45,22 @@ def check_values_count(condition: models.ValuesCount, values: list[Any] | None) 
     return True
 
 
+def check_is_empty_value(is_empty: bool, value: Any) -> bool:
+    if value is None:
+        return is_empty
+    if isinstance(value, list):
+        return (len(value) == 0) == is_empty
+    return not is_empty
+
+
+def check_is_null_value(is_null: bool, value: Any) -> bool:
+    if value is None:
+        return is_null
+    if isinstance(value, list):
+        return any(v is None for v in value) == is_null
+    return not is_null
+
+
 def check_geo_radius(condition: models.GeoRadius, values: Any) -> bool:
     if isinstance(values, dict) and "lat" in values and "lon" in values:
         lat = values["lat"]
@@ -227,6 +243,19 @@ def check_condition(
         if condition.has_vector in has_vector and has_vector[condition.has_vector]:
             return True
     elif isinstance(condition, models.FieldCondition):
+        if condition.values_count is None and (
+            condition.is_empty is not None or condition.is_null is not None
+        ):
+            # values_count keeps its own branch below, this must not shadow it
+            raw_values = value_by_key(payload, condition.key, flat=False)
+            if not raw_values:
+                # nothing stored under the key: the server counts that as empty, not null
+                if condition.is_empty is not None:
+                    return condition.is_empty
+                return not condition.is_null
+            if condition.is_empty is not None:
+                return any(check_is_empty_value(condition.is_empty, v) for v in raw_values)
+            return any(check_is_null_value(condition.is_null, v) for v in raw_values)
         values = value_by_key(payload, condition.key)
         if condition.match is not None:
             if values is None:
--- a/qdrant_client/local/tests/test_payload_filters.py
+++ b/qdrant_client/local/tests/test_payload_filters.py
@@ -187,3 +187,152 @@ def test_geo_polygon_filter_query():
 
     res = check_filter(query, payload, 0, has_vector={})
     assert res is False
+
+
+def test_field_condition_is_empty():
+    # `FieldCondition.is_empty` is the shorthand syntax for `IsEmptyCondition`.
+    # A value is empty if it is null, an empty array, or the key is absent.
+    payloads = {
+        1: {"reports": [1, 2]},
+        2: {"reports": []},
+        3: {"reports": None},
+        4: {},
+        5: {"reports": [None, 1]},
+    }
+
+    def matches(is_empty: bool) -> list[int]:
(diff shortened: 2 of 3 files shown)


sainikhiljuluri:
Thanks for the review. Took one of the two:

**Compare more than `scroll`** — done in cfc6b95. The loop now compares `count()` alongside `scroll()` for all 19 filters, so the parity claim covers a second operation on the same filter path rather than resting on `scroll` alone. Verified green against `qdrant/qdrant:dev` (1.18.3-dev).

**Extract the `is_empty`/`is_null` evaluation into a helper** — skipping this one. `check_condition` deliberately handles every condition type inline in one `if`/`elif` chain, and the neighbouring `IsEmptyCondition`, `IsNullCondition`, `HasIdCondition` and `NestedCondition` branches are all written that way. Pulling just this branch out into a `bool | None`-returning helper would make it the odd one out, and the tri-state return is harder to follow than the guard it would replace. Happy to do it if a maintainer prefers the extracted shape.

joein:
Hey @sainikhiljuluri 

Thanks for fixing this!

It seems there might be some other issues involved, I'll return to your pr once we address them 🙏

joein (approved):

PR #1310, opened 2026-08-03
Fix float precision loss in datetime_to_microseconds

Nimra3261:
## Bug

`datetime_to_microseconds` computes `int(dt.timestamp() * 1_000_000)`. `timestamp()` returns a float64, and multiplying it by 1e6 before truncating amplifies the rounding error enough to be off by one microsecond for a large fraction of timestamps:

```python
>>> from datetime import datetime, timezone
>>> dt = datetime(1970, 7, 21, 14, 9, 16, 146413, tzinfo=timezone.utc)
>>> int(dt.timestamp() * 1_000_000)
17417356146412   # should be 17417356146413
```

In random testing across a range of years, ~35% of timestamps were off by one microsecond. Dates far enough from 1970 (e.g. year 9999) are off by a full **second**, because the whole-seconds part alone exceeds what float64 can represent precisely alongside the fractional part.

## Why it matters

This value feeds ORDER BY comparisons in local mode (`Filter` range/datetime conditions, `scroll` ordering). An off-by-one microsecond can flip a boundary comparison, causing local mode to disagree with what the real Qdrant server returns for the same query - a local/remote congruence bug.

## Fix

Route through datetime subtraction instead of `timestamp()`: subtracting two aware datetimes is exact integer arithmetic internally, no floats involved. A naive input is first attached to the system's local timezone via `astimezone()` - the same "assume local time" behavior `timestamp()` has for naive input - which itself returns a `datetime`, not a lossy float, so precision holds all the way through.

## Testing

Added `tests/test_order_by.py`, cross-checking against an independent exact-arithmetic reference implementation across a wide range of years/microseconds, including the year-9999 case. Verified the tests fail against the original implementation and against a first fix attempt that only replaced the multiplication but still used `timestamp()` for the whole-seconds part (that one still got year 9999 wrong by a full second).

Opened against `dev` since that's where the related recent local-mode fixes (#1259 etc.) landed.

### DIFF
--- a/qdrant_client/local/order_by.py
+++ b/qdrant_client/local/order_by.py
@@ -11,16 +11,21 @@ _EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
 def datetime_to_microseconds(dt: datetime) -> int:
     # Not `int(dt.timestamp() * MICROS_PER_SECOND)`: `timestamp()` returns a float, and
     # both the float itself and multiplying it by 1e6 before truncating lose precision -
-    # off by one microsecond for a large fraction of timestamps in testing, and off by a
-    # full second for dates far enough from 1970 (e.g. year 9999) that the whole-seconds
-    # part alone exceeds what float64 can hold alongside the fractional part.
+    # off by one microsecond for a large fraction of timestamps in testing. The error grows
+    # with distance from 1970 (float64 has 52 mantissa bits, so the representable precision
+    # shrinks as the magnitude grows): dates near year 9999 can be off by up to ~32
+    # microseconds, though never anywhere near a full second.
     #
     # `datetime - datetime` (both aware) is exact integer arithmetic internally (no
     # floats), so route through that instead. A naive `dt` is first attached to the
     # system's local timezone via `astimezone()` - the same "assume local time" behavior
     # `timestamp()` has for naive input - which itself returns a datetime, not a lossy
     # float, so precision is preserved all the way through.
-    if dt.tzinfo is None:
+    #
+    # `utcoffset() is None` (not `tzinfo is None`) is the correct naive-datetime check: a
+    # tzinfo subclass can be attached and still report no offset, and `dt - _EPOCH` raises
+    # TypeError if naive and aware datetimes are mixed.
+    if dt.utcoffset() is None:
         dt = dt.astimezone()
     delta = dt - _EPOCH
     return (
--- a/qdrant_client/local/order_by.py
+++ b/qdrant_client/local/order_by.py
@@ -1,13 +1,33 @@
-from datetime import datetime
+from datetime import datetime, timezone
 
 from qdrant_client.http.models import OrderValue
 from qdrant_client.local.datetime_utils import parse
 
 MICROS_PER_SECOND = 1_000_000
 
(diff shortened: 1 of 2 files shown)


shashvat-singham (commented):
Ran the branch. The precision fix itself checks out — cross-checking both implementations against exact integer arithmetic over 20k random timestamps in the 0–2e15µs range, the old `int(dt.timestamp() * MICROS_PER_SECOND)` is off by a microsecond on ~1.3% of them, and the new version matches exactly everywhere I tried, including pre-1970 dates where `timedelta` normalisation makes `days` negative and `seconds`/`microseconds` positive.

Two things:

**1. The comment overstates the far-future case.** It says the old code is "off by a full second for dates far enough from 1970 (e.g. year 9999)". For `datetime(9999, 12, 31, 23, 59, 59, 999999, tzinfo=utc)` the old code gives `253402300800000000` vs `253402300799999999` exact — off by 1µs, not 1s. float64 has 52 bits of mantissa and that value needs ~58 bits, so the ULP there is ~32µs, nowhere near a second. Worth toning down so the comment doesn't send the next reader hunting for a second-scale bug that isn't there.

**2. Preserving `timestamp()`'s naive-datetime behaviour keeps a machine-dependent inconsistency.** The comment justifies `dt.astimezone()` as matching the old "assume local time" behaviour for naive input, which it does. But that behaviour disagrees with how a naive datetime *string* is handled two functions over — `datetime_utils.parse` explicitly assumes **UTC** when there's no offset, matching qdrant core. So on this machine (IST, UTC+5:30):

```python
to_order_value("2024-06-15 12:30:45")          # 1718454645000000  (UTC)
to_order_value(datetime(2024, 6, 15, 12, 30, 45))  # 1718434845000000  (local)
```

Same wall-clock time, 19800000000µs apart, and the gap changes with the machine's timezone — so `order_by` results depend on where the client runs. The aware-UTC object agrees with the string, so it's specifically the naive-object path that diverges.

That's pre-existing and arguably out of scope, but this PR is rewriting exactly this function and pinning the behaviour in a comment as intentional, so it seems like the moment to decide. Making it `dt.replace(tzinfo=timezone.utc)` would line the object path up with the string path and with core. Happy to file it separately if you'd rather keep this PR purely about precision.

Nimra3261:
Thanks both — pushed an update addressing this.

@shashvat-singham:
1. Fixed the comment — verified empirically (both the exact year-9999 case and a 5k-sample sweep across years 9000–9999): the actual error is up to ~32µs, matching your float64-mantissa math, never a full second. Corrected the wording.
2. Agreed on scope — since #1352 already fixes the naive-as-local-vs-UTC inconsistency independently (and deliberately avoids touching `datetime_to_microseconds` to not conflict with this PR), I've left that out here. This PR stays focused on the precision fix.

@coderabbitai: fixed — now checking `dt.utcoffset() is None` instead of `dt.tzinfo is None`. Confirmed the old check actually crashes (`TypeError: can't subtract offset-naive and offset-aware datetimes`) on a tzinfo subclass that reports no offset, and added a regression test for it that fails on the old code and passes on the new.

joein:
Hey @Nimra3261, thanks for the PR!

I added your fixes to #1417  and set you as a co-author, closing this in favour of that pr

Issue #615, opened 2024-04-27
AsyncQdrantClient calls blocking event loop


amey-wynk:
Currently trying to utilise qdrant for a production usecase and this requires building an API for real time vector search. It is crucial for this to be async so that it can scale well for multiple users. But when using the async client with FastAPI and Uvicorn, latency increases as number of concurrent users increase.

## Current Behavior
Latency of the endpoint increases as concurrent users increase.

## Steps to Reproduce

```
from fastapi import FastAPI
import asyncio
from qdrant_client import AsyncQdrantClient

app = FastAPI()
asyncClient = AsyncQdrantClient(url = 'prod.qdrant.com', port = 80)

@app.post('/sleep')
async def sleep(request: Request) -> dict:
    await asyncio.sleep(0.1) # simulate 100ms I/O call
    return {'message': 'success'}

@app.post('/recommend')
async def recommend(request: Request) -> dict:
    await asyncClient.recommend(
        collection_name = 'my_collection',
        positive = ['4be0643f1d98573b97cdca98a65347dd'],
        limit = 10,
    )
    return {'message': 'success'}
```

1. Run the above app using uvicorn
2. Load test the endpoint using [locust](https://github.com/locustio/locust)
3. The /sleep endpoint scales to 50 concurrent users while maintaining a p99 latency of 110ms (as expected since the requests are concurrent)
4.The /recommend endpoint does not do this. At 1 concurrent users the p99 latency is 10ms, 65ms at 10 users and 260ms for 50 concurrent users. (this is unexpected if the requests are processed concurrently)

Uvicorn version: 0.27.1
FastAPI version: 0.110.0
Qdrant version: 1.7.3 (both server and client)
Qdrant is deployed on k8s with 10 pods and 20vCPUs each. All vectors are in memory and utilization of pods is <1 vCPU during testing.

## Expected Behavior
The requests should be processed concurrently. Maybe the async call using the client is blocking the event loop and preventing FastAPI from processing other requests

generall:
> Latency of the endpoint increases as concurrent users increase.

This is kinda expected. You can't scale indefinitely by just adding parallel calls.

But also considering that 

> memory and utilization of pods is <1 vCPU during testing.

there might be two possibilities, depending on your configuration it of the collection: either bottleneck is in disk, or bottleneck is on the client side.

Could you please try to do the same with gRPC?

amey-wynk:
Thanks for the response @generall 

[...]

PR #1328, opened 2026-08-09
fix: local mode AsyncQdrantClient blocks the event loop

codechrl:
## What

`AsyncQdrantLocal`'s read methods (`query_points`, `query_groups`, `scroll`, `count`,
`facet`, `search_matrix_offsets`, `search_matrix_pairs`, `retrieve`) are all `async def`
but their bodies just call the matching synchronous `LocalCollection` method with no
`await` inside — so the full duration of a brute-force numpy scan runs inline on the
event loop instead of yielding. Any other coroutine sharing that loop (another request,
a background task, a heartbeat) stalls for as long as the scan takes.

Fixes the local-mode half of #615.

## Why

#615 was originally filed against the remote client, and a very thorough recent
[comment from @ErenAta16](https://github.com/qdrant/qdrant-client/issues/615#issuecomment-3113595668)
correctly pins that specific report on REST/gRPC response-decode cost, not on I/O
blocking — that's a separate, harder problem this PR does **not** touch. But the same
comment also flags, almost in passing, that "in local mode the loop really is blocked"
and points at `AsyncQdrantLocal.query_points` having no `await` in its body. That part
was still open, so this PR fixes it.

## How

`qdrant_client/local/async_qdrant_local.py` is generated from
`qdrant_client/local/qdrant_local.py` by `tools/async_client_generator/local_generator.py`
— hand-editing the generated file directly would just get overwritten on the next
`bash -x tools/generate_async_client.sh`. So the fix lives in the generator:

- `LocalCallTransformer` gets a small `TO_THREAD_METHODS` whitelist. When it sees a call
  to one of those method names on something other than `self` (i.e. the `LocalCollection`
  instance, not a recursive call to another async client method), it rewrites the call
  into `await asyncio.to_thread(<original call>)` instead of leaving it bare.
- New `LocalEnsureImportTransformer` prepends `import asyncio` to the generated module,
  since the sync source has no reason to import it itself.
- Regenerated `qdrant_client/local/async_qdrant_local.py` and hand-verified the diff
  matches what the generator now produces (a couple of unrelated `for (a, b) in ...` →
  `for a, b in ...` `ast.unparse` formatting differences were reverted to keep this diff
  scoped to the actual fix — those come from a Python-version difference in `ast.unparse`
  output, not from this change).

`update_vectors`/`delete_vectors`/`upsert`/etc. (write paths) are deliberately **not**
touched here — #1196 is already adding locking to `LocalCollection` for concurrent
writes, and moving writes onto a thread pool before that lands would reintroduce the
exact race #1193 describes.

## Testing

New test `tests/test_async_in_memory.py::test_async_local_query_points_does_not_block_event_loop`:
upserts 20k points into an in-memory collection, then runs a `query_points` loop
concurrently with a `asyncio.sleep(0.01)` heartbeat loop via `asyncio.gather`. Confirmed
test-first — against the unfixed code the heartbeat gets delayed to ~0.5s while a scan
runs; with the fix it stays close to the scheduled 0.01s.

Also ran the full `tests/test_in_memory.py` and `tests/test_local_persistence.py` suites
locally — all pass, no regressions.


codechrl:
Confirmed real, thanks for catching this. Traced through `_add_point`: several non-atomic steps (`self.ids_inv.append`, `self.payload.append`, `self.deleted = np.append(...)`, per-vector `np.resize`/reassignment) with no locking. Before this PR, reads and writes never truly ran in parallel -- everything was single-threaded on the event loop with no yield points, so a blocking read accidentally serialized against writes. Offloading reads to `asyncio.to_thread` breaks that: a write on the event loop thread can now interleave with a read on a worker thread mid-mutation.

A complete fix needs a lock shared between offloaded reads and writes -- but #1196 is already adding `threading.RLock` to `LocalCollection` for write-write safety, and explicitly leaves reads unlocked because the old single-threaded assumption held. I'd rather not invent a second, possibly-conflicting locking scheme here. Proposing: land this PR as the event-loop-blocking fix it's scoped as, then once #1196 merges, wire the offloaded reads into that same lock as a fast follow-up. Happy to open that follow-up PR once #1196 is in.

PR #1333, opened 2026-08-12
fix: local mode matches the wrong points on null and empty-`should` filters

nazsats:
Two divergences between local mode and the server, both returning the wrong point set silently
— no error, no warning. Found by running identical filters against local mode and a real Qdrant
1.19.0 server and diffing the matched ids.

## 1. `MatchExcept` matches an explicit `null`

A null payload value is absence, not a value that happens to differ from everything in the list.
`MatchExcept` is the only negated match condition, so it is the only one where treating null as
an ordinary value flips the result:

```python
return not any(values_match(value, v) for v in condition.except_)
```

For `value = None` every `values_match` is False, so `not any(...)` is True and the point
matches. The server drops nulls before matching and does not.

Every sibling condition already guards its input — `MatchText` and `MatchPrefix` check
`isinstance(value, str)`, `MatchValue` goes through `values_match`, which compares types. Only
the negated one has no guard, and negation is exactly where a missing guard changes the answer.

```python
# points: {"n": 1}, {"n": 5}, {"n": None}, {"n": [1, None]}
flt = models.Filter(must=[models.FieldCondition(
    key="n", match=models.MatchExcept(**{"except": [1]}))])

local.scroll("t", scroll_filter=flt)   # [2, 3, 4]   <- 3 and 4 should not be here
remote.scroll("t", scroll_filter=flt)  # [2]
```

Reproduces identically for a null `int`, a null `str` and a null array — the payload type is
irrelevant, the null is what matters.

## 2. An empty `should` matches nothing instead of everything

`check_should` is `any(...)`, and `any([])` is False, so `should=[]` rejects every point. The
server reads an empty `should` as no constraint and returns every point — the exact inverse.

`must=[]` and `must_not=[]` are both `all(...)`, vacuously True, and already agree with the
server. Only `should` disagrees, and only because `any` and `all` differ on the empty list.

```python
flt = models.Filter(should=[])
local.scroll("t", scroll_filter=flt)   # []
remote.scroll("t", scroll_filter=flt)  # every point
```

This bites callers building filters programmatically: a `should` list assembled from a loop that
happens to append nothing silently returns zero results instead of being ignored.

## Tests

Added to `tests/congruence_tests/test_scroll.py`, following the existing
`test_bool_int_float_filters_do_not_cross_match` pattern — a shared `_init_*_collection` helper,
`compare_client_results` against a live server, parametrised over `prefer_grpc`:

- `test_match_except_does_not_match_null_payload_values` — nulls as int, str and array, plus
  `[1, None]` where 1 is excluded, and a point missing the key entirely.
- `test_empty_should_is_not_a_constraint` — includes a **non-empty** `should` case, so the fix
  cannot be "ignore `should`".

Both fail on `dev` and pass with this change.

## Verification

- `pytest tests/congruence_tests/test_scroll.py` → 10 passed.
- Reverting `payload_filters.py` alone and re-running the two new tests → 4 failed, confirming
  they pin the behaviour rather than passing vacuously.
- Full `pytest tests/congruence_tests/` against a live 1.19.0 server → 285 passed. The 6
  failures are `test_search_with_persistence*`, all `PermissionError: [WinError 32]` from
  `TemporaryDirectory` cleanup on Windows; they fail identically on unmodified `dev` and none is
  an `AssertionError`.
- `ruff format --line-length=99` clean.

## Not included

`min_should.min_count=0` is rejected by the server with a 422
(`min_count must be greater than 0`) and silently accepted by local mode, where it matches
everything. That is an input-validation gap rather than a wrong-results bug and belongs in its
own change — most likely a `Field(ge=1)` constraint on the model rather than a local-mode patch.
Happy to open it separately if you'd like it.

---

**AI assistance:** I used AI assistance to help probe local-vs-server behaviour and draft this
description. The diagnosis, the fix and the tests were reviewed and verified by me against a
live Qdrant server.


joein:
Hello, @nazsats 

Unfortunately, the PR is too difficult to review cause it points to `master` branch
Could you please recreate your PR and point it to `dev` branch instead?

nazsats:
Thanks @joein ; already sorted, no recreate needed.

I've retargeted this PR to dev, so it now shows just the two files:

qdrant_client/local/payload_filters.py (+11/−4)
tests/congruence_tests/test_scroll.py (+97/−0)
Apologies for the noise — GitHub defaulted the base to master and I didn't catch it before opening.

joein (approved):

Issue #1280, opened 2026-07-23
LocalCollection: `limit=0` silently treated as absent in `_query_no_prefetch()`


tsushanth:
## Bug

In `qdrant_client/local/local_collection.py`, line ~906:

```python
limit = limit or 10
```

The falsy `or` guard treats `limit=0` identically to `limit=None`. A caller that explicitly passes `limit=0` expecting zero results (e.g. to check scores without fetching payloads, or to mirror a server response that returned no results) instead receives up to 10 results.

The same line also appears for `offset`:
```python
offset = offset or 0
```
which is fine because 0 is the correct fallback, but the `limit` case is a bug.

## Reproduction

```python
from qdrant_client import QdrantClient, models

client = QdrantClient(":memory:")
client.create_collection("test", vectors_config=models.VectorParams(size=4, distance=models.Distance.COSINE))
client.upsert("test", points=[
    models.PointStruct(id=i, vector=[float(i), 0.0, 0.0, 0.0]) for i in range(5)
])

# Expect 0 results; actually returns up to 10
results = client.query_points("test", query=[1.0, 0.0, 0.0, 0.0], limit=0)
print(len(results.points))  # prints 5, not 0
```

## Fix

```python
# Before
limit = limit or 10
# After
limit = limit if limit is not None else 10
```

This matches how the Qdrant server itself handles `limit=0` (returns an empty list) and is consistent with the `None`-means-unset convention used throughout the client.

joein:
Hey @tsushanth 

Thanks for reporting this!

Just out of curiosity, what's your use case?

joein:
I could not reproduce this behaviour with the Qdrant server

Instead, it returns:

```
b'{"status":{"error":"Validation error in JSON body: [scroll_request.limit: value 0 invalid, must be 1 or larger]"},"time":0.0}'
```

joein:
#1281 

the fix will be available with the next release

PR #1339, opened 2026-08-14
fix: validate local query parameters

adityat210:
### All Submissions:

* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

### Changes to Core Features:

## Problem

Local mode accepted several invalid query bounds that the server rejects:

* negative Query API offsets could silently truncate the result set
* zero nested-prefetch limits were treated as the default limit
* zero group sizes produced groups with no hits
* invalid matrix limits and samples returned empty responses

The server schemas define Query API `offset >= 0` and `limit >= 1`, grouped-query `group_size >= 1`, and matrix `limit >= 1` / `sample >= 2`. See the [Query points](https://api.qdrant.tech/api-reference/search/query-points/), [Query point groups](https://api.qdrant.tech/api-reference/search/query-points-groups), and [Distance matrix offsets](https://api.qdrant.tech/master/api-reference/search/matrix-offsets) references.

This follows the local/server validation parity work in #1280 and #1281, which covered top-level positive limits but not these remaining bounds.

## Fix

* validate query offsets and grouped-query sizes before local execution
* validate matrix limits and sample sizes for both response formats
* recursively validate limits across nested prefetch trees
* regenerate the async local client from the canonical sync implementation
* cover sync, async, batch-query, and nested-prefetch paths

Valid requests are unchanged. Invalid local requests now fail early with `ValueError` instead of returning misleading results.

## Impact (X-Y-Z)

* **X — before:** 5 invalid-input classes were accepted across local query and matrix APIs.
* **Y — after:** all affected sync and async entry points reject those values before collection access, including batch and nested-prefetch dispatch.
* **Z — coverage:** 15 new regression assertions cover 8 sync cases and 7 async cases; the implementation adds only constant-time checks plus traversal of the already-provided prefetch tree.

## Validation

* Python 3.10: focused sync/async regression suite (`9 passed`)
* Python 3.11: full in-memory suite (`12 passed`)
* conversion coverage suite (`22 passed`, `99%` coverage)
* async-client generation consistency check
* `mypy`, `pyright`, local collection doctest, Ruff, and all pre-commit hooks


shashvat-singham (commented):
Ran the branch — `tests/test_in_memory.py` is 12 passed, and I couldn't find a hole in the coverage.

Since the whole point here is matching server behaviour, I cross-checked each new bound against the validation attributes in `lib/api/src/rest/schema.rs` rather than assuming:

| check | this PR | qdrant core |
|---|---|---|
| `limit` | `< 1` | `#[validate(range(min = 1))]` |
| `sample` | `< 2` | `#[validate(range(min = 2))]` |
| `group_size` | `< 1` | `#[validate(range(min = 1))]` |
| prefetch `limit` | `< 1` | `#[validate(range(min = 1))]` |
| `offset` | `< 0` | `Option<usize>` |

All five line up. `sample < 2` was the one I expected to be off — it's the only non-obvious constant — but core really does require `min = 2`, so it's right.

I also probed for paths the recursion might miss, and they're all covered:

```
batch: offset=-1           -> ValueError
batch: prefetch limit=0    -> ValueError
groups: prefetch limit=0   -> ValueError
nested prefetch limit=0    -> ValueError   # prefetch(limit=5, prefetch=prefetch(limit=0))
scroll(limit=0)            -> ValueError
```

The recursive `_validate_prefetch_limits` handling both the single-`Prefetch` and `list[Prefetch]` shapes is what makes the nested and batch cases work, and it's easy to get that wrong given `prefetch` is overloaded that way in the models.

One deliberate behaviour change worth calling out in the description: validation now runs *before* `_get_collection`, so `query_points("missing", offset=-1)` raises `ValueError: offset value -1 is invalid` where it previously raised collection-not-found. The tests encode that by using `"missing"` as the collection name. I think it's correct — the server validates the request body before it looks the collection up, so a 400 beats a 404 there, and it matches where the pre-existing `limit < 1` check already sat — but it is a visible change for anyone catching the old error, so it's worth being explicit rather than leaving it implied by the test fixtures.

Nothing blocking from me.

joein:
thanks for adding this!

joein (approved):

PR #1345, opened 2026-08-15
fix(order_by): read a naive datetime as UTC, not local time

### LINKED ISSUE #1342: order_by treats a naive datetime object as local time, but a naive datetime string as UTC
### Describe the bug

`order_by` converts a datetime to microseconds differently depending on whether you pass it as a string or as a `datetime` object, when neither carries a timezone. The same wall-clock time gets two different order values, and the gap is whatever the client machine's UTC offset happens to be.

On a machine in IST (UTC+5:30):

```python
>>> from datetime import datetime, timezone
>>> from qdrant_client.local.order_by import to_order_value

>>> to_order_value("2024-06-15 12:30:45")                          # naive string
1718454645000000
>>> to_order_value(datetime(2024, 6, 15, 12, 30, 45))              # naive object
1718434845000000
>>> to_order_value(datetime(2024, 6, 15, 12, 30, 45, tzinfo=timezone.utc))
1718454645000000
```

The string and the aware-UTC object agree. The naive object is 19800 seconds off — exactly the machine's offset. Run the same code in UTC and the discrepancy disappears; run it in UTC-8 and it flips sign. So local-mode ordering depends on where the client runs.

### Cause

The two paths make opposite assumptions about what "no timezone" means.

`datetime_utils.parse`, used for strings, explicitly assumes UTC — matching qdrant core, whose accepted formats treat a missing offset as UTC:

```python
if dt.tzinfo is None:
    # Assume UTC if no timezone is provided
    dt = dt.replace(tzinfo=timezone.utc)
```

`to_order_value` passes a `datetime` object straight to `datetime_to_microseconds`, which calls `dt.timestamp()`. For a naive datetime, `timestamp()` interprets it as **local time**, not UTC.

### Expected behavior

A naive `datetime` object should be interpreted the same way as a naive datetime string — as UTC — so that ordering is reproducible regardless of the client's timezone and consistent with the server.

Concretely, `datetime_to_microseconds` attaching `timezone.utc` to a naive input (rather than letting `timestamp()` assume local time) would line the two paths up.

### Note

#1310 is currently rewriting `datetime_to_microseconds` to fix a separate float-precision problem, and deliberately preserves the naive-input behaviour via `dt.astimezone()` to match what `timestamp()` did. I raised this there and said I'd file it separately so it's tracked on its own — the two are independent, and whichever way #1310 goes this question still needs an answer.

### Environment

- qdrant-client @ `dev` (f003e6c)
- Python 3.11.9, Windows, machine timezone UTC+5:30

Aryan-Pardeshi:
Fixes #1342

Replaces #1344, which I branched from `master` by mistake and so carried unrelated commits against a `dev` base. Same change, clean history off `dev`.

`to_order_value()` passed `datetime` objects straight to `datetime.timestamp()`, which interprets a naive datetime as **local** time. `datetime_utils.parse()` assumes **UTC** for a datetime string with no offset, matching qdrant core, so the same wall-clock time produced two different order values — off by whatever the client machine's UTC offset happens to be.

On a UTC+5:30 machine, before:

```python
>>> to_order_value("2024-06-15 12:30:45")               # naive string
1718454645000000
>>> to_order_value(datetime(2024, 6, 15, 12, 30, 45))   # naive object
1718434845000000   # 19800s off, exactly the machine offset
```

`datetime_to_microseconds()` now attaches `timezone.utc` to a naive input, mirroring the `parse()` branch one function over. Datetimes that carry an offset are unaffected and still converted, not stripped.

### Tests

New `tests/test_order_by.py`: naive object resolves to the UTC value; naive object / naive string / aware-UTC all agree; an aware non-UTC datetime is still converted by its offset (so "naive means UTC" did not become "everything is UTC"); and the same call returns the same value under `TZ=UTC`, `Asia/Kolkata` and `America/Los_Angeles` (skipped where `time.tzset` is unavailable).

### Relation to #1310

#1310 rewrites `datetime_to_microseconds` for a separate float-precision problem and deliberately preserves the naive-input behaviour via `dt.astimezone()`. Independent changes to a one-line region — happy to rebase on top of #1310 if you'd prefer that order.

### DIFF
--- a/qdrant_client/local/order_by.py
+++ b/qdrant_client/local/order_by.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 
 from qdrant_client.http.models import OrderValue
 from qdrant_client.local.datetime_utils import parse
@@ -7,6 +7,12 @@ MICROS_PER_SECOND = 1_000_000
 
 
 def datetime_to_microseconds(dt: datetime) -> int:
+    if dt.tzinfo is None:
+        # Assume UTC if no timezone is provided, matching `datetime_utils.parse`
+        # and qdrant core. `datetime.timestamp()` would otherwise read a naive
+        # datetime as local time, making order values depend on where the
+        # client runs.
+        dt = dt.replace(tzinfo=timezone.utc)
     return int(dt.timestamp() * MICROS_PER_SECOND)
 
 
--- a/tests/test_order_by.py
+++ b/tests/test_order_by.py
@@ -1,0 +1,58 @@
+import os
+import time
+from datetime import datetime, timedelta, timezone
+
+import pytest
+
+from qdrant_client.local.order_by import to_order_value
+
+# 2024-06-15 12:30:45 UTC
+WALL_CLOCK = (2024, 6, 15, 12, 30, 45)
+EXPECTED_MICROS = 1718454645000000
+
+
+def test_naive_datetime_object_is_interpreted_as_utc() -> None:
+    """A datetime with no tzinfo means UTC, matching how qdrant core reads a
+    datetime string with no offset."""
+    assert to_order_value(datetime(*WALL_CLOCK)) == EXPECTED_MICROS
+
+
+def test_naive_object_string_and_aware_utc_agree() -> None:
+    """The same wall-clock time must order identically however it is spelled."""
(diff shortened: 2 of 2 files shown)


joein:
thank you for the fix!

though, I think this might be a better fix  #1417

Issue #1351, opened 2026-08-16
KeywordIndexParams.prefix is the only optional bool that does not survive a REST to gRPC round trip


ErenAta16:
`KeywordIndexParams.prefix` is the only optional bool in the index-params family that does not survive a REST to gRPC round trip: an explicit `False` comes back as `None`.

I am filing this as a question rather than a bug report, because the answer may be "the gRPC schema cannot express it", in which case the useful outcome is a docstring rather than a code change. I could not verify the server-side half, so I am not claiming a behavioural impact.

## Measured

`rest -> grpc -> rest` over every optional bool field in the three index-params models, on `master` at `550484d`:

```
KeywordIndexParams
   is_tenant         True -> True  | False -> False |  None -> None
   on_disk           True -> True  | False -> False |  None -> None
   enable_hnsw       True -> True  | False -> False |  None -> None
   prefix            True -> True  | False -> None   LOST |  None -> None

IntegerIndexParams
   lookup, range, is_principal, on_disk, enable_hnsw    all preserved

TextIndexParams
   lowercase, ascii_folding, phrase_matching, on_disk, enable_hnsw    all preserved
```

Fifteen fields, fourteen preserved, one not.

## Why

The two directions disagree about what "unset" means:

```python
# rest -> grpc, conversion.py:4386
prefix=grpc.KeywordPrefixParams() if model.prefix else None

# grpc -> rest, conversion.py:1692
prefix=True if model.HasField("prefix") else None
```

REST types it as a tri-state `Optional[bool]` with `default=None` and the description "If true, enable prefix matching ... Default: false", while gRPC types it as an optional `KeywordPrefixParams` message, so presence is the only signal. `False` and `None` both become absent, and absent becomes `None`.

The immediate asymmetry is that the outbound line uses truthiness where its neighbours in the same call use `is not None`:

```python
memory=cls.convert_memory(model.memory) if model.memory is not None else None,
```

but changing only that line would send a `KeywordPrefixParams()` message for `prefix=False`, which would be worse. The real gap is that the gRPC message has no field to carry the value.

## Why the existing suite does not catch it

`tests/conversions/test_validate_conversions.py` walks `grpc -> rest -> grpc`. Starting from a gRPC fixture, `prefix` is either present or absent, so the REST tri-state is not expressible in the input and the case cannot arise. Running the reverse direction is what surfaced it. The suite is otherwise green here, 22 passed.

## The question

Which of these is intended?

[...]

PR #1356, opened 2026-08-18
fix: pin keyword index prefix=False round-trip behavior over gRPC

2sumtech:
Addresses #1351 (following @ErenAta16's option 1).

`grpc.KeywordPrefixParams` is an empty message (`proto/collections.proto:705`): presence is the only signal it can carry, so an explicit `prefix=False` cannot be represented in gRPC. `RestToGrpc.convert_keyword_index_params` sends it as absent — which the server also treats as disabled, per the documented default — and `GrpcToRest` recovers it as `None`.

Since the value is unrepresentable without a proto change, this PR documents the limitation at both conversion sites and pins the behavior with a reverse-direction (rest→grpc→rest) test covering `True`/`None`/`False`. The REST models are code-generated, so a docstring note there wouldn't survive regeneration. If maintainers would rather the proto carry the value, that would be an upstream follow-up in qdrant/qdrant.

**Tests:** `pytest tests/conversions/ -q` → 23 passed (22 before). `ruff format --check --line-length=99` clean on touched files.

Disclosure: prepared with AI assistance (Claude Code); I reviewed the change and take responsibility for it.

### DIFF
--- a/qdrant_client/conversions/conversion.py
+++ b/qdrant_client/conversions/conversion.py
@@ -1689,6 +1689,8 @@ class GrpcToRest:
             on_disk=model.on_disk if model.HasField("on_disk") else None,
             memory=cls.convert_memory(model.memory) if model.HasField("memory") else None,
             enable_hnsw=model.enable_hnsw if model.HasField("enable_hnsw") else None,
+            # presence of grpc.KeywordPrefixParams is the only signal, an explicit `prefix=False`
+            # cannot be represented in grpc and comes back as `None`
             prefix=True if model.HasField("prefix") else None,
         )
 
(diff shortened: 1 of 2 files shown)


joein (dismissed):
thank you for adding this!

joein:
hey @2sumtech 

thanks for adding this!
Could you please recreate the PR and point it to the `dev` branch, not `master`? Or would u like me to do this?

2sumtech:
done — rebased onto dev and retargeted, conversion tests green. and thanks for landing #1364!

joein (approved):

joein:
thank you!

PR #1357, opened 2026-08-18
Fix in-place mutation of inputs in cosine_similarity

aravgarg28:
### All Submissions:

* [x] Contributions should target the `dev` branch. Did you create your branch from `dev`? Yes.
* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change? Yes — no open PR touches `qdrant_client/local/distances.py`.

### What & why

`cosine_similarity` in `qdrant_client/local/distances.py` normalizes its `query` and `vectors` arguments **in place** using `/=`:

```python
vectors /= np.where(vectors_norm != 0.0, vectors_norm, EPSILON)
...
query /= np.where(query_norm != 0.0, query_norm, EPSILON)
```

Because `ndarray /= ...` is an in-place operation, this has two side effects that the sibling distance functions (`dot_product`, `euclidean_distance`, `manhattan_distance`) do not have — they build new arrays and never touch their inputs:

1. **It mutates caller-owned arrays.** The `query`/`vectors` a caller passes in are silently overwritten with their normalized values. Today this is masked in the client only because callers happen to hand in freshly-copied / already-normalized arrays, but the function contract is still wrong: a `calculate_*`/similarity helper should not rewrite its inputs. The existing test even documented this with `# cosine modifies vectors inplace`.

2. **It raises on integer-dtype inputs.** `int_array /= float_array` raises `numpy.core._exceptions.UFuncTypeError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int64')`, so `cosine_similarity` crashes on an integer array where `dot`/`euclid`/`manhattan` succeed.

The fix switches the three `/=` to out-of-place `x = x / ...`, which rebinds the local names to new arrays and leaves the inputs untouched. **Computed results are identical** (verified against the existing `test_distances` expectations), and integer inputs now work.

### How is this tested?

* Updated `test_distances` (removed the two now-obsolete `# cosine modifies vectors inplace` comments).
* Added `test_cosine_does_not_mutate_inputs` — asserts `query` and `vectors` are unchanged after a call, for both the 1D and 2D (multivector-style) query paths, and that results are still correct.
* Added `test_cosine_accepts_integer_dtype_inputs` — asserts integer-dtype inputs no longer raise.

Both new tests fail on `dev` (mutation assertion + `UFuncTypeError`) and pass with this change. Ran locally:

* `pytest qdrant_client/local/tests/test_distances.py tests/conversions/ tests/test_in_memory.py` → all pass, no regressions.
* `ruff format --check --line-length=99` on the changed files → clean.
* `mypy` with the repo's flags on the changed files → clean.

### Changes to Core Features:

### DIFF
--- a/qdrant_client/local/distances.py
+++ b/qdrant_client/local/distances.py
@@ -127,8 +127,12 @@ def cosine_similarity(query: types.NumpyArray, vectors: types.NumpyArray) -> typ
     Returns:
         distances
     """
+    # `vectors` is normalized in place: through the normal client API it is
+    # always already unit-normalized on cosine collections (normalized on
+    # upsert), so this is a no-op in practice, and avoiding the copy matters
+    # since `vectors` is the (potentially large) candidate set being searched.
     vectors_norm = np.linalg.norm(vectors, axis=-1)[:, np.newaxis]
-    vectors = vectors / np.where(vectors_norm != 0.0, vectors_norm, EPSILON)
+    vectors /= np.where(vectors_norm != 0.0, vectors_norm, EPSILON)
 
     if len(query.shape) == 1:
         query_norm = np.linalg.norm(query)
--- a/qdrant_client/local/distances.py
+++ b/qdrant_client/local/distances.py
@@ -128,15 +128,15 @@ def cosine_similarity(query: types.NumpyArray, vectors: types.NumpyArray) -> typ
         distances
     """
     vectors_norm = np.linalg.norm(vectors, axis=-1)[:, np.newaxis]
-    vectors /= np.where(vectors_norm != 0.0, vectors_norm, EPSILON)
+    vectors = vectors / np.where(vectors_norm != 0.0, vectors_norm, EPSILON)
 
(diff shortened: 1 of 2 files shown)


joein:
Hey @aravgarg28 

Qdrant normalizes the vectors on upsert, so the right-hand operand here (`vectors`) is kinda always normalized if used via the proper API (e.g. `client.query_points(...)`. so this change here will just create a copy of the already normalized array. 
I can agree with the query change, since the overhead is minimal.

aravgarg28:
Thanks @joein you're right. I checked and confirmed cosine collections normalize both dense and multivector storage at upsert (`local_collection.py`), so `vectors` is always already unit-normalized through the client API by the time it reaches this function.

Pushed 46c7c8d: `vectors` is back to in-place normalization, `query` stays copied. Updated the tests to match: dropped the `vectors`-immutability and integer-dtype assertions for `vectors` (not a real-world case), kept them for `query`.

joein (approved):

PR #1359, opened 2026-08-19
fix: do not drop shard keys in delete and payload operations

### LINKED ISSUE #1358: delete/set_payload and other write operations silently apply to all shards when shard key is falsy (grpc) or embedded in the selector (REST)
**Environment:** qdrant-client current master (550484d), Python 3.11.

With custom sharding, several write operations drop the shard key and get applied to **all shards** in two situations:

**1. `prefer_grpc=True` and a falsy shard key.** `0` and `""` are valid shard keys (`ShardKey = StrictInt | StrictStr`, proto `uint64`), but `delete`, `delete_vectors`, `set_payload`, `overwrite_payload`, `delete_payload` and `clear_payload` merge them with `shard_key_selector or opt_shard_key_selector` (`qdrant_client/qdrant_remote.py`, e.g. the `delete` method), so the gRPC request is sent without any `shard_key_selector`:

```python
from unittest.mock import MagicMock
from qdrant_client import QdrantClient, models, grpc

client = QdrantClient(prefer_grpc=True)
stub = MagicMock()
stub.Delete.return_value = grpc.PointsOperationResponse(
    result=grpc.UpdateResult(operation_id=1, status=grpc.UpdateStatus.Completed), time=0.0)
client._client._grpc_points_client_pool = [stub]

client.delete("c", models.PointIdsList(points=[1, 2]), shard_key_selector=0)
request = stub.Delete.call_args[0][0]
print("shard_key_selector set:", request.HasField("shard_key_selector"))
```

Actual: `shard_key_selector set: False` — the delete runs on every shard. With `shard_key_selector=1`: `True`. REST mode sends `shard_key: 0` for the identical call, so the same code does different things on different transports.

**2. REST mode and a shard key embedded in the points selector.** `client.delete("c", models.PointIdsList(points=[...], shard_key="us"))` keeps the shard key under `prefer_grpc=True` (extracted by `_try_argument_to_grpc_selector`), but in REST mode `_try_argument_to_rest_selector` overwrites it with `None` (and `_try_argument_to_rest_points_and_filter` never forwards it for `delete_vectors`/`set_payload`/`overwrite_payload`/`delete_payload`), so the operation again hits all shards.

This is the same bug class as #1260/#1279 (falsy shard keys dropped in conversion), but in the client methods rather than the conversion layer. Happy to submit a PR — fix is replacing the truthiness fallback with an `is None` check and forwarding the embedded shard key on the REST paths.

2sumtech:
Fixes #1358

Write operations (`delete`, `delete_vectors`, `set_payload`, `overwrite_payload`, `delete_payload`, `clear_payload`) could silently run on **all shards** instead of the selected one: with `prefer_grpc=True` a falsy `shard_key_selector` (`0`, `""`) was discarded by `shard_key_selector or opt_shard_key_selector`, and in REST mode a shard key embedded in the points selector (`PointIdsList(shard_key=...)`) was overwritten with `None` or not forwarded, while gRPC mode honored it.

**Changes:** replace the truthiness fallback with an explicit `is None` check; make `_try_argument_to_rest_selector` preserve the embedded shard key unless an explicit selector is given; return the embedded shard key from `_try_argument_to_rest_points_and_filter` and use it as fallback in its callers. Precedence is now identical on both transports: explicit argument (including falsy) > embedded > none. Same bug class as #1260/#1279, which fixed it in the conversion layer — these are the remaining sites in the client methods.

**Async client:** regenerated with `tools/generate_async_client.sh` under Python 3.10 (not hand-edited); `tests/async-client-consistency-check.sh` → "No diffs found" for all generated files.

**Tests:** new `tests/test_shard_key_selector.py` — parametrized over all six methods × falsy keys `0`/`""` asserting the gRPC request carries the shard key, plus embedded-key preservation, explicit-override, and REST-path tests. 15 fail without the fix, all pass with it; `tests/conversions/` 39 passed total; `ruff format --check --line-length=99` clean; mypy clean on touched files.

Disclosure: prepared with AI assistance (Claude Code); I reviewed the change and take responsibility for it.

### DIFF
--- a/qdrant_client/async_qdrant_remote.py
+++ b/qdrant_client/async_qdrant_remote.py
@@ -1108,7 +1108,8 @@ class AsyncQdrantRemote(AsyncQdrantBase):
     ) -> types.UpdateResult:
         if self._prefer_grpc:
             (points_selector, opt_shard_key_selector) = self._try_argument_to_grpc_selector(points)
-            shard_key_selector = shard_key_selector or opt_shard_key_selector
+            if shard_key_selector is None:
+                shard_key_selector = opt_shard_key_selector
             if isinstance(ordering, models.WriteOrdering):
                 ordering = RestToGrpc.convert_write_ordering(ordering)
             if isinstance(shard_key_selector, get_args_subscribed(models.ShardKeySelector)):
@@ -1130,7 +1131,9 @@ class AsyncQdrantRemote(AsyncQdrantBase):
             assert grpc_result is not None, "Delete vectors returned None result"
             return GrpcToRest.convert_update_result(grpc_result)
         else:
-            (_points, _filter) = self._try_argument_to_rest_points_and_filter(points)
+            (_points, _filter, _shard_key) = self._try_argument_to_rest_points_and_filter(points)
+            if shard_key_selector is None:
+                shard_key_selector = _shard_key
             return (
                 await self.openapi_client.points_api.delete_vectors(
                     collection_name=collection_name,
@@ -1260,7 +1263,8 @@ class AsyncQdrantRemote(AsyncQdrantBase):
             points_selector.shard_key = shard_key_selector
         elif isinstance(points, get_args(models.PointsSelector)):
             points_selector = points
-            points_selector.shard_key = shard_key_selector
+            if shard_key_selector is not None:
+                points_selector.shard_key = shard_key_selector
         elif isinstance(points, models.Filter):
             points_selector = construct(
                 models.FilterSelector, filter=points, shard_key=shard_key_selector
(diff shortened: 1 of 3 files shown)


2sumtech:
Added parameterized REST-boundary tests covering all six mutation operations — embedded shard keys and explicit falsy keys (`0`/`""`) asserted at the generated request for each: 18 new cases, 34 total in the file, all passing.

2sumtech:
Good catch — the falsy-override tests now run against a selector carrying an embedded `shard_key="us"`, so they assert precedence (explicit `0`/`""` wins over embedded) rather than mere propagation. 34 tests still green.

joein:
Hey @2sumtech 

Thank your for addressing this issue!

Unfortunately, merging PRs pointed to `master` is complicated, all the new branches have to be pointed at `dev` instead, I'll have to recreate your PR.

I did some refactoring and will create it soon, I'll set you as co-author 👍

PR #1361, opened 2026-08-19
fix(snapshots): fix get_snapshots trying to parse tar as JSON

### LINKED ISSUE #1360: get_snapshot fails as it tries to parse snapshot as json
# Issue
The HTTP `get_snapshot` API wrapper always fails due to JSON parsing error.

# How to reproduces
1. Create a collection snapshot
2. Use `AsyncQdrantClient.http.snapshots_api.get_snapshot`
3. Get an error similar to `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 3073: invalid start byte`

# Trace/Cause
`qdrant_client.http.api.snapshots_api.AsyncSnapshotsApi.get_snapshot` calls `qdrant_client.http.api_client.request` which in turn calls `qdrant_client.http.api_client.send`, which always parses the response as JSON:
https://github.com/qdrant/qdrant-client/blob/550484d767d319857d4f46e97d4551ba419ee670/qdrant_client/http/api_client.py#L219

But snapshots are tar files, not JSON, thus `get_snapshot` will always fail.

(Also, maybe related: Why is `file` set to `None`? https://github.com/qdrant/qdrant-client/blob/550484d767d319857d4f46e97d4551ba419ee670/qdrant_client/http/api/snapshots_api.py#L15)

bvolkmer:
This PR fixes #1360.

The fix is rather simple, but I'm not sure if it is the best way to do it, as I have no in-depth insights into the project. I assumed `file=None` has no specific meaning
and thus changed it to `file=bytes` and added a special handling for `type_==bytes` that passes the response content
through without trying to parse as JSON.

`pytest` passes with:
```
563 passed
29 skipped
17927 warnings
```

A simple local roundtrip test worked: `client.create_snapshot` -> `client.http.snapshots_api.get_snapshot` -> `client.http.snapshots_api.recover_from_uploaded_snapshot`

### DIFF
--- a/qdrant_client/http/api/snapshots_api.py
+++ b/qdrant_client/http/api/snapshots_api.py
@@ -12,7 +12,7 @@ Model = TypeVar("Model", bound="BaseModel")
 
 SetIntStr = Set[Union[int, str]]
 DictIntStrAny = Dict[Union[int, str], Any]
-file = None
+file = bytes
 
 
 def to_json(model: BaseModel, *args: Any, **kwargs: Any) -> str:
--- a/qdrant_client/http/api_client.py
+++ b/qdrant_client/http/api_client.py
@@ -126,6 +126,8 @@ class ApiClient:
                 raise ResourceExhaustedResponse(message, retry_after_s)
 
         if response.status_code in [200, 201, 202]:
+            if type_ is bytes:
+                return response.content
             try:
                 return parse_as_type(response.json(), type_)
             except ValidationError as e:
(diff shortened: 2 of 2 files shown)


joein:
Hey @bvolkmer 

Thanks for the attempt to address this issue, however, there are a couple of problems with it:

1) we don't expose `get_snapshot` api  in the clients at the moment, instead we're proposing to use the pure rest api https://qdrant.tech/documentation/snapshots/?q=snapshot#retrieve-snapshot
2) the files you changed are auto-generated and in order to change them properly - it has to be done in another private repository

I'll close this for now, but thank you for the interest!

PR #1366, opened 2026-08-21
fix(http): propagate per-call timeout to httpx in AsyncApiClient

### LINKED ISSUE #1325: AsyncQdrantClient (REST): per-call timeout= is not applied to the underlying httpx client
## Summary

`AsyncQdrantClient`'s REST transport silently drops the per-call `timeout=` argument at the HTTP-client level, so async REST operations that ask for extra time (e.g. `await client.create_collection(..., timeout=30)`) are still bound by httpx's default 5-second connect/read/write/pool timeout instead. The sync `QdrantClient` does not have this problem.

## Environment

- qdrant-client version: current `dev`/`master` (commit `550484d767d319857d4f46e97d4551ba419ee670`), also present in the latest PyPI release — the code path is unchanged back to 2024
- Python: 3.12.10
- httpx: 0.28.1
- `prefer_grpc`: default (`False`) — this affects the default REST transport, not just an opt-in path

## Reproduction

Both `ApiClient` (sync) and `AsyncApiClient` (async) build a request the same way from `qdrant_client/http/api_client.py`, but only the sync one promotes a per-call `timeout` found in the query params up to httpx's own client-side timeout. This is directly observable without a running server, by inspecting the kwargs each one hands to `build_request`:

```python
import asyncio
from unittest.mock import patch
from qdrant_client.http.api_client import ApiClient, AsyncApiClient

captured = {}

def make_capture(target):
    def _capture(self, method, url, **kwargs):
        captured[target] = kwargs
        class Dummy: pass
        return Dummy()
    return _capture

sync_client = ApiClient(host="http://localhost:6333")
with patch.object(sync_client._client, "build_request", make_capture("sync").__get__(sync_client._client)):
    try:
        sync_client.request(type_=None, method="GET", url="/collections/{c}",
                             path_params={"c": "x"}, params={"timeout": "50"})
    except Exception:
        pass

async_client = AsyncApiClient(host="http://localhost:6333")
with patch.object(async_client._async_client, "build_request", make_capture("async").__get__(async_client._async_client)):
    try:
        asyncio.run(async_client.request(type_=None, method="GET", url="/collections/{c}",
                                          path_params={"c": "x"}, params={"timeout": "50"}))
    except Exception:
        pass

print("sync build_request kwargs: ", captured["sync"])
print("async build_request kwargs:", captured["async"])
```

Output:
```
sync build_request kwargs:  {'params': {'timeout': '50'}, 'timeout': 50}
async build_request kwargs: {'params': {'timeout': '50'}}
```

The sync call ends up passing `timeout=50` to httpx's `build_request`, overriding httpx's default. The async call passes no `timeout` at all, so httpx's `Timeout(timeout=5.0)` default applies regardless of what the caller asked for.

## Expected behavior

A per-call `timeout=` on any REST method (e.g. `create_collection`, `query_points`, `upsert`, ...) should give the underlying httpx client at least that much time, for both `QdrantClient` and `AsyncQdrantClient` — this is exactly what `tests/test_qdrant_client.py::test_timeout_propagation` asserts for the sync client.

## Actual behavior

Only the sync client does this. Concretely, in `qdrant_client/http/api_client.py`:

```python
# ApiClient.request (sync) — present
if "params" in kwargs and "timeout" in kwargs["params"]:
    kwargs["timeout"] = int(kwargs["params"]["timeout"])
```

`AsyncApiClient.request` builds the request the same way (`urljoin`, `build_request(method, url, **kwargs)`) but has no equivalent block, so a slow operation that legitimately needs more than 5 seconds can fail with a spurious httpx `ReadTimeout`/`ResponseHandlingException` under `AsyncQdrantClient` even though the caller explicitly asked for more time via `timeout=`.

## Analysis

This looks like sync/async drift rather than an intentional difference: the override block was added to the sync `ApiClient.request` by #534 ("propagate timeout from methods to httpx") and never ported to `AsyncApiClient.request`, which lives in the same file. There's a regression test for the sync path (`tests/test_qdrant_client.py:1740`, `test_timeout_propagation`) but no async equivalent in `tests/test_async_qdrant_client.py`.

This is related to but distinct from #948 ("Unexpected Timeout Behavior in AsyncQdrantClient") — that issue and its in-progress fix are about the **gRPC** transport (the protobuf `timeout` field not being populated), and a comment there explicitly scopes the HTTP path out: _"HTTP path is out of scope for this PR — same `timeout=timeout` pattern, but `httpx` treats `None` as 'no timeout' which is a different bug; that one needs a separate discussion."_ This issue is that separate discussion, for the REST/HTTP transport specifically.

## Proposed fix

Add the same `if "params" in kwargs and "timeout" in kwargs["params"]: kwargs["timeout"] = int(kwargs["params"]["timeout"])` block to `AsyncApiClient.request` before it calls `build_request`, plus an async regression test mirroring `test_timeout_propagation`. Happy to open a PR for this.

Harsh23Kashyap:
Fixes #1325

## Problem

`AsyncApiClient.request` (REST) dropped the per-call `timeout=` argument before handing the request to `httpx.AsyncClient.build_request`, so async callers were silently bound by httpx's default 5s timeout regardless of what they asked for. Slow operations on `AsyncQdrantClient` could fail with a spurious `ReadTimeout` even when the caller explicitly set `timeout=`.

The sync `ApiClient.request` already promotes the timeout up to httpx (added in #534). The async path was never ported.

Repro from the issue:

```python
sync build_request kwargs:  {'params': {'timeout': '50'}, 'timeout': 50}
async build_request kwargs: {'params': {'timeout': '50'}}
```

## Fix

`qdrant_client/http/api_client.py` — `AsyncApiClient.request` now mirrors the sync block:

```python
if "params" in kwargs and "timeout" in kwargs["params"]:
    kwargs["timeout"] = int(kwargs["params"]["timeout"])
```

Two lines, same logic as the sync client.

## Test

`tests/test_qdrant_client.py::test_async_rest_timeout_propagation` — a pure-unit regression that patches `httpx.AsyncClient.build_request` and asserts the timeout kwarg reaches it. Fails on master, passes with the fix. No live server required, so it runs in CI without docker.

The existing sync `test_timeout_propagation` (line 1740) already covers the sync path against a live server.

## Scope

- 1 production change (2 lines in `AsyncApiClient.request`)
- 1 new test (~28 lines)
- 1 import added (`from unittest.mock import patch`)

No API changes, no dependency changes, no behavior changes for the existing path. Distinct from #948 (gRPC timeout propagation), which is out of scope here.

### DIFF
--- a/qdrant_client/http/api_client.py
+++ b/qdrant_client/http/api_client.py
@@ -183,6 +183,8 @@ class AsyncApiClient:
         # in order to do a correct join, url join requires base_url to end with /, and url to not start with /,
         # since url is treated as an absolute path and might truncate prefix in base_url
         url = urljoin(host, url.format(**path_params))
+        if "params" in kwargs and "timeout" in kwargs["params"]:
+            kwargs["timeout"] = int(kwargs["params"]["timeout"])
         request = self._async_client.build_request(method, url, **kwargs)
         return await self.send(request, type_)
 
--- a/tests/test_qdrant_client.py
+++ b/tests/test_qdrant_client.py
@@ -8,6 +8,7 @@ import uuid
 from pprint import pprint
 from tempfile import mkdtemp
 from time import sleep
+from unittest.mock import patch
 
 import numpy as np
 import pytest
@@ -1755,6 +1756,35 @@ def test_timeout_propagation():
     client.create_collection(
         collection_name=COLLECTION_NAME, vectors_config=vectors_config, timeout=10
     )
+
+
+def test_async_rest_timeout_propagation():
+    # Regression for #1325: AsyncApiClient.request dropped the per-call
+    # `timeout=` from kwargs before build_request, so async callers were
+    # bound by httpx's default 5s timeout regardless of what they asked
+    # for. The sync ApiClient.request already promotes that value.
+    from qdrant_client.http.api_client import AsyncApiClient
+
+    captured: dict = {}
+
+    def _capture(method, url, **kwargs):
+        captured["kwargs"] = kwargs
+        return type("DummyRequest", (), {"headers": {}})()
+
+    async_client = AsyncApiClient(host="http://localhost:6333")
+    with patch.object(async_client._async_client, "build_request", _capture):
+        with pytest.raises(Exception):
+            asyncio.run(
+                async_client.request(
+                    type_=None,
+                    method="GET",
+                    url="/collections/{c}",
+                    path_params={"c": "x"},
+                    params={"timeout": "50"},
+                )
+            )
+
+    assert captured["kwargs"].get("timeout") == 50
 
 
 def test_grpc_options():
--- a/tests/test_qdrant_client.py
+++ b/tests/test_qdrant_client.py
@@ -1759,11 +1759,11 @@ def test_timeout_propagation():
 
 
 def test_async_rest_timeout_propagation():
-    # Regression for #1325: AsyncApiClient.request dropped the per-call
-    # `timeout=` from kwargs before build_request, so async callers were
-    # bound by httpx's default 5s timeout regardless of what they asked
-    # for. The sync ApiClient.request already promotes that value.
+    # Regression for #1325: async client dropped the per-call `timeout=`
+    # from kwargs before build_request, leaving callers bound by httpx's
+    # default 5s timeout. Sync ApiClient.request already promotes it.
     from qdrant_client.http.api_client import AsyncApiClient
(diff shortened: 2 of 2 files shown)


joein:
Hey @Harsh23Kashyap 

Thanks for proposing a fix!

Unfortunately, files under `http` are auto-generated and have to be fixed in a private repo, I'll fix it there and create a new PR

Also, all the PRs have to target `dev` branch.

I'll close this, but thank you for the attempt!

PR #1371, opened 2026-08-22
fix(local): compare score_threshold in the query's own score direction (#1370)

### LINKED ISSUE #1370: score_threshold is applied backwards for Recommend/Discover/Context queries on Euclidean or Manhattan collections (local mode)
### Describe the bug

In local/in-memory mode, `score_threshold` is compared in the wrong direction for **Recommend** (`best_score` / `sum_scores` strategy), **Discover** and **Context** queries whenever the collection's vector distance is **Euclidean** or **Manhattan**. This makes `query_points` (and the `recommend`/`discover` local implementations that feed it) silently return an empty (or truncated) result set even when the threshold should trivially be satisfied.

### Root cause

In `qdrant_client/local/local_collection.py`, `LocalCollection.search()`:

```python
required_order = distance_to_order(distance)

if required_order == DistanceOrder.BIGGER_IS_BETTER or isinstance(
    query_vector,
    (
        DiscoveryQuery,
        ContextQuery,
        RecoQuery,
        MultiDiscoveryQuery,
        MultiContextQuery,
        MultiRecoQuery,
    ),  # sparse structures are not required, sparse always uses DOT
):
    order = np.argsort(scores)[::-1]
else:
    order = np.argsort(scores)

...

if score_threshold is not None:
    if required_order == DistanceOrder.BIGGER_IS_BETTER:
        if score < score_threshold:
            break
    else:
        if score > score_threshold:
            break
```

Recommend (`best_score`/`sum_scores`), Discover and Context queries always compute a sigmoid-based synthetic score where **bigger is always better**, regardless of the collection's underlying distance metric - `calculate_recommend_best_scores`, `calculate_discovery_scores`, `calculate_context_scores` (and their multi-vector counterparts) all pass their raw distances through `scaled_fast_sigmoid`/`fast_sigmoid`. This is exactly why the sort-order check above special-cases these query types with an `isinstance(...)` override on top of `required_order`.

The `score_threshold` check a few lines below does not repeat that override - it branches solely on `required_order`, which is derived only from `distance_to_order(distance)` (`EUCLID`/`MANHATTAN` => `SMALLER_IS_BETTER`). So on a Euclidean/Manhattan collection, the threshold ends up compared with the operator meant for raw distances, even though the actual score in hand is the "bigger is better" synthetic one. Since points are iterated best-first, the loop `break`s immediately on the very first (best-scoring) point, discarding the whole result set.

### Reproduction (verified against current `master`)

`qdrant_client/local/local_collection.py` was fetched fresh from `https://raw.githubusercontent.com/qdrant/qdrant-client/master/qdrant_client/local/local_collection.py` and diffed against the file used for this reproduction - identical (only CRLF/LF whitespace differences). Reproduced with `qdrant-client==1.17.1` from PyPI, in local/in-memory mode (no server required).

Minimal repro:

```python

client.create_collection(
    "recommend_euclid",
    vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID),
)
client.upsert(
    "recommend_euclid",
    points=[
        models.PointStruct(id=1, vector=[0.0, 0.0]),
        models.PointStruct(id=2, vector=[1.0, 0.0]),
        models.PointStruct(id=3, vector=[2.0, 0.0]),
        models.PointStruct(id=4, vector=[3.0, 0.0]),
        models.PointStruct(id=5, vector=[10.0, 0.0]),
    ],
)
query = models.RecommendQuery(
    recommend=models.RecommendInput(positive=[1], strategy=models.RecommendStrategy.BEST_SCORE)
)

print(client.query_points("recommend_euclid", query=query, limit=10).points)
print(client.query_points("recommend_euclid", query=query, limit=10, score_threshold=0.05).points)
```

Output:

```
Recommend (BEST_SCORE) on a EUCLID collection, no score_threshold:
  id=2 score=0.25
  id=3 score=0.09999999403953552
  id=4 score=0.050000011920928955
  id=5 score=0.004950493574142456

Same query with score_threshold=0.05:
  -> 0 result(s) returned
  Expected: the points scoring >= 0.05 (ids 2, 3, 4) should be kept.
  Actual:   0 results - the loop breaks on the very first (best) point.
```

The same mismatch reproduces for a Discover query on a Manhattan collection:

```
Discover query on a MANHATTAN collection, no score_threshold:
  id=4 score=1.2272727191448212
  id=5 score=1.1666666567325592

Same query with score_threshold=-100:
  -> 0 result(s) returned
  Expected: both points should trivially pass such a low threshold.
  Actual:   0 results.
```

As a control, the identical `BEST_SCORE` recommend query against a Cosine collection (already bigger-is-better, so `required_order` happens to match) behaves correctly:

```
Control: same BEST_SCORE recommend query on a COSINE collection,
score_threshold=0.5 (bigger-is-better metric, no mismatch possible):
  id=2 score=0.7492331266403198
  id=3 score=0.5497245788574219
  -> 2 result(s) returned, correctly non-empty.
```

Full standalone reproduction script (self-contained, only needs `qdrant-client` installed):

```python


def recommend_on_euclidean():
    client.create_collection(
        "recommend_euclid",
        vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID),
    )
    client.upsert(
        "recommend_euclid",
        points=[
            models.PointStruct(id=1, vector=[0.0, 0.0]),
            models.PointStruct(id=2, vector=[1.0, 0.0]),
            models.PointStruct(id=3, vector=[2.0, 0.0]),
            models.PointStruct(id=4, vector=[3.0, 0.0]),
            models.PointStruct(id=5, vector=[10.0, 0.0]),
        ],
    )
    query = models.RecommendQuery(
        recommend=models.RecommendInput(
            positive=[1], strategy=models.RecommendStrategy.BEST_SCORE
        )
    )

    no_threshold = client.query_points("recommend_euclid", query=query, limit=10).points
    print("Recommend (BEST_SCORE) on a EUCLID collection, no score_threshold:")
    for p in no_threshold:
        print(f"  id={p.id} score={p.score}")

    threshold = 0.05  # permissive: 3 of the 4 candidates score >= 0.05
    filtered = client.query_points(
        "recommend_euclid", query=query, limit=10, score_threshold=threshold
    ).points
    print(f"\nSame query with score_threshold={threshold}:")
    print(f"  -> {len(filtered)} result(s) returned")


def discover_on_manhattan():
    client.create_collection(
        "discover_manhattan",
        vectors_config=models.VectorParams(size=2, distance=models.Distance.MANHATTAN),
    )
    client.upsert(
        "discover_manhattan",
        points=[
            models.PointStruct(id=1, vector=[0.0, 0.0]),
            models.PointStruct(id=2, vector=[1.0, 0.0]),
            models.PointStruct(id=3, vector=[5.0, 0.0]),
            models.PointStruct(id=4, vector=[1.2, 0.0]),
            models.PointStruct(id=5, vector=[2.0, 0.0]),
        ],
    )
    query = models.DiscoverQuery(
        discover=models.DiscoverInput(
            target=1, context=[models.ContextPair(positive=2, negative=3)]
        )
    )

    no_threshold = client.query_points("discover_manhattan", query=query, limit=10).points
    print("\nDiscover query on a MANHATTAN collection, no score_threshold:")
    for p in no_threshold:
        print(f"  id={p.id} score={p.score}")

    threshold = -100
    filtered = client.query_points(
        "discover_manhattan", query=query, limit=10, score_threshold=threshold
    ).points
    print(f"\nSame query with score_threshold={threshold}:")
    print(f"  -> {len(filtered)} result(s) returned")


if __name__ == "__main__":
    recommend_on_euclidean()
    discover_on_manhattan()
```

### Expected behavior

For Recommend (`best_score`/`sum_scores`), Discover and Context queries, `score_threshold` should always filter out points whose synthetic score is below the threshold (bigger-is-better semantics), regardless of the collection's underlying distance metric - matching the same `isinstance(...)` override that already governs the sort direction a few lines above.

### Actual behavior

On a Euclidean or Manhattan collection, `score_threshold` uses the comparison direction meant for raw distances instead, so the result-building loop breaks on the very first (best-scoring) candidate, and the query returns far fewer results than it should - typically zero, for any threshold that isn't extreme.

### Environment

- `qdrant-client` 1.17.1 (installed from PyPI)
- Verified the exact code (`qdrant_client/local/local_collection.py`) is unchanged on current `master` by diffing against the file fetched fresh from `https://raw.githubusercontent.com/qdrant/qdrant-client/master/qdrant_client/local/local_collection.py`
- Reproduced entirely in local/in-memory mode (`QdrantClient(":memory:")`), no server needed

Anai-Guo:
Fixes #1370.

## Problem

In local/in-memory mode, `score_threshold` is compared in the wrong direction for **Recommend** (`best_score` / `sum_scores`), **Discover** and **Context** queries whenever the collection's distance is **Euclid** or **Manhattan**. The threshold silently drops every point that should have passed, so `query_points` returns an empty result set.

`LocalCollection.search()` decides the sort order like this:

```python
required_order = distance_to_order(distance)

if required_order == DistanceOrder.BIGGER_IS_BETTER or isinstance(
    query_vector,
    (DiscoveryQuery, ContextQuery, RecoQuery,
     MultiDiscoveryQuery, MultiContextQuery, MultiRecoQuery),
):
    order = np.argsort(scores)[::-1]
else:
    order = np.argsort(scores)
```

The `isinstance(...)` arm is there because `calculate_recommend_best_scores`, `calculate_recommend_sum_scores`, `calculate_discovery_scores` and `calculate_context_scores` all build their score on top of `calculate_distance_core`, which already negates Euclid/Manhattan, and then run it through `scaled_fast_sigmoid` / `fast_sigmoid`. For these query types bigger is always better, whatever the collection's distance is.

The threshold check a few lines below did not repeat that override:

```python
if score_threshold is not None:
    if required_order == DistanceOrder.BIGGER_IS_BETTER:   # <-- ignores query type
        if score < score_threshold:
            break
    else:
        if score > score_threshold:
            break
```

On a Euclid/Manhattan collection `required_order` is `SMALLER_IS_BETTER`, so the points — already sorted biggest-first — are cut with `score > score_threshold`, and the very first point breaks the loop.

## Fix

Hoist the decision into a single `bigger_is_better` flag and use it for both the sort order and the threshold comparison, so the two cannot drift apart again. The flag is the exact expression the sort branch already used, so ordering behaviour is unchanged for every query type (including `NaiveFeedbackQuery`, which keeps following `required_order` as before).

## Reproduction

```python

c = QdrantClient(":memory:")
c.create_collection("t", vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID))
c.upsert("t", points=[models.PointStruct(id=i, vector=v) for i, v in
         [(1, [1.0, 0.0]), (2, [0.9, 0.1]), (3, [0.0, 1.0]), (4, [-1.0, 0.0])]], wait=True)

q = models.RecommendQuery(recommend=models.RecommendInput(
    positive=[[1.0, 0.0]], negative=[[-1.0, 0.0]],
    strategy=models.RecommendStrategy.BEST_SCORE))

print([(p.id, p.score) for p in c.query_points("t", query=q, limit=10).points])
# [(1, 0.5), (2, 0.4902), (3, -0.1667), (4, -0.5)]

print([p.id for p in c.query_points("t", query=q, limit=10, score_threshold=-0.51).points])
# before: []            <- threshold below every score, yet nothing comes back
# after:  [1, 2, 3, 4]
```

## Verification

Run against `dev`, for `Cosine` / `Euclid` / `Manhattan`:

| check | before | after |
|---|---|---|
| Recommend `best_score`, threshold below every score | Cosine ok; **Euclid/Manhattan return `[]`** | all three return every point |
| Recommend, thresholds that should cut (`0.45`, `0.0`, `-0.2`) | n/a (returned nothing) | matches `score >= threshold` exactly |
| Discover / Context on Euclid, threshold below every score | returned `[]` | returns every point |
| Plain dense search on Euclid (`smaller is better`) | matches `score <= threshold` | unchanged, still matches |

`ruff-format --line-length=99` (v0.4.3, per `.pre-commit-config.yaml`) reports the file already formatted. `tests/test_in_memory.py` and `tests/conversions` pass; the 4 failures in `tests/test_local_persistence.py` are a pre-existing Windows tempfile-locking artifact and reproduce identically on an unmodified tree.

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -654,7 +654,10 @@ class LocalCollection:
 
         required_order = distance_to_order(distance)
 
-        if required_order == DistanceOrder.BIGGER_IS_BETTER or isinstance(
+        # Recommend/discovery/context queries score through a sigmoid, so bigger is
+        # always better for them, whatever the collection's distance is. Both the sort
+        # order and the score_threshold comparison below have to follow this flag.
+        bigger_is_better = required_order == DistanceOrder.BIGGER_IS_BETTER or isinstance(
             query_vector,
             (
                 DiscoveryQuery,
@@ -664,7 +667,9 @@ class LocalCollection:
                 MultiContextQuery,
                 MultiRecoQuery,
             ),  # sparse structures are not required, sparse always uses DOT
-        ):
+        )
+
+        if bigger_is_better:
             order = np.argsort(scores)[::-1]
         else:
             order = np.argsort(scores)
@@ -684,7 +689,7 @@ class LocalCollection:
             point_id = self.ids_inv[idx]
 
             if score_threshold is not None:
-                if required_order == DistanceOrder.BIGGER_IS_BETTER:
+                if bigger_is_better:
                     if score < score_threshold:
                         break
                 else:


bijay-odyssey:
I independently reproduced the bug from #1370 and verified this fix against current `dev`.

Confirmed with both repros from the issue — Recommend (`BEST_SCORE`) on a Euclidean collection, and Discover on a Manhattan collection — plus a Cosine control case to make sure the fix doesn't regress the already-correct path (Cosine is bigger-is-better already, so no metric/score mismatch was ever possible there). `bigger_is_better` is the right minimal fix: the sort-order branch and the `score_threshold` comparison need to agree on direction, and pulling the shared condition into one variable is exactly what keeps them from drifting apart again.

One gap: neither this PR nor #1374 includes a regression test. Every other recent local-mode fix in this file shipped with one (#1424, #1427, #1402), and this specific bug class (a threshold silently returning zero results) is easy to reintroduce without a test catching it. Here's one that fails on unpatched `dev` and passes with this fix applied, if useful:

```python
import pytest

@pytest.fixture
def qdrant() -> QdrantClient:
    return QdrantClient(":memory:")


def test_recommend_score_threshold_on_euclidean_collection(qdrant: QdrantClient):
    qdrant.create_collection(
        "recommend_euclid",
        vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID),
    )
    qdrant.upsert(
        "recommend_euclid",
        points=[
            models.PointStruct(id=1, vector=[0.0, 0.0]),
            models.PointStruct(id=2, vector=[1.0, 0.0]),
            models.PointStruct(id=3, vector=[2.0, 0.0]),
            models.PointStruct(id=4, vector=[3.0, 0.0]),
            models.PointStruct(id=5, vector=[10.0, 0.0]),
        ],
    )
    query = models.RecommendQuery(
        recommend=models.RecommendInput(positive=[1], strategy=models.RecommendStrategy.BEST_SCORE)
    )

    unfiltered = qdrant.query_points("recommend_euclid", query=query, limit=10).points
    assert [p.id for p in unfiltered] == [2, 3, 4, 5]

    filtered = qdrant.query_points(
        "recommend_euclid", query=query, limit=10, score_threshold=0.05
    ).points
    assert [p.id for p in filtered] == [2, 3, 4]


def test_discover_score_threshold_on_manhattan_collection(qdrant: QdrantClient):
    qdrant.create_collection(
        "discover_manhattan",
        vectors_config=models.VectorParams(size=2, distance=models.Distance.MANHATTAN),
    )
    qdrant.upsert(
        "discover_manhattan",
        points=[
            models.PointStruct(id=1, vector=[0.0, 0.0]),
            models.PointStruct(id=2, vector=[1.0, 0.0]),
            models.PointStruct(id=3, vector=[5.0, 0.0]),
            models.PointStruct(id=4, vector=[1.2, 0.0]),
            models.PointStruct(id=5, vector=[2.0, 0.0]),
        ],
    )
    query = models.DiscoverQuery(
        discover=models.DiscoverInput(target=1, context=[models.ContextPair(positive=2, negative=3)])
    )

    unfiltered = qdrant.query_points("discover_manhattan", query=query, limit=10).points
    assert len(unfiltered) > 0

    filtered = qdrant.query_points(
        "discover_manhattan", query=query, limit=10, score_threshold=-100
    ).points
    assert [p.id for p in filtered] == [p.id for p in unfiltered]


def test_recommend_score_threshold_on_cosine_collection_still_works(qdrant: QdrantClient):
    qdrant.create_collection(
        "recommend_cosine",
    )
    qdrant.upsert(
        "recommend_cosine",
        points=[
            models.PointStruct(id=1, vector=[1.0, 0.0]),
            models.PointStruct(id=2, vector=[0.9, 0.1]),
            models.PointStruct(id=3, vector=[0.1, 0.9]),
        ],
    )
    query = models.RecommendQuery(
        recommend=models.RecommendInput(positive=[1], strategy=models.RecommendStrategy.BEST_SCORE)
    )

    filtered = qdrant.query_points(
        "recommend_cosine", query=query, limit=10, score_threshold=0.6
    ).points
    assert [p.id for p in filtered] == [2]
```

Also flagging for triage: #1374 looks like an independent duplicate of this same fix, opened a few days later.

PR #1373, opened 2026-08-23
fix: apply root filters after local fusion

### LINKED ISSUE #1372: Local mode ignores top-level `query_filter` on fusion and formula queries with prefetches
## Current Behavior

In local mode (`QdrantClient(":memory:")` or `path=...`), a top-level `query_filter` is ignored when the query has prefetches and the root query is a fusion or formula query. The results come back the same as running with no filter. The same call against a server filters correctly.

Moving the filter into each `Prefetch(filter=...)` works in local mode.

## Steps to Reproduce

1. Run the script below. Against `:memory:` it prints `[0, 1, 2]`.
2. Change the client line to `QdrantClient(url="http://localhost:6333")`. It prints `[0, 1]`.

```python

client.create_collection(
    "repro",
)
client.upsert("repro", points=[
    models.PointStruct(id=0, vector=[1.0, 0.0], payload={"group": "a"}),
    models.PointStruct(id=1, vector=[0.0, 1.0], payload={"group": "a"}),
    # id 2 is the nearest point to both prefetch queries, and the filter must exclude it
    models.PointStruct(id=2, vector=[0.9, 0.1], payload={"group": "b"}),
])

res = client.query_points(
    "repro",
    prefetch=[
        models.Prefetch(query=[1.0, 0.05], limit=10),
        models.Prefetch(query=[0.9, 0.20], limit=10),
    ],
    query=models.FusionQuery(fusion=models.Fusion.RRF),
    limit=10,
    query_filter=models.Filter(
        must=[models.FieldCondition(key="group", match=models.MatchValue(value="a"))]
    ),
)
print(sorted(p.id for p in res.points))
```

Affected root queries: `FusionQuery` (both RRF and DBSF), `RrfQuery`, and `FormulaQuery`. Same result through `query_points_groups` and `query_batch_points`. Queries without prefetches filter correctly in local mode.

## Expected Behavior

`[0, 1]` in local mode, matching the server.

## Possible Solution

Possibly related to #1136, where `score_threshold` was also dropped for fusion queries in local mode.

## Context (Environment)

- qdrant-client 1.19.0, Python 3.14.6
- Server 1.18.2, also checked against 1.12.0
- Reproduced on client 1.10.0, 1.11.0, 1.12.0, 1.13.0, 1.14.0, 1.15.0, 1.16.0, 1.17.0, 1.18.0, and 1.19.0

Prototyping in local mode and deploying against a server gives different results for the same query, with nothing to signal the change.

aiedwardyi:
Fixes #1372 by folding the root filter into the prefetch search, so prefetch limits apply to already-filtered candidates and fusion sees the same candidate set the server does.

Review fixes applied
- 69f756c: filter sources before fusion, so excluded candidates no longer take RRF rank slots or feed the DBSF distribution.
- 2a64e5b: push the root filter into `_prefetch` rather than filtering prefetch results, fixing the case where a prefetch limit drops a point that satisfies both filters.

Verified against qdrant/qdrant:v1.19.0 in Docker, local vs remote on the same queries: all flat prefetch cases match (RRF, DBSF, formula, nearest-rescore). Nested prefetch scoring still differs, but it differs on master with no filters involved - see the comment thread.


linhongyu510 (commented):
Blocking score-parity gap on current head `4bbdd0c9`: the filter is applied after RRF/DBSF, so excluded candidates still affect ranks and distribution normalization. The added test checks only `sorted(point.id)`, which passes while the scores remain observably different from the server.

I ran the PR head and Qdrant Server v1.19.0 against the same four-point collection, two prefetches, and root `group == a` filter. Both return ids `[0, 1]`, but:

```text
                 local PR                    server 1.19.0
RRF id 0        0.8333333333                1.0
RRF id 1        0.5                         0.6666667
DBSF id 0       1.2501247830                1.2357023
DBSF id 1       0.9474276887                0.7642977
RrfQuery        same mismatch as RRF
Formula         matches (point-local score)
```

The excluded high-ranking candidates consume RRF rank positions and contribute to DBSF mean/std before `_filter_scored_points(fused, query_filter)` removes them. This changes downstream score thresholds and ranking composition even though this tiny repro retains the same two ids.

Filtering each prefetch source before calling `reciprocal_rank_fusion`, `distribution_based_score_fusion`, or formula rescoring fixes the contract:

```python
if query_filter is not None:
    sources = [self._filter_scored_points(source, query_filter) for source in sources]
```

With only that change applied to this head, local results become `RRF [(0, 1.0), (1, 0.6666666667)]`, `DBSF [(0, 1.2357022604), (1, 0.7642977396)]`, and Formula unchanged, matching the server to serialization precision. After pre-filtering, `merge_limit`/`rescore_limit` can return to `limit + offset` and the post-fusion filter is redundant.

Please extend the regression beyond ids: assert order and approximate scores for RRF, DBSF, and `RrfQuery`, with an excluded candidate ranked above an included one in at least one source. That locks the actual server contract rather than only result membership. The current GitHub rollup is Netlify-only, so it does not provide Python test evidence for this path.

aiedwardyi:
You're right, thanks - reproduced the rank/distribution leak locally before changing anything. Fixed in 69f756c: sources are filtered once at the top of `_merge_sources`, `merge_limit`/`rescore_limit` are back to `limit + offset`, and the post-fusion filter is gone. Local now returns RRF `[(0, 1.0), (1, 0.6666666667)]` and DBSF `[(0, 1.2357022604), (1, 0.7642977396)]`, matching your server run. The regression asserts order and approximate scores for RRF, DBSF, RrfQuery and Formula, and a companion test pins that the excluded point outranks an included one in both prefetch sources - it fails on the previous head for three of the four queries.

joein:
Hey @aiedwardyi @linhongyu510 

thanks for your interest!

there is a problem with this PR though:


```python


for cl_type, cl in ("local", QdrantClient(":memory:")), ("remote", QdrantClient()):
    print(cl_type)

    cl.create_collection(
        "test", vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT)
    )
    points = [
        models.PointStruct(id=1, vector=[0.3, 0.3], payload={"a": 1, "b": 1}),
        models.PointStruct(id=5, vector=[0.2, 0.3], payload={"a": 1, "b": 2}),
    ]

    cl.upsert("test", points)
    print(cl.query_points(
        "test",
        query=models.RrfQuery(rrf=models.Rrf(k=2)),
        prefetch=models.Prefetch(
            query=[0.1, 0.3],
            filter=models.Filter(
                must=[models.FieldCondition(key="a", match=models.MatchValue(value=1))]
            ),
            limit=1,
        ),
        query_filter=models.Filter(must=[models.FieldCondition(key="b", match=models.MatchValue(value=2))]),
        limit=1
    ))
```

Output:

```python
local
points=[]
remote
points=[ScoredPoint(id=5, version=1, score=0.5, payload={'a': 1, 'b': 2}, vector=None, shard_key=None, order_value=None)]
```

As you can notice, the point which satisfies both conditions gets filtered out by the prefetch leading to the empty result.

aiedwardyi:
Good catch, thanks - your repro was right and the approach was wrong. Both my attempts filtered the prefetch *results*, so the prefetch limit picked id 1 and then dropped it; the server folds the root filter into the prefetch search and never considers id 1 at all.

Fixed in 2a64e5b by pushing the root filter down into `_prefetch` (ANDed with the prefetch's own filter) instead of filtering afterwards. Your repro now returns `[(5, 0.5)]` locally, and @linhongyu510's RRF/DBSF score parity still holds.

I ran a differential harness against qdrant/qdrant:v1.19.0 in Docker, same queries local vs remote, 8 cases: upstream master 1/8, first commit 1/8, second 4/8, this one 5/8. All flat prefetch cases now match (fusion, DBSF, formula, nearest-rescore).

The 3 that still differ are all **nested** prefetches, and they differ with no filter anywhere in the query:

```python
prefetch=models.Prefetch(prefetch=models.Prefetch(query=[0.1, 0.3], limit=3),
                         query=[0.9, 0.2], limit=3)
query=models.FusionQuery(fusion=models.Fusion.RRF)   # no query_filter, no prefetch filter
```
local `[(7, 0.75), (9, 0.75), (1, 0.666667)]` vs server `[(7, 0.5), (1, 0.333333), (9, 0.25)]`

That reproduces on master, so it looks like a separate pre-existing bug in nested prefetch scoring rather than anything to do with filters. Happy to open a separate issue for it, or fold it in here if you'd prefer - let me know which you want.

joein:
@aiedwardyi what 8 cases are we talking about?

aiedwardyi:
Sorry - I quoted counts without showing the cases. Here they are.

First a correction: one of my 8 was a bad test. It scored two points at exactly 0.5, so the tie order flipped between runs and the count wasn't stable. With a non-degenerate vector the counts are **1/8 master, 2/8 `4bbdd0c`, 4/8 `69f756c`, 5/8 `2a64e5b`** - I previously said the first commit was 1/8, it's 2/8.

One collection, DOT, 2 dims, 4 points:

| id | vector | payload |
|---|---|---|
| 1 | [0.3, 0.3] | a=1, b=1, g=x |
| 5 | [0.2, 0.3] | a=1, b=2, g=x |
| 7 | [0.9, 0.1] | a=1, b=2, g=y |
| 9 | [0.1, 0.9] | a=2, b=2, g=y |

On `2a64e5b`, local vs qdrant/qdrant:v1.19.0:

```
match   1  flat  RrfQuery, prefetch filter a=1 limit 1, root b=2   (your repro)
match   2  flat  RRF, 2 prefetches no filter, root g=x
match   3  flat  DBSF, 2 prefetches no filter, root g=x
DIFFER  4  NESTED  RRF over prefetch(prefetch), root b=2
          local  [(7, 0.833333), (9, 0.833333)]
          server [(7, 0.5), (9, 0.333333)]
DIFFER  5  NESTED  same + intermediate filter a=1, root b=2
          local  [(7, 0.833333), (5, 0.583333), (9, 0.5)]
          server [(7, 0.5), (5, 0.333333)]
DIFFER  6  NESTED  inner filter a=1 limit 1, root b=2
          local  [(7, 1.0)]
          server [(7, 0.5)]
match   7  flat  FormulaQuery, root b=2
match   8  flat  NearestQuery rescore, root b=2

5/8 match
```

Cases 4-6 are the nested ones. They also differ on master with no filter anywhere in the query, which is why I think they're separate from this PR.

<details>
<summary>harness script</summary>

```python

VP = models.VectorParams(size=2, distance=models.Distance.DOT)
PTS = [
    models.PointStruct(id=1, vector=[0.3, 0.3], payload={"a": 1, "b": 1, "g": "x"}),
    models.PointStruct(id=5, vector=[0.2, 0.3], payload={"a": 1, "b": 2, "g": "x"}),
    models.PointStruct(id=7, vector=[0.9, 0.1], payload={"a": 1, "b": 2, "g": "y"}),
    models.PointStruct(id=9, vector=[0.1, 0.9], payload={"a": 2, "b": 2, "g": "y"}),
]

def f(**kw):
    return models.Filter(must=[models.FieldCondition(key=k, match=models.MatchValue(value=v))
                               for k, v in kw.items()])

RRF = models.FusionQuery(fusion=models.Fusion.RRF)
CASES = [
 ("1  flat  RrfQuery, prefetch filter a=1 limit 1, root b=2   (your repro)",
  dict(query=models.RrfQuery(rrf=models.Rrf(k=2)),
       prefetch=models.Prefetch(query=[0.1, 0.3], filter=f(a=1), limit=1),
       query_filter=f(b=2), limit=5)),
 ("2  flat  RRF, 2 prefetches no filter, root g=x",
  dict(query=RRF,
       prefetch=[models.Prefetch(query=[0.1, 0.3], limit=10),
                 models.Prefetch(query=[0.9, 0.2], limit=10)],
       query_filter=f(g="x"), limit=10)),
 ("3  flat  DBSF, 2 prefetches no filter, root g=x",
  dict(query=models.FusionQuery(fusion=models.Fusion.DBSF),
       prefetch=[models.Prefetch(query=[0.1, 0.3], limit=10),
                 models.Prefetch(query=[0.9, 0.2], limit=10)],
       query_filter=f(g="x"), limit=10)),
 ("4  NESTED  RRF over prefetch(prefetch), root b=2",
  dict(query=RRF,
       prefetch=models.Prefetch(prefetch=models.Prefetch(query=[0.1, 0.3], limit=2),
                                query=[0.9, 0.2], limit=2),
       query_filter=f(b=2), limit=5)),
 ("5  NESTED  same + intermediate filter a=1, root b=2",
  dict(query=RRF,
       prefetch=models.Prefetch(prefetch=models.Prefetch(query=[0.1, 0.3], limit=3),
                                query=[0.9, 0.2], filter=f(a=1), limit=3),
       query_filter=f(b=2), limit=5)),
 ("6  NESTED  inner filter a=1 limit 1, root b=2",
  dict(query=models.RrfQuery(rrf=models.Rrf(k=2)),
       prefetch=models.Prefetch(prefetch=models.Prefetch(query=[0.1, 0.3], filter=f(a=1), limit=1),
                                query=[0.9, 0.2], limit=3),
       query_filter=f(b=2), limit=5)),
 ("7  flat  FormulaQuery, root b=2",
  dict(query=models.FormulaQuery(formula=models.MultExpression(mult=["$score", 1.0])),
       prefetch=[models.Prefetch(query=[0.1, 0.3], limit=10)],
       query_filter=f(b=2), limit=10)),
 ("8  flat  NearestQuery rescore, root b=2",
  dict(query=models.NearestQuery(nearest=[0.6, 0.4]),
       prefetch=[models.Prefetch(query=[0.1, 0.3], limit=10)],
       query_filter=f(b=2), limit=10)),
]

def run(cl, kw):
    cl.delete_collection("t"); cl.create_collection("t", vectors_config=VP); cl.upsert("t", PTS)
    try:
        return [(p.id, round(p.score, 6)) for p in cl.query_points("t", **kw).points]
    except Exception as e:
        return f"ERROR {type(e).__name__}"

loc = QdrantClient(":memory:")
rem = QdrantClient(url="http://localhost:6333")
ok = 0
for name, kw in CASES:
    l, r = run(loc, kw), run(rem, kw)
    match = l == r
    ok += match
    print(f"{'match ' if match else 'DIFFER'}  {name}")
    if not match:
        print(f"          local  {l}")
        print(f"          server {r}")
print(f"\n{ok}/{len(CASES)} match")
```
</details>

joein:
@aiedwardyi thanks for the contribution
I adjusted the pr a bit
Also regarding the failing tests: https://github.com/qdrant/qdrant-client/pull/1397

aiedwardyi:
Thanks! The field-wise merge makes sense. Nesting the two filters like I did would've quietly ANDed the `should` clauses. Glad #1397 covers the nested case.

PR #1375, opened 2026-08-25
Fix datetime parse truncated bug

### LINKED ISSUE #1349: datetime_utils.parse accepts truncated datetimes: the hour-only-offset retry is unguarded
### Describe the bug

`qdrant_client.local.datetime_utils.parse()` accepts datetime strings that qdrant core rejects, so local mode silently succeeds on values that would fail server-side.

```python
>>> from qdrant_client.local.datetime_utils import parse
>>> parse("2024-06-15 12")
datetime.datetime(2024, 6, 15, 12, 0, tzinfo=datetime.timezone.utc)
>>> parse("2024-06-15T12:30")
datetime.datetime(2024, 6, 15, 12, 30, tzinfo=datetime.timezone.utc)
```

Neither is an accepted format. `available_formats` has no `%Y-%m-%d %H` and no `%Y-%m-%dT%H:%M`, and the [core list the docstring points at](https://github.com/qdrant/qdrant/blob/0ed86ce0575d35930268db19e1f7680287072c58/lib/segment/src/types.rs#L1388-L1410) is RFC 3339 plus `%Y-%m-%dT%H:%M:%S%.f`, `%Y-%m-%d %H:%M:%S%.f`, `%Y-%m-%d %H:%M`, `%Y-%m-%d`.

### Cause

The final fallback appends `:00` to **any** string that failed the format list:

```python
# Python can't parse timezones containing only hours (+HH), but it can parse timezones with hours and minutes
# So we add :00 to the assumed timezone and try parsing it again
return parse_available_formats(date_str + ":00")
```

It exists to complete an hour-only UTC offset (`...000+01` -> `...000+01:00`), but it is unguarded, so it also completes truncated datetimes:

- `"2024-06-15 12"` + `":00"` -> `"2024-06-15 12:00"`, which matches `%Y-%m-%d %H:%M`
- `"2024-06-15T12:30"` + `":00"` -> `"2024-06-15T12:30:00"`, which matches `%Y-%m-%dT%H:%M:%S`

Confirmed against `dev` (`f003e6c`) — neither string matches any entry in `available_formats` directly:

```python
>>> from qdrant_client.local import datetime_utils as d
>>> from datetime import datetime
>>> [f for f in d.available_formats if _parses("2024-06-15 12", f)]
[]
>>> [f for f in d.available_formats if _parses("2024-06-15 12:00", f)]
['%Y-%m-%d %H:%M']
```

### Expected behavior

The `:00` retry should only apply to a string that actually ends in an hour-only offset. Anything else that fails the format list should return `None`, so local mode and the server agree on what is a valid datetime.

### Impact

A payload or filter value with a truncated datetime is accepted by `QdrantLocal` and rejected by a real Qdrant instance, and `"2024-06-15 12"` is read as `12:00:00` rather than surfacing as invalid input. It is the local/server congruence property this client tests for elsewhere.

### Environment

- qdrant-client @ `dev` (`f003e6c`)
- Python 3.14

TarunSinghChauhan:
Fixes #1349

### DIFF
--- a/qdrant_client/local/datetime_utils.py
+++ b/qdrant_client/local/datetime_utils.py
@@ -1,3 +1,4 @@
+import re
 from datetime import datetime, timezone
 
 # These are the formats accepted by qdrant core
@@ -46,4 +47,12 @@ def parse(date_str: str) -> datetime | None:
     # dt examples to handle:
     # "2021-01-01 00:00:00.000+01"
     # "2021-01-01 00:00:00.000-10"
-    return parse_available_formats(date_str + ":00")
+    # This retry must only fire for an actual hour-only UTC offset at the end of
+    # the string (e.g. "+01" or "-10"). Without this guard, any string that fails
+    # every format above (including truncated datetimes like "2024-06-15 12" or
+    # "2024-06-15T12:30") would also get ":00" appended and could accidentally
+    # match a valid format, silently accepting input that qdrant core rejects.
+    if re.search(r"[+-]\d{2}$", date_str):
+        return parse_available_formats(date_str + ":00")
+
+    return None
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -697,7 +697,12 @@ class LocalCollection:
 
         required_order = distance_to_order(distance)
 
-        if required_order == DistanceOrder.BIGGER_IS_BETTER or isinstance(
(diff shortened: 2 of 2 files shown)


TarunSinghChauhan:
Closing in favor of a cleaner PR — this branch accidentally included commits from #1374. Opening a fresh PR isolated to just the datetime fix.

PR #1377, opened 2026-08-25
Fix TypeError when sorting heterogeneous facet values / mixed-type point ids in local mode

### LINKED ISSUE #1376: Local mode: facet and search_matrix crash with TypeError sorting heterogeneous values (mixed payload types / mixed int+UUID ids)
**Current Behavior**

Two local-mode (`:memory:` / on-disk) operations crash with `TypeError: '<' not supported` on data Qdrant otherwise accepts, because they sort union-typed values with a raw key.

```python
c = QdrantClient(":memory:")

# (1) facet — a payload key holding mixed scalar types, on a count tie
c.create_collection("f", vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT))
c.upsert("f", points=[
    models.PointStruct(id=1, vector=[0.1, 0.2], payload={"k": 1}),
    models.PointStruct(id=2, vector=[0.2, 0.3], payload={"k": "x"}),
])
c.facet("f", key="k")   # TypeError: '<' not supported between instances of 'str' and 'int'

# (2) search_matrix — a collection mixing integer and UUID point ids
c.create_collection("m", vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT))
c.upsert("m", points=[
    models.PointStruct(id=1, vector=[0.1, 0.9]),
    models.PointStruct(id="550e8400-e29b-41d4-a716-446655440000", vector=[0.9, 0.1]),
])
c.search_matrix_pairs("m", sample=10, limit=1)   # TypeError: '<' not supported between instances of 'int' and 'str'
```

`facet` (`local_collection.py`) breaks a count tie with the raw facet value, and `_search_distance_matrix` sorts samples by raw point id — both raise when the values span types.

**Expected Behavior**

Both return normally. Mixed payload value types and mixed integer/UUID point ids are both valid in Qdrant, so the local backend should order them with a type-safe key rather than crash.

I have a fix ready (route both through the existing `_universal_id` helper, which `_scroll_by_id` already uses) plus regression tests.

winklemad:
Fixes #1376

## Problem

Two sorts in `qdrant_client/local/local_collection.py` use a key that Python cannot order across types, so they crash on inputs Qdrant otherwise accepts:

1. **`LocalCollection.facet`** breaks a count tie with the raw facet value (`key=lambda x: (-x[1], x[0])`). A payload key may hold mixed scalar types across points (e.g. `{"k": 1}` and `{"k": "x"}`), so on a tie the comparison raises `TypeError: '<' not supported between instances of 'str' and 'int'`.
2. **`_search_distance_matrix`** (reached from `search_matrix_pairs` / `search_matrix_offsets`) sorts samples by raw point id (`key=lambda x: x.id`). Qdrant allows integer and UUID (string) point ids to coexist in one collection, so this raises `TypeError: '<' not supported between instances of 'int' and 'str'`.

Both are crashes on valid data.

### Reproduce (local mode, no server)

```python
c = QdrantClient(":memory:")
c.create_collection("f", vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT))
c.upsert("f", points=[
    models.PointStruct(id=1, vector=[0.1, 0.2], payload={"k": 1}),
    models.PointStruct(id=2, vector=[0.2, 0.3], payload={"k": "x"}),
])
c.facet("f", key="k")   # TypeError: '<' not supported between instances of 'str' and 'int'
```

## Fix

Route both sorts through the collection's existing `_universal_id` key, which already groups a point id by type (`str` -> `(id, 0)`, `int` -> `("", id)`) before ordering within the type — the same helper `_scroll_by_id` already uses. Homogeneous inputs keep their previous ordering; only the previously-crashing mixed-type inputs gain a defined total order. Result semantics are unchanged: facet still orders by descending count first, and the matrix id order stays deterministic.

Swept the other `sort`/`sorted` sites in the file: `_scroll_by_id` already uses `_universal_id`; `_scroll_by_value` sorts numeric-only `OrderValue` (from `to_order_value`, always int/float); the rescore sort is by float score. Only the two sites above have the union-typed hazard.

## Tests

Added two regression tests to `tests/test_in_memory.py` (pure `:memory:` local mode) mirroring the repros:
- facet over a payload key with mixed `int` / `str` / `bool` values on a count tie — returns without error and in a deterministic, type-grouped order.
- `search_matrix_pairs` and `search_matrix_offsets` over a collection mixing integer and UUID ids — return without error.

Both fail on unmodified `dev` (`TypeError`) and pass after the change. `ruff-format`, `mypy`, and `pyright` (strict) are clean on the changed module.


winklemad:
Good catch — `_universal_id` is only injective for point ids; for facet values it maps `""`→`("", 0)` and `0`→`("", 0)` (and `False`/`0`, `True`/`1`) to the same key, so equal-count ties could reorder with the upsert order.

I've switched the facet tie-break to a dedicated `_facet_value_sort_key` that discriminates by type first (bool < int < str) so distinct values never collide, and kept `_universal_id` for the `search_matrix` point-id sort (ids are int/UUID, where it's already injective). Added a regression test that upserts `""` and `0` in both orders and asserts identical ordering.

winklemad:
@coderabbitai thanks — two things:

- The `local_collection.py:1291` note is now stale: this PR already routes the facet tie-break through the new `_facet_value_sort_key` (type-discriminated: bool < int < str), so `""` and `0` no longer collide there.
- The `False`/`0` (and `True`/`1`) merge you flag at the counting layer is real, but it's a **separate, pre-existing** issue: `facet_hits`/`values_set` are plain `dict`/`set`, and `False == 0` with the same hash in Python, so those values merge into one hit *before* any sorting. That's independent of this PR (which fixes the `TypeError` and tie determinism); I intentionally scoped the regression test to `""` vs `0` (genuinely distinct keys) to avoid conflating the two. Happy to open a follow-up to key the facet counters by `(type, value)` if the maintainers want bool/int treated as distinct facet values.

linhongyu510 (commented):
Independent verification on current head `954b8f08`: no additional blocking finding within this PRs stated crash/determinism scope. The complete `tests/test_in_memory.py` file is **7 passed in 0.78s**, including all three new heterogeneous-value/id regressions.

I also compared the mixed-ID matrix path against a real Qdrant Server v1.19.0 with the same three ids inserted out of order. Local and server `search_matrix_offsets(sample=10, limit=1).ids` both return `[1, 7, "550e8400-e29b-41d4-a716-446655440000"]`, supporting `_universal_id` for the point-id ordering contract.

For facets, the real server requires a typed payload index; sequential bool/integer/keyword index creation leaves the active schema type-specific, so I would limit the claim to local-mode no-crash and deterministic ordering rather than claiming a cross-type server facet order. The already-open CodeRabbit thread correctly tracks the separate pre-existing `False == 0` / `True == 1` aggregation collapse before sorting; I did not duplicate that finding. On this head, local mode returns `False:2`, `True:2`, and `"0":1` for five distinct JSON values, confirming that follow-up remains real but is not introduced by these two sort-key changes.

Gate attribution: the current GitHub rollup contains only Netlify deploy contexts, not Python tests, typing, or lint. Thus the evidence is the independent seven-test run plus the real-server matrix comparison, not a claim that code CI ran remotely.

ErenAta16 (commented):
I ran this branch locally against master while reviewing #1389, and both halves
of it hold up. The `_search_distance_matrix` change is the part I want to point
at, because it is the one nothing else on the tracker covers.

On master, a collection mixing integer and UUID point ids raises out of
`sorted(samples, key=lambda x: x.id)`:

```
search_matrix_pairs, ids [1, "550e8400-...", 3]
  master  -> TypeError: '<' not supported between instances of 'str' and 'int'
  #1377   -> pairs=6
```

Reusing `_universal_id` there is the right call rather than inventing a second
ordering, since point ids already have a canonical total order in this file.

The reason I am leaving this note is that the facet half of your change overlaps
#1389, and I would not want that to read as this branch being redundant. #1389
moves the type rank into the dict and set keys instead of the sort key, which
fixes the bool/int merge in #1388 and makes the tie-break total as a consequence,
so it happens to cover your facet crash as well. It does not touch
`_search_distance_matrix` at all, so that fix is unique to this PR. Both are
worth having, and whichever lands second will need a small rebase, since the two
touch the same `sorted(facet_hits.items(), ...)` line and both add a static
method just after `_universal_id`.

Worth knowing for that rebase: your `_facet_value_sort_key` and their
`_facet_value_identity` produce the same relative order, so `["a", 1]` comes out
as `[(int, 1), (str, "a")]` either way. The merge is a textual conflict rather
than a behavioural one.

winklemad:
Thanks for taking the time to run it locally @ErenAta16, appreciate the independent check.

ErenAta16:
The reuse of `_universal_id` is what made it quick to check, so some of that is
your doing. A change that borrows an ordering the file already defines is one I can
verify by running two branches, rather than one I have to reason about.

If the facet halves do end up conflicting when one of them lands, ping me and I
will re-run the same matrix against the merged result. It is scripted now, so it
costs nothing to repeat.

PR #1379, opened 2026-08-25
Fix score-orientation handling for recommend/discovery/context/feedback queries in local mode

### LINKED ISSUE #1378: Local mode: recommend/discovery/context/feedback queries mis-order and mis-threshold on Euclid/Manhattan
**Current Behavior**

Recommend / discovery / context / relevance-feedback queries score points from the internal *core* distance, which is oriented so a higher score is always better regardless of the collection metric. On `Euclid`/`Manhattan` collections in local mode, two paths in `LocalCollection.search` treat those scores using the raw distance order instead, producing wrong results.

**1. Relevance-feedback returns the farthest points first (reversed order).**

```python
c = QdrantClient(":memory:")
c.create_collection("rf", vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID))
c.upsert("rf", points=[
    models.PointStruct(id=1, vector=[0.1, 0.0]),   # nearest to target [0, 0]
    models.PointStruct(id=5, vector=[5.0, 0.0]),
    models.PointStruct(id=9, vector=[9.0, 0.0]),   # farthest
])
c.query_points("rf", query=models.RelevanceFeedbackQuery(relevance_feedback=models.RelevanceFeedbackInput(
    target=[0.0, 0.0],
    feedback=[models.FeedbackItem(example=[3.0, 0.0], score=0.5),
              models.FeedbackItem(example=[4.0, 0.0], score=0.5)],
    strategy=models.NaiveFeedbackStrategy(naive=models.NaiveFeedbackStrategyParams(a=1.0, b=1.0, c=1.0)),
)), limit=3).points
# ids -> [9, 5, 1]   (expected [1, 5, 9])
```

**2. `score_threshold` drops all results for the whole higher-is-better family (recommend/discovery/context/feedback).**

```python
c.create_collection("t", vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID))
c.upsert("t", points=[models.PointStruct(id=i, vector=[float(i), 0.0]) for i in (1, 2, 3, 4)])
c.query_points("t",
    query=models.RecommendQuery(recommend=models.RecommendInput(
        positive=[[1.0, 0.0]], negative=[], strategy=models.RecommendStrategy.BEST_SCORE)),
    limit=10, score_threshold=0.13).points
# returns []   (expected the points scoring >= 0.13)
```

**Expected Behavior**

Feedback results ordered nearest-first (`[1, 5, 9]`), and `score_threshold` keeps the points scoring at/above the threshold for the recommend/discovery/context/feedback family.

**Root cause**

Both the result ordering and the `score_threshold` check derive their direction from `distance_to_order(distance)` (the raw distance order). For the recommend/discovery/context/feedback family the scores are higher-is-better on every metric, so on Euclid/Manhattan the ordering special-case must apply (it omitted `NaiveFeedbackQuery`) and the threshold direction must match — instead the threshold followed the raw order and inverted.

I have a fix that derives a single higher-is-better flag from the query type and uses it for both the ordering and the threshold, plus regression tests over both metrics.

winklemad:
Fixes #1378

## Problem

Recommend, discovery, context and relevance-feedback queries score points from the internal **core** distance (`calculate_distance_core`), which negates the Euclid/Manhattan distance so that a **higher score is always better**, independent of the collection's distance metric. The raw distance order (`distance_to_order`) therefore does not describe how those scores should be sorted or thresholded. Two places in `LocalCollection.search` got this wrong on `Euclid`/`Manhattan` collections in local mode:

**1. Ordering — relevance-feedback returned the farthest points first.**
The ordering already special-cased the recommend/discovery/context family to sort descending, but `NaiveFeedbackQuery` was omitted from that tuple, so it fell to the smaller-is-better branch:

```python
c.create_collection("rf", vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID))
c.upsert("rf", points=[
    models.PointStruct(id=1, vector=[0.1, 0.0]),   # nearest to target [0, 0]
    models.PointStruct(id=5, vector=[5.0, 0.0]),
    models.PointStruct(id=9, vector=[9.0, 0.0]),   # farthest
])
c.query_points("rf", query=models.RelevanceFeedbackQuery(relevance_feedback=models.RelevanceFeedbackInput(
    target=[0.0, 0.0],
    feedback=[models.FeedbackItem(example=[3.0, 0.0], score=0.5),
              models.FeedbackItem(example=[4.0, 0.0], score=0.5)],
    strategy=models.NaiveFeedbackStrategy(naive=models.NaiveFeedbackStrategyParams(a=1.0, b=1.0, c=1.0)),
)), limit=3).points
# ids -> [9, 5, 1]   (expected [1, 5, 9])
```

**2. `score_threshold` — the whole higher-is-better family was filtered in the wrong direction.**
The threshold check keyed off `required_order` for every query type, so on Euclid/Manhattan it took the smaller-is-better branch and dropped all results (or applied no filtering) instead of removing the low-scoring points. This affects recommend/discovery/context too, not just feedback:

```python
c.create_collection("t", vectors_config=models.VectorParams(size=2, distance=models.Distance.EUCLID))
c.upsert("t", points=[models.PointStruct(id=i, vector=[float(i), 0.0]) for i in (1, 2, 3, 4)])
c.query_points("t",
    query=models.RecommendQuery(recommend=models.RecommendInput(
        positive=[[1.0, 0.0]], negative=[], strategy=models.RecommendStrategy.BEST_SCORE)),
    limit=10, score_threshold=0.13).points
# returns []   (expected the points scoring >= 0.13)
```

## Fix

Derive a single `higher_score_is_better` flag from the query type once, and use it for **both** the ordering and the `score_threshold` comparison. Add `NaiveFeedbackQuery` to the family. `Cosine`/`Dot` collections are already higher-is-better and are unchanged; plain nearest queries (which score by displayed distance) keep the smaller-is-better path.

I traced every branch that assigns `scores` in this method: the ones built on `calculate_distance_core` (recommend/discovery/context, dense and multi, plus feedback) are exactly the higher-is-better set; plain dense/sparse nearest use the displayed distance and stay smaller-is-better.

## Tests

Added parametrized regression tests to `tests/test_in_memory.py` (pure `:memory:` local mode), each covering both `Euclid` and `Manhattan`:
- relevance-feedback ordering is nearest-first (`[1, 5, 9]`), with non-increasing scores;
- `score_threshold` on a recommend query keeps exactly the points scoring at/above the threshold.

Both fail on `dev` and pass with the fix. Full `tests/test_in_memory.py`: 8 passed; `ruff format`/`ruff check`/`mypy` clean.

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -712,6 +712,7 @@ class LocalCollection:
                 MultiDiscoveryQuery,
                 MultiContextQuery,
                 MultiRecoQuery,
+                NaiveFeedbackQuery,
             ),  # sparse structures are not required, sparse always uses DOT
         ):
             order = np.argsort(scores)[::-1]
--- a/tests/congruence_tests/test_query.py
+++ b/tests/congruence_tests/test_query.py
@@ -1108,6 +1108,31 @@ class TestSimpleSearcher:
             limit=10,
         )
 
+    def relevance_feedback_query_euclid(self, client: QdrantBase) -> models.QueryResponse:
+        # `code` is a euclid vector. Feedback scores come from the internal core
+        # distance, which is oriented higher-is-better whatever the metric, so the
+        # results still have to come back best-first instead of farthest-first.
+        return client.query_points(
+            collection_name=COLLECTION_NAME,
+            query=models.RelevanceFeedbackQuery(
+                relevance_feedback=models.RelevanceFeedbackInput(
+                    target=self.dense_vector_query_code,
+                    feedback=[
+                        models.FeedbackItem(example=1, score=0.9),
+                        models.FeedbackItem(example=2, score=0.8),
+                        models.FeedbackItem(example=3, score=0.7),
+                        models.FeedbackItem(example=4, score=0.6),
+                    ],
+                    strategy=models.NaiveFeedbackStrategy(
+                        naive=models.NaiveFeedbackStrategyParams(a=0.5, b=1.0, c=0.7)
+                    ),
+                )
+            ),
+            using="code",
+            with_payload=True,
+            limit=10,
+        )
+
 
 def group_by_keys():
     return ["maybe", "rand_digit", "two_words", "city.name", "maybe_null", "id"]
@@ -2194,4 +2219,7 @@ def test_relevance_feedback_queries():
 
     compare_clients_results(
         local_client, http_client, grpc_client, searcher.relevance_feedback_query
+    )
+    compare_clients_results(
+        local_client, http_client, grpc_client, searcher.relevance_feedback_query_euclid
     )


joein:
Hey @winklemad 

Thanks for the contribution!

Unfortunately, we can't change higher_is_better right now because the client will diverge from the server.

I think there might be a bug in the server and I reached out to someone on the team.

What I am going to do now is to remove the threshold part of the PR and keep the NaiveFeedbackQuery check which is still useful and a nice catch!

I am also going to remove the threshold test, and move the relevance feedback test to congruence tests, since that's the place where we compare the server implementation with the local one.

joein (approved):

PR #1385, opened 2026-08-28
fix(local): skip missing multivectors during search

### LINKED ISSUE #1383: Local mode: multivector search raises IndexError when a point has no vector under that name
## Current behavior

Searching a named multivector crashes if any point in the collection has no vector under that name. A dense named vector in the same collection, on the same point, is handled correctly.

```python

c = QdrantClient(":memory:")
c.create_collection("c", vectors_config={
    "dense": models.VectorParams(size=4, distance=models.Distance.COSINE),
    "multi": models.VectorParams(size=4, distance=models.Distance.COSINE,
        multivector_config=models.MultiVectorConfig(comparator=models.MultiVectorComparator.MAX_SIM))})
c.upsert("c", points=[
    models.PointStruct(id=1, vector={}, payload={"k": 1}),                        # neither vector
    models.PointStruct(id=2, vector={"dense": [1.0] * 4, "multi": [[1.0] * 4]})])

c.query_points("c", query=[1.0] * 4, using="dense", limit=5)      # [(2, 1.0)], point 1 skipped
c.query_points("c", query=[[1.0] * 4], using="multi", limit=5)    # IndexError: invalid index to scalar variable.
```

Reproduces on 1.18.0 from PyPI and on `dev`. A Qdrant 1.18.1 server returns `[(2, 1.0)]` for the multivector query as well.

## Cause

Missing vectors are not filtered out before scoring. The mask is applied to the results afterwards (`mask = mask & ~self.deleted_per_vector[vector_name]`), so a placeholder always reaches the distance function. The two placeholders are not shaped alike:

- **Dense** uses `np.ones(size, dtype=np.float32)`, one well-formed row inside the 2-D matrix that `cosine_similarity` receives. It scores harmlessly and the mask drops it afterwards.
- **Multivectors** use `np.array([])`, which is 1-D. `calculate_multi_distance_core` iterates the candidates and hands each matrix to `cosine_similarity` on its own, so that empty 1-D array arrives alone at:

```python
vectors_norm = np.linalg.norm(vectors, axis=-1)[:, np.newaxis]
```

There `np.linalg.norm(..., axis=-1)` returns a scalar, and the subscript raises.

## Expected behavior

A multivector search skips points that do not have that vector, the way the dense path and the server already do.

## Possible fix

Shaping the placeholder like the dense one looks sufficient. `np.empty((0, size), dtype=np.float32)` is 2-D, so the norm returns an empty column and the point scores as empty before the mask removes it. Checked in isolation:

```
np.array([])       -> IndexError
np.empty((0, 4))   -> ok, shape (0, 4)
```

Happy to open a PR if that direction looks right.

mikemikimike:
Fixes #1383

## Summary
- Treat empty multivector matrices as non-matching points during local distance calculation.
- Preserve normal scoring while allowing the existing deleted-vector mask to skip missing vectors.
- Add a local regression test for searching a named multivector when another point has no value for it.

## Testing
- `pytest -q qdrant_client/local/tests/test_distances.py qdrant_client/local/tests/test_vectors.py`
- `pytest -q qdrant_client/local/tests`
- `uv tool run --from ruff==0.4.3 ruff check qdrant_client/local/multi_distances.py qdrant_client/local/tests/test_vectors.py`
- `git diff --check`

The full test suite reached 78 passed and 2 skipped before `tests/congruence_tests/test_aliases.py` failed because the local Qdrant endpoint at `localhost:6333` returned HTTP 502.

### DIFF
--- a/qdrant_client/local/multi_distances.py
+++ b/qdrant_client/local/multi_distances.py
@@ -82,6 +82,7 @@ def calculate_multi_distance_core(
     matrices: list[types.NumpyArray],
     distance_type: models.Distance,
 ) -> types.NumpyArray:
+    """Calculate similarities for each candidate multivector matrix."""
     def euclidean(q: types.NumpyArray, m: types.NumpyArray, *_: Any) -> types.NumpyArray:
         return -np.square(m - q, dtype=np.float32).sum(axis=-1, dtype=np.float32)
 
--- a/qdrant_client/local/multi_distances.py
+++ b/qdrant_client/local/multi_distances.py
@@ -100,6 +100,9 @@ def calculate_multi_distance_core(
         dist_func = calculate_distance  # type: ignore
 
     for matrix in matrices:
+        if matrix.size == 0:
+            similarities.append(-np.inf)
+            continue
         sim_matrix = dist_func(query_matrix, matrix, distance_type)
         similarity = float(np.sum(np.max(sim_matrix, axis=-1)))
         similarities.append(similarity)
--- a/qdrant_client/local/tests/test_vectors.py
+++ b/qdrant_client/local/tests/test_vectors.py
@@ -5,6 +5,7 @@ from qdrant_client.local.local_collection import LocalCollection, DEFAULT_VECTOR
 
 
 def test_get_vectors():
+    """Test retrieving vectors from a local collection."""
     collection = LocalCollection(
         models.CreateCollection(
             vectors=models.VectorParams(size=2, distance=models.Distance.MANHATTAN)
@@ -22,6 +23,7 @@ def test_get_vectors():
 
 
 def test_multivector_search_skips_points_without_a_vector():
+    """Test that missing named multivectors are skipped during local search."""
     collection = LocalCollection(
         models.CreateCollection(
             vectors={
--- a/qdrant_client/local/tests/test_vectors.py
+++ b/qdrant_client/local/tests/test_vectors.py
@@ -19,3 +19,30 @@ def test_get_vectors():
     assert collection._get_vectors(idx=1, with_vectors=DEFAULT_VECTOR_NAME)
     assert collection._get_vectors(idx=2, with_vectors=True)
     assert collection._get_vectors(idx=3, with_vectors=False) is None
+
+
+def test_multivector_search_skips_points_without_a_vector():
+    collection = LocalCollection(
+        models.CreateCollection(
+            vectors={
+                "dense": models.VectorParams(size=4, distance=models.Distance.COSINE),
+                "multi": models.VectorParams(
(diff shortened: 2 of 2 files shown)


Gronoxx:
Issue author here. The -inf approach looks right to me. It keeps the deleted-vector mask as the single place that decides skipping, instead of adding a second guard, and it covers my case: a collection mixing points with and without the named multivector.

Caveat: this is from reading the diff. I haven't run your branch against my reproduction yet.

Gronoxx:
Ran my reproduction against this branch, plus a set of cases around it.

The reported case. On qdrant-client==1.18.0 it raises `IndexError: invalid index to scalar variable.`. On this branch there is no error, and the point without the multivector is skipped instead of matched.

Why your approach beats the one I suggested. In the issue I proposed reshaping the placeholder to `np.empty((0, size))`. I implemented that in all four places the 1-D placeholder is created and tested it. The `IndexError` does disappear at `np.linalg.norm`, but the failure moves to `np.max(sim_matrix, axis=-1)` in `calculate_multi_distance_core`, as `ValueError: zero-size array to reduction operation maximum which has no identity`. An empty set has no maximum. Skipping before the reduction avoids both, so my suggestion would not have been enough on its own.

**It fixes more than what I reported.** The guard sits in `calculate_multi_distance_core`, which has ten call sites, so the same `IndexError` is gone from recommend, discover, context search and MMR. All four raise on 1.18.0 and work here.

Regressions I looked for and did not find. Scores for non-empty multivectors are bit-identical to 1.18.0 across three points with 2, 1 and 3 tokens. All four distances behave. `-inf` never reaches the result, not at `limit=1000` and not at `score_threshold=-1e30`, so JSON serialization stays valid. Same through `query_batch_points`, with a `query_filter`, and on a collection persisted to disk and reopened. With four valid points and one missing the vector, all four still come back. Local suite: 73 passed.

**One adjacent case, outside this issue's scope.** An explicitly empty multivector (`vector={"multi": []}`) still raises the same `IndexError` on the write path, in `_add_point`, `_update_point` and `_update_named_vectors`, from the same `np.linalg.norm(vector_np, axis=-1)[:, np.newaxis]` pattern. That is unchanged from 1.18.0 and not something I reported. I traced it only as far as my time allowed, so I am not certain whether an empty multivector is meant to be valid input in the first place. Flagging it in case it deserves a separate issue.

joein:
Hey @mikemikimike 

Unfortunately, this time I have to close the PR, but we appreciate the attempt!

All the PRs have to point `dev` branch due to the way we make releases. It's a bit of a mess to do all the rebases and I also had to fix the approach a little bit, I found it easier to just create another PR #1399

PR #1389, opened 2026-08-30
fix(local): keep facet bool and int values distinct

### LINKED ISSUE #1388: Local facet conflates boolean and integer values before sorting (`False == 0`, `True == 1`)
## Current behavior

Local-mode `facet()` conflates boolean and integer JSON values before sorting:

```python
values_set: set[FacetValue] = set()
...
values_set.add(v)
...
facet_hits[v] += 1
```

Python considers `False == 0` and `True == 1`, with identical hashes. The per-point de-duplication set and global count dictionary therefore merge values that Qdrant's generated `FacetValue` model represents as distinct `StrictBool` and `StrictInt` variants.

This is separate from #1376 / #1377. That PR fixes heterogeneous sort crashes and adds a type-discriminated tie-break, but the values have already been merged before its sort key runs.

## Reproduction

Tested on current `master` at `550484d767d319857d4f46e97d4551ba419ee670`:

```python

client.create_collection(
    "facet_identity",
    vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT),
)
client.upsert(
    "facet_identity",
    points=[
        models.PointStruct(id=1, vector=[0.1, 0.2], payload={"k": False}),
        models.PointStruct(id=2, vector=[0.2, 0.3], payload={"k": 0}),
        models.PointStruct(id=3, vector=[0.3, 0.4], payload={"k": True}),
        models.PointStruct(id=4, vector=[0.4, 0.5], payload={"k": 1}),
        models.PointStruct(id=5, vector=[0.5, 0.6], payload={"k": "0"}),
    ],
)

for hit in client.facet("facet_identity", key="k", exact=True, limit=10).hits:
    print(type(hit.value).__name__, repr(hit.value), hit.count)
```

Actual:

```text
bool False 2
bool True 2
str '0' 1
```

There are five distinct JSON scalar values, but local mode returns three buckets.

## Real-server comparison

I ran the same points against Qdrant Server v1.19.0 (`74f3e85b`) with an exact facet and one typed payload index at a time:

```text
bool index:    False:1, True:1
integer index: 0:1, 1:1
keyword index: "0":1
```

This confirms that the server does not merge boolean and integer identity. It also exposes a design boundary: server facet is schema-indexed, while local mode currently permits an unindexed field and returns every accepted scalar type.

## Expected behavior / design decision

Could maintainers confirm which local-mode contract is preferred?

1. Preserve the current permissive local behavior, but keep JSON scalar types distinct. Internally key both per-point de-duplication and global aggregation by a tagged identity such as `(type_tag, normalized_value)`, while retaining the original value for `FacetValueHit`.
2. Move toward server parity by honoring a configured payload-index schema and ignoring/rejecting values of other types.

Option 1 is the smaller compatibility fix. Option 2 has stronger server parity but is broader because local payload indexes are currently not required for faceting.

## Regression requirements

- Separate `False` and `0` buckets with count 1 each.
- Separate `True` and `1` buckets with count 1 each.
- Apply the same type-tagged identity in both `values_set` and `facet_hits`.
- Preserve UUID-string normalization before identity construction.
- Keep #1377's deterministic mixed-type ordering and mixed-ID matrix behavior.

SiluPanda:
### All Submissions:

### Changes to Core Features:

## Summary

Fixes #1388.

Local-mode faceting currently uses raw Python values as both the per-point de-duplication key and the global aggregation key. Because `False == 0` and `True == 1`, boolean and integer payload values collapse into the same bucket before sorting.

This change switches local faceting to a type-tagged identity `(type_tag, value)` so that:

- `False` and `0` stay distinct
- `True` and `1` stay distinct
- per-point de-duplication and global counts use the same typed identity
- UUID normalization still happens before the identity is built
- facet sorting stays deterministic even when different scalar types tie on count

## Testing

- `pytest -q tests/test_in_memory.py`
- manual local reproduction of the issue now returns five distinct buckets for `False`, `0`, `True`, `1`, and `"0"`

### DIFF
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -1262,6 +1262,13 @@ class LocalCollection:
         facet_filter: types.Filter | None = None,
         limit: int = 10,
     ) -> types.FacetResponse:
+        # (bool, int, str). A value's position in this tuple is used as a type tag, so
+        # that `False`/`0` and `True`/`1` (equal and same-hash in python) don't collapse
+        # into a single bucket, and so that values of different types stay totally
+        # ordered when their counts tie. The server never mixes types in one facet (it
+        # reads a single payload index), so this cross-type order is local-mode only.
+        value_types = get_args_subscribed(types.FacetValue)
+
         facet_hits: dict[tuple[int, types.FacetValue], int] = defaultdict(int)
 
         mask = self._payload_and_non_deleted_mask(facet_filter)
@@ -1283,7 +1290,7 @@ class LocalCollection:
 
             # Sanitize to use only valid values
             for v in values:
-                if type(v) not in get_args_subscribed(types.FacetValue):
+                if type(v) not in value_types:
                     continue
 
                 # If values are UUIDs, format with hyphens
@@ -1291,7 +1298,7 @@ class LocalCollection:
                 if as_uuid:
                     v = str(as_uuid)
 
-                values_set.add(self._facet_value_identity(v))
+                values_set.add((value_types.index(type(v)), v))
 
             for v in values_set:
                 facet_hits[v] += 1
@@ -1967,17 +1974,6 @@ class LocalCollection:
         elif isinstance(point_id, int):
             return "", point_id
         raise TypeError(f"Incompatible point id type: {type(point_id)}")
-
-    @staticmethod
-    def _facet_value_identity(value: types.FacetValue) -> tuple[int, types.FacetValue]:
-        """Build a sortable key that preserves JSON scalar type identity."""
-        if isinstance(value, bool):
-            return 0, value
-        if isinstance(value, int):
-            return 1, value
-        if isinstance(value, str):
-            return 2, value
-        raise TypeError(f"Incompatible facet value type: {type(value)}")
 
     def scroll(
         self,
--- a/qdrant_client/local/local_collection.py
+++ b/qdrant_client/local/local_collection.py
@@ -1262,7 +1262,7 @@ class LocalCollection:
         facet_filter: types.Filter | None = None,
         limit: int = 10,
     ) -> types.FacetResponse:
-        facet_hits: dict[types.FacetValue, int] = defaultdict(int)
+        facet_hits: dict[tuple[int, types.FacetValue], int] = defaultdict(int)
 
         mask = self._payload_and_non_deleted_mask(facet_filter)
 
@@ -1279,7 +1279,7 @@ class LocalCollection:
                 continue
 
             # Only count the same value for each point once
-            values_set: set[types.FacetValue] = set()
+            values_set: set[tuple[int, types.FacetValue]] = set()
 
             # Sanitize to use only valid values
             for v in values:
@@ -1291,14 +1291,14 @@ class LocalCollection:
                 if as_uuid:
                     v = str(as_uuid)
 
-                values_set.add(v)
+                values_set.add(self._facet_value_identity(v))
 
             for v in values_set:
                 facet_hits[v] += 1
 
         hits = [
             models.FacetValueHit(value=value, count=count)
-            for value, count in sorted(
+            for (_, value), count in sorted(
                 facet_hits.items(),
                 # order by count descending, then by value ascending
                 key=lambda x: (-x[1], x[0]),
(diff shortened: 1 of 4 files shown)


linhongyu510 (approved):
Verified the updated head `cf98f25` against the two collision layers described in #1388. The implementation uses the same tagged identity for both the per-point de-duplication set and the global count dictionary, with the `bool` check before `int`, so Python's `bool` subclassing does not collapse the tags. The revised diff also removes the unrelated import-formatting churn from the previous SHA.

Local exact-head evidence: `pytest -q tests/test_in_memory.py` => `5 passed`. I also ran an additional matrix with `[False, 0]` on one point, `[True, 1, "0"]` on another, and three equivalent UUID spellings on one point. It produced separate one-count bool/int/string buckets while normalizing the UUID spellings into one one-count canonical string bucket:

```text
[('bool', False, 1), ('bool', True, 1), ('int', 0, 1), ('int', 1, 1), ('str', '0', 1), ('str', '550e8400-e29b-41d4-a716-446655440000', 1)]
```

No functional blocker found. A non-blocking test improvement would be to add the UUID-equivalence assertion to the committed regression, since preserving normalization was an explicit requirement in #1388. At review time the remote check rollup exposes only Netlify metadata, not Python test evidence, so the local result above should not be read as remote CI status.

SiluPanda:
Added the suggested UUID-equivalence case to the committed regression in `b58636f`: compact and hyphenated spellings on one point now assert one canonical, one-count UUID bucket alongside the distinct bool/int/string buckets. Re-ran `pytest -q tests/test_in_memory.py` (`5 passed`).

linhongyu510 (approved):
Verified the current head against `dev`. The typed identity is applied consistently to both per-point de-duplication and global aggregation; checking `bool` before `int` correctly avoids Python's `bool` subclass collision, UUID normalization still precedes identity construction, and the type tag provides deterministic cross-type ordering. The regression also covers same-point `False`/`0`, `True`/`1`, string identity, and UUID de-duplication. Local validation: `python3 -m pytest tests/test_in_memory.py qdrant_client/local/tests -q` — 79 passed; `git diff --check` passed. No blocking issue found.

SiluPanda:
Thanks again for the earlier approval, @linhongyu510. I rechecked the branch against `dev`: there have been no code changes since `1198fb3`, and the visible GitHub checks are green/neutral Netlify preview checks only. When you have a moment, could you please take another quick look? Thank you.

ErenAta16 (commented):
I checked this out and ran it against master locally, and the fix does what it
says. Keying the per-point set and the count dict by `(type_rank, value)` is the
right level to fix it at, because the merge happens at insertion and no sort key
applied later can undo it.

On master, with one point per value:

```
payload [False, 0, True, 1]  ->  [(bool, False, 2), (bool, True, 2)]
```

The int points are not just mis-sorted, they are absorbed, and the surviving
entry reports the bool type with an inflated count. On this branch the same
input gives four hits of count 1 each. The per-point dedupe case matters too:
a single point with `{"k": [False, 0]}` counted once on master and counts both
here, which is the behaviour the `StrictBool | StrictInt | StrictStr` union in
`FacetValue` implies.

The thing I would flag before this is merged is the relationship with #1377,
because it is not the one it looks like from the outside. Those two are not
duplicates and neither one is complete on its own. Measured on all three
branches, same inputs:

| | #1388 conflation | #1376 facet sort | #1376 `search_matrix` |
|---|---|---|---|
| master | broken | TypeError | TypeError |
| #1377 | still broken | fixed | fixed |
| #1389 | fixed | fixed | still TypeError |

This branch fixes the heterogeneous facet sort as a side effect, since once the
dict key carries a type rank the tie-break can never compare an `int` with a
`str` again. What it does not carry is the other half of #1377, the
`_search_distance_matrix` change, where `sorted(samples, key=lambda x: x.id)`
still raises on a collection that mixes integer and UUID point ids. So both are
worth merging, and whichever goes second will conflict: both edit the same
`sorted(facet_hits.items(), ...)` call and both add a static method immediately
after `_universal_id`.

One thing that made me more comfortable rather than less: the two branches
independently chose the same relative order, so `["a", 1]` comes out as
`[(int, 1), (str, "a")]` under either. That is worth stating explicitly because
the type rank is not cosmetic once `limit` truncates. With every count tied at 1
and `limit=2`, the rank alone decides which two values a caller sees, so the two
PRs disagreeing there would have been a user-visible difference rather than a
merge detail.

Minor, take it or leave it: the `raise TypeError` at the end of
`_facet_value_identity` is unreachable. The `type(v) not in
get_args_subscribed(types.FacetValue)` guard a few lines up already drops
anything that is not `bool`, `int` or `str`, and I confirmed a `float` payload
value never reaches the helper. It is harmless as a guard against a future
caller, though it does read as if the union were wider than it is.

Test-wise, `tests/test_in_memory.py` is 5 passed on this branch against 4 on
master, so the new case is the difference. The four `PermissionError` failures
in `tests/test_local_persistence.py` are present on master too, they are a
Windows tempdir cleanup artefact and nothing to do with this change.

joein (approved):

joein:
Thank you for the contribution @SiluPanda 

I've made some updates and now I think it is good to go.
I'll update and merge #1377 next

SiluPanda:
> Thank you for the contribution @SiluPanda
> 
> I've made some updates and now I think it is good to go. I'll update and merge #1377 next

thank you so much @joein

PR #1390, opened 2026-08-30
fix(local): preserve near-zero cosine vectors

Amaan00844:
## Summary

Local cosine normalization was rewriting near-zero vectors even when their norm was below `EPSILON`. That mutated stored vectors during local-mode search and made the client diverge from Qdrant server behavior.

This PR keeps near-zero vectors unchanged unless their norm is above the epsilon guard, matching server semantics.

## Changes

- guard cosine normalization with `EPSILON`
- apply the same behavior in dense and multivector local paths
- add a regression test covering near-zero cosine vectors

## Why

Previously, the local code used a pattern equivalent to:

- normalize if norm != 0
- otherwise fall back to `EPSILON`

That meant tiny but non-zero vectors were rewritten during cosine scoring, which is not what the server does.

## Validation

- `pytest -q qdrant_client/local/tests/test_distances.py`

### DIFF
--- a/qdrant_client/local/distances.py
+++ b/qdrant_client/local/distances.py
@@ -128,15 +128,15 @@ def cosine_similarity(query: types.NumpyArray, vectors: types.NumpyArray) -> typ
         distances
     """
     vectors_norm = np.linalg.norm(vectors, axis=-1)[:, np.newaxis]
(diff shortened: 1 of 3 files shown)


Amaan00844:
Fixed the local cosine-distance edge case near zero by making the computation stable for near-parallel vectors. I also added a regression check to cover this scenario and verified it with the targeted distance tests.

PR #1393, opened 2026-09-01
new: QdrantServerless client prototype

generall:
> **Stacked on #1394** (CI congruence fixes) to keep integration tests green; merge #1394 first, then retarget this PR to `dev`.

## What

Prototype of a dedicated client for Qdrant Serverless, parallel to the regular `QdrantClient`, per the serverless client proposal.

```python
from qdrant_client.serverless import QdrantServerless, DenseVectorConfig, Distance

client = QdrantServerless(url="https://serverless....qdrant.io", api_key="...")

client.create_collection(          # serverless-specific input
    "my-collection",
    dense_vectors=DenseVectorConfig(size=1536, distance=Distance.COSINE),
)
client.get_collection("my-collection")   # serverless-specific output
client.query_points("my-collection", query=[...])  # same as regular qdrant
```

## Why

Serverless is not 100% identical to a regular cluster deployment: collection management exposes only a simplified tenant-facing config (no quantization/WAL/segments), and point operations don't support read consistency, shard selection, write ordering, or filtered updates. A separate client keeps those parameters out of the serverless interface instead of forcing one client to paper over the divergence.

## How

Everything lives in a dedicated `qdrant_client/serverless/` module; nothing is added to the top-level package, so regular users don't pay any import cost.

- **gRPC stubs** generated from [`qdrant-cloud-public-api`'s `serverless/collections.proto`](https://github.com/qdrant/qdrant-cloud-public-api/blob/main/proto/qdrant/serverless/collections.proto) with the same pinned grpcio-tools 1.62.0 as the existing stubs (`tools/generate_serverless_grpc_client.sh` regenerates). The proto is renamed to `serverless_collections.proto` because the protobuf descriptor pool registers files by name and `collections.proto` is already taken by the regular client — without the rename, importing both clients in one process crashes. Generated types stay internal.
- **Models**: hand-written pydantic models (v1.10+ compatible) mirroring the tenant-facing config — `DenseVectorConfig`, `SparseVectorConfig`, `PrecisionTier`, the eight payload-index types, `CollectionConfig`, `CollectionInfo`, `CollectionSummary`. The existing `Distance` / `TokenizerType` enums are reused so values look familiar. These may later be replaced with models generated from a serverless OpenAPI spec.
- **Client**: `QdrantServerless` wraps an internal `QdrantRemote(prefer_grpc=True, check_compatibility=False)` and delegates point operations to it, reusing all existing conversion machinery; collection methods talk to the new `CollectionsService` stub on the same channel (one place for TLS, `api-key` metadata, user-agent). The API divergence is expressed purely in signatures:
  - point methods (`query_points`, `upsert`, `retrieve`, `scroll`, `count`, `delete`, `set_payload`, `delete_payload`) are the regular ones minus `consistency` / `shard_key_selector` / `ordering`
  - `delete` / payload methods take `ids` only, since serverless rejects filtered updates
  - `create_collection` accepts a bare config as the unnamed `""` vector and returns the service's result string (`"created"` / `"already exists"`); `get_collection` returns `CollectionInfo` with `exists` / `point_count` instead of raising on missing collections
  - default `grpc_port` is 443, not 6334

## Not included (mechanical follow-ups)

`query_batch_points` / `batch_update_points` / `update_vectors`, `update_collection` (no such RPC yet).

## Async client

`AsyncQdrantServerless` is generated from the sync client the same way the regular async client is: `tools/async_client_generator/serverless_generator.py` runs as part of `tools/generate_async_client.sh`, delegates to `AsyncQdrantRemote`, awaits the CollectionsService stub RPCs on the aio channel, and drops the sync context manager (the regular async client has none either). The generated file is verified by `tests/async-client-consistency-check.sh` alongside the other generated clients.

## Testing

- `tests/test_serverless.py`: model↔gRPC round-trips, proto3 optional-field presence, offline client construction (headers, port, TLS)
- mypy clean on the new module
- not yet run against a live serverless endpoint


joein:
The example code crashes with the following error:


```python
created
[CollectionSummary(collection_name='my-collection', point_count=None)]
exists=True config=CollectionConfig(dense_vectors={'': DenseVectorConfig(size=4, distance=<Distance.COSINE: 'Cosine'>, multivector=False, precision_tier=None)}, sparse_vectors={}, payload_indexes={'color': KeywordIndex(type='keyword')}) point_count=None
Traceback (most recent call last):
  File "../qdrant/qdrant_client/examples/serverless_client.py", line 52, in <module>
    main()
  File "../qdrant/qdrant_client/examples/serverless_client.py", line 45, in main
    print(client.query_points("my-collection", query=[0.1, 0.2, 0.3, 0.4]))
  File "../qdrant/qdrant_client/qdrant_client/serverless/client.py", line 355, in query_points
    return self._remote.query_points(
  File "../qdrant/qdrant_client/qdrant_client/qdrant_remote.py", line 535, in query_points
    res: grpc.QueryResponse = self.grpc_points.Query(
  File "../qdrant/qdrant_client/venv/lib/python3.10/site-packages/grpc/_interceptor.py", line 276, in __call__
    response, ignored_call = self._with_call(
  File "../qdrant/qdrant_client/venv/lib/python3.10/site-packages/grpc/_interceptor.py", line 331, in _with_call
    return call.result(), call
  File "../qdrant/qdrant_client/venv/lib/python3.10/site-packages/grpc/_channel.py", line 447, in result
    raise self
  File "../qdrant/qdrant_client/venv/lib/python3.10/site-packages/grpc/_interceptor.py", line 314, in continuation
    response, call = self._thunk(new_method).with_call(
  File "../qdrant/qdrant_client/venv/lib/python3.10/site-packages/grpc/_channel.py", line 1182, in with_call
    return _end_unary_response_blocking(state, call, True, None)
  File "../qdrant/qdrant_client/venv/lib/python3.10/site-packages/grpc/_channel.py", line 999, in _end_unary_response_blocking
    raise _InactiveRpcError(state)  # pytype: disable=not-instantiable
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.INVALID_ARGUMENT
	details = "Not existing vector name error: "
	debug_error_string = "INVALID_ARGUMENT:Not existing vector name error: "
>
```


In the beginning, it failed just on the first attempt and then worked when I reran it.
I tried running it again, and got the problem with not existing vector name 5 times in a row before it worked.

joein (approved):

PR #1407, opened 2026-09-05
fix(local): honor nested json-path keys in delete_payload

winklemad:
## Problem

In local mode, `delete_payload` only removed top-level dict keys:

```python
for key in keys:
    if key in self.payload[idx]:
        self.payload[idx].pop(key)
```

A key given as a json path — `a.b`, `location[0].name`, `location[].name` — never matches a top-level key, so the delete is a silent no-op. The server deletes nested keys via dot notation and preserves the rest of the payload, so local mode diverged from it.

```python
# ... upsert id=1 with payload {"a": {"b": 1, "c": 2}, "top": 9}
client.delete_payload("t", keys=["a.b"], points=[1])
client.retrieve("t", ids=[1], with_payload=True)[0].payload
# local:  {"a": {"b": 1, "c": 2}, "top": 9}   <- a.b not removed
# server: {"a": {"c": 2}, "top": 9}
```

## Change

`set_payload` and filters already resolve these paths through `parse_json_path`; `delete_payload` was the one payload operation ignoring them. This adds a `delete_value_by_key` helper next to `set_value_by_key` that walks the same `JsonPathItem` path and removes the leaf — a path that does not resolve is a no-op, and sibling values are preserved — then calls it from `delete_payload`. Top-level deletes behave exactly as before.

## Tests

- `tests/local/test_delete_value_by_key.py` — unit coverage for the helper: top-level, nested dict path, deep path, array index, array wildcard, non-existent path no-op, and traversal through a non-dict. Fails on the current code (the helper does not exist).
- `tests/congruence_tests/test_payload.py::test_delete_payload_with_nested_key` — asserts local and remote agree for a nested dict key, a non-existent nested path, an array-wildcard field, and mixed top-level + nested keys.

## Checklist

- [x] Branch created from `dev` and targets `dev`
- [x] Followed the contributing conventions (sole-author, conventional-commit title, pre-commit ran clean)
- [x] Checked for duplicate open PRs


winklemad:
Good catch on the indexed-deletion semantics — I traced it through the server source and fixed both divergences:

- **Terminal array index is now a no-op.** In `lib/segment/src/json_path/mod.rs`, `value_remove` leaves the leaf `(Index, Array)` case empty ("Deleting array indices is not idempotent, so we don't support it"). Local mode was doing `del data[i]` there, so it now matches the server and does nothing.
- **Negative indices no longer traverse or delete.** The server parses indices as `usize` and addresses elements with a non-negative index, so `loc[-1]`/`loc[-1].x` can't be represented server-side. Local mode was falling through to Python's negative indexing; nested traversal now requires `0 <= index < len`, so those paths no-op too.

Added local cases for `loc[0]`, `loc[-1]`, `loc[-1].x` and a terminal-wildcard clear, plus a `loc[0]` congruence case (both sides no-op). I kept `parse_json_path` untouched so `set_payload` and filter behavior are unchanged — this stays scoped to the delete path.

joein (approved):

PR #1414, opened 2026-09-10
fix(local): skip bool group-by keys to match server GroupId semantics

### LINKED ISSUE #1413: fix(local): group_by admits bool payload values that the server never groups
### Description

Local mode (`query_groups` / `search_groups` in `qdrant_client/local/local_collection.py`) groups points by a bool payload value together with the equal integer, while the server never puts booleans into any group.

### Minimal reproduction (pure local mode, `:memory:`)

```python
from qdrant_client.http import models

client = QdrantClient(location=":memory:")
client.create_collection(
    "g",
)
client.upsert(
    "g",
    points=[
        models.PointStruct(id=1, vector=[1.0, 0.0], payload={"g": True}),
        models.PointStruct(id=2, vector=[1.0, 0.0], payload={"g": 1}),
        models.PointStruct(id=3, vector=[1.0, 0.0], payload={"g": False}),
        models.PointStruct(id=4, vector=[1.0, 0.0], payload={"g": 0}),
    ],
)
res = client.query_points_groups("g", group_by="g", query=[1.0, 0.0], limit=10, group_size=10)
for grp in res.groups:
    print(repr(grp.id), [p.id for p in grp.hits])
```

Actual local output:

```
1 [2, 1]   # point 1 (True) collapsed into the int-1 group
0 [4, 3]   # point 3 (False) collapsed into the int-0 group
```

### Root cause

Both grouping paths share the same hunk (`query_groups` ~L1190 and `search_groups` ~L1263):

```python
group_values = list(set(v for v in group_values if isinstance(v, (str, int))))
```

In Python `isinstance(True, int)` is `True`, and `True == 1` / `hash(True) == hash(1)`, so bools pass the filter and then collapse with the equal int inside `set()`.

The server cannot do this: `GroupId::try_from(&JsonValue)` (`lib/segment/src/data_types/groups.rs`) maps `Bool(_) => Err(())` ("Only allows Strings and Numbers to be converted into GroupId"), and `GroupsAggregator::add_points` (`lib/shard/src/grouping/aggregator.rs`) ignores `BadKeyType` points — a bool-valued point is silently left out of every group, and the unit test pins `json!(false) => Err(AggregatorError::BadKeyType)` with zero groups created.

### Relation to already-merged work

Same root cause (Python bool/int equality) as #1259 (filters) and #1389 (facet). `facet()` was hardened with exact-type checks; the two grouping hunks were left behind.

### Proposed fix

Mirror the exact-type convention in both grouping hunks — admit only values whose type is exactly `str` or `int` (bools skipped, matching `GroupId::try_from`):

```python
group_values = list(set(v for v in group_values if type(v) in (str, int)))
```

plus a pure-local regression test (`qdrant_client/local/tests/test_groups.py`, no server needed). Happy to submit a PR targeting `dev` if this direction looks right.

feiiiiii5:
Fixes #1413

### All Submissions:

* [x] Contributions target the `dev` branch (branched from `dev` @ `e5e26b8`)
* [x] Followed the guidelines in our Contributing document (no CONTRIBUTING.md in-repo; PR template checklist followed)
* [x] Checked for other open PRs for the same change (search: `group_by bool` / `group bool` in title+body, open+closed — only #1259 filters and #1389 facet, both merged family members)

### Changes to Core Features:

* [x] Explanation below; [x] new regression tests; [x] ran tests locally (see Test)

## Root Cause

Both grouping paths (`LocalCollection.query_groups` and `search_groups`) filter group-by values with `isinstance(v, (str, int))`. In Python `bool` subclasses `int`, so `True`/`False` pass the filter, and since `True == 1` with equal hashes, `set()` collapses them into the int groups: point {`g`: True} lands in group `1`, {`g`: False} in group `0`.

The server can never do this: `GroupId::try_from(&JsonValue)` (`qdrant/lib/segment/src/data_types/groups.rs`) maps `Bool(_) => Err(())` — "Only allows Strings and Numbers to be converted into GroupId" — and `GroupsAggregator::add_points` (`qdrant/lib/shard/src/grouping/aggregator.rs`) ignores `BadKeyType` points, pinned by its unit test (`json!(false) => Err(AggregatorError::BadKeyType)`, zero groups). Same root cause as #1259 (filters) and #1389 (facet); the two grouping hunks were left behind when `facet()` was hardened with exact-type checks.

## Fix

Exact-type filter in both hunks, mirroring `facet()`'s convention (bools skipped, matching `GroupId::try_from`):

```python
group_values = list(set(v for v in group_values if type(v) in (str, int)))
```

## Test

- [x] New pure-local regression suite `qdrant_client/local/tests/test_groups.py` (both paths: `query_points_groups` end-to-end + `LocalCollection.search_groups` twin hunk; covers True/1, False/0, "1", float 1.5, missing key)
- [x] pre-fix FAIL proof (unmodified `e5e26b8`): `{1: [1, 2]} != {1: [2]}`, `{0: [3, 4]} != {0: [4]}` — exactly the predicted collapse, both tests
- [x] post-fix: new suite 2 passed; full `qdrant_client/local/tests/` 82 passed; `tests/test_in_memory.py` 4 passed
- [x] `ruff format --check --line-length=99` clean on both files (repo-pinned 0.4.3 config; checked with system ruff)

## Diff scope

2 files, +87/-2 (core 1 file, +12/-2 incl. comments; test 1 file, +75/-0)

## Not PR-related failures

`tests/test_qdrant_client.py`: 67 failed / 8 passed identically on the unmodified base (verified via `git stash`) — pre-existing environment failures (server-backed fixtures unavailable here), unrelated to this change.


joein (approved):

joein:
thank you @feiiiiii5 for pointing it out and fixing it!

PR #1425, opened 2026-09-13
fix(local): ignore nonnumeric coordinates in geo filters

### LINKED ISSUE #1422: Local geo filters crash on nonnumeric payload coordinates
### Description

In local mode, a geo filter raises `TypeError` if any stored payload has `lat` or `lon` set to a nonnumeric JSON value, such as `null` or a string. The payload is accepted by `upsert`, but subsequently prevents the query from returning even the valid points. Boolean coordinates can instead be incorrectly treated as numbers.

This affects radius, bounding-box, and polygon filters. The server's corresponding [condition checkers](https://github.com/qdrant/qdrant/blob/6ab21cac18ebb6f4ae29102c7f8f5cc11affd5de/lib/segment/src/payload_storage/condition_checker.rs#L330) only apply geometry operations when both coordinates are JSON numbers; other values do not match. Local mode currently checks only that the two keys exist.

### Reproduction

```python

client.create_collection(
    "geo", vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT)
)
client.upsert("geo", points=[
    models.PointStruct(id=1, vector=[1.0, 0.0], payload={
        "location": {"lon": None, "lat": 0}
    }),
    models.PointStruct(id=2, vector=[1.0, 0.0], payload={
        "location": {"lon": 0, "lat": 0}
    }),
])
points, _ = client.scroll("geo", scroll_filter=models.Filter(must=[
    models.FieldCondition(
        key="location",
        geo_radius=models.GeoRadius(center=models.GeoPoint(lon=0, lat=0), radius=1000),
    )
]))
print([point.id for point in points])
```

Expected: `[2]`. Actual: `TypeError` during distance calculation.

Reproduced against `dev` at `bdb2fa67d93632d980627bfc9e83e7f8802366bc`, Python 3.12.12, macOS/ARM64. Public-client regression tests fail for all three geo condition types. No embeddings or server are required for the local reproduction; server behavior above was checked in source.

pentaoa:
Fixes #1422.

Local geo filters currently pass any payload object containing `lat` and `lon` into geometry calculations. A record with a null or string coordinate can therefore raise `TypeError` and abort a query that should still return other valid records. Boolean coordinates can incorrectly match because Python treats booleans as integers.

Require both coordinates to be JSON numbers before evaluating radius, bounding-box, or polygon filters. This follows the server's `as_f64()` checks. A public-client regression covers all three filters, integer and float coordinates, nonnumeric coordinates on either axis, and an array containing both a nonmatching value and a valid location.

### Validation

- Before the fix, all three new regression cases fail with `TypeError`.
- `pytest tests/test_in_memory.py qdrant_client/local/tests -q`: **125 passed**.
- Ruff checks, formatting, and `git diff --check` pass.
- Server-backed integration tests were not run; the server behavior was checked against `condition_checker.rs` at `6ab21cac18ebb6f4ae29102c7f8f5cc11affd5de`.

### All Submissions

- [x] Targets `dev` and is based on `dev`.
- [x] Checked existing pull requests for the same change.

### Changes to Core Features

- [x] Explained the change and its purpose.
- [x] Added regression coverage.
- [x] Ran the applicable local tests.

Prepared with OpenAI Codex.


joein (approved):
thanks for fixing it!

PR #1427, opened 2026-09-13
fix(local): update IDF statistics on deletion

### LINKED ISSUE #1426: Local IDF statistics retain deleted sparse vectors and produce negative scores
### What happened?

In local mode, deleting points or named sparse vectors leaves their document frequencies in the global IDF counters. The number of live vectors does decrease, so a normal deletion can produce negative scores. Global IDF and an IDF corpus selecting every point then disagree. Closing and reopening a persistent collection rebuilds the counters and changes the scores again.

### Reproduction

```python
from qdrant_client import QdrantClient, models as m

client.create_collection(
    "test", vectors_config={},
    sparse_vectors_config={"text": m.SparseVectorParams(modifier=m.Modifier.IDF)},
)
vector = m.SparseVector(indices=[0], values=[1.0])
client.upsert("test", [m.PointStruct(id=i, vector={"text": vector}) for i in range(3)])
client.delete("test", points_selector=[1, 2])
for params in [None, m.SearchParams(idf=m.IdfCorpusParams(corpus=m.Filter()))]:
    result = client.query_points("test", using="text", query=vector, search_params=params)
    print(result.points[0].score)
client.close()
```

Actual: approximately `-0.559616` for global IDF, `0.287682` for the match-all corpus. Both should use the remaining document and produce `log(2 / 1.5)`, approximately `0.287682`. Replacing `delete` with `delete_vectors("test", vectors=["text"], points=[1, 2])` has the same problem.

Reproduced against `dev` at `bdb2fa67d93632d980627bfc9e83e7f8802366bc`, Python 3.12.12 on macOS arm64. No Qdrant server or embedding model is needed.

pentaoa:
Fixes #1426.

Deleting points or named sparse vectors in local mode reduces the live corpus size but leaves their term frequencies in the global IDF counters. After inserting three identical sparse vectors and deleting two, the remaining score becomes `-0.559616`; a match-all IDF corpus correctly gives `0.287682`. This can change rankings and threshold filtering, and reopening a persistent collection changes the result again.

Remove each live sparse vector's frequencies when it is deleted. Respect deletion state when replacing vectors so repeated deletions and subsequent upserts or vector updates neither double-subtract nor count deleted points. The update remains incremental; searches do not need to rescan the collection.

Regression coverage exercises point deletion, vector deletion followed by either upsert or vector update, repeated deletion, records without the vector, and persistence, including updating an already-deleted point and reopening without restoring it. It checks actual query scores against the IDF formula and verifies that global and match-all corpus scopes agree. The deletion regressions failed before the fix; all **141 relevant local tests pass** afterward. Ruff checks, formatting and `git diff --check` pass. Server-backed integration tests were not run.

### All Submissions

- [x] Targets `dev` and is based on `dev`.
- [x] Read the repository contribution guidance.
- [x] Checked existing pull requests for this change.

### Changes to Core Features

- [x] Explained the change and its purpose.
- [x] Added regression tests.
- [x] Ran the applicable tests locally.

Prepared with OpenAI Codex.


joein:
Hey @pentaoa 

Thanks for proposing a fix!
After looking into the problem I have found something more to fix and updated the PR, I am about to merge it rn!

joein (approved):

PR #1428, opened 2026-09-13
test(models): add unit test suite for unsigned int and UUID point ID normalization

gcoinstash-cmd:
### Summary
Adds unit test verification for Point ID type normalization:
- Validates positive integer IDs remain unchanged.
- Asserts string UUID casing normalization.


joein:
Hey @gcoinstash-cmd 

I am not sure what this tests is checking, since it does not import anything from qdrant-client

Other than this, all the PRs should point to the dev branch and should be created from it

PR #1429, opened 2026-09-13
fix: reject 0 raised to a negative exponent in formula PowExpression

bijay-odyssey:
## Problem

`evaluate_expression`'s `PowExpression` branch decides whether an input is valid like this:

```python
if base >= 0 or (base != 0 and float(exponent).is_integer()):
    return math.pow(base, exponent)
```

`base >= 0` is true when `base == 0`, regardless of what the exponent is. But `0` raised to a negative exponent is a pole (division by zero) — mathematically undefined, same as a negative base with a non-integer exponent, which this same function already rejects cleanly.

Instead of being rejected the same way, `math.pow(0, negative_exponent)` raises a raw `ValueError: math domain error` that isn't caught by the surrounding `except OverflowError`, so it leaks straight out of `query_points` instead of going through `raise_non_finite_error`, which every other non-finite case in this function (`DivExpression`, `SqrtExpression`, `Log10Expression`, `LnExpression`, `PowExpression`'s own negative-base case, ...) uses for a consistent, descriptive message.

Repro:

```python

client.create_collection("test", vectors_config=models.VectorParams(size=2, distance=models.Distance.COSINE))
client.upsert("test", points=[models.PointStruct(id=1, vector=[1.0, 0.0], payload={"base": 0.0, "exp": -1.0})])

formula = models.FormulaQuery(formula=models.PowExpression(pow=models.PowParams(base="base", exponent="exp")))
client.query_points("test", query=formula, prefetch=models.Prefetch(query=[1.0, 0.0], limit=10))
# ValueError: math domain error
# expected: ValueError: The expression 0.0^-1.0 produced a non-finite number
```

## Fix

Split the `base >= 0` check into `base > 0` (always fine) and `base == 0` (fine only when the exponent is non-negative), leaving the existing `base < 0` + integer-exponent branch untouched. `0` raised to a non-negative exponent is unaffected (`0**0 == 1`, `0**3 == 0`).

## Testing

Added `test_pow_expression` to `qdrant_client/local/tests/test_formula.py`, covering the previously-working cases (positive base, negative base with integer/non-integer exponent) plus the new `0 ** negative` regression.

- `pytest qdrant_client/local/tests/test_formula.py`: 4 passed
- Confirmed the new test fails on unpatched code (raw `math domain error`) and passes with the fix
- `pytest tests/test_in_memory.py tests/test_local_persistence.py tests/test_common.py`: 33 passed, 1 pre-existing unrelated failure (`test_prevent_parallel_access`, a Windows temp-dir lock cleanup issue, reproduces identically on unpatched `dev`)
- `mypy qdrant_client/hybrid/formula.py --disallow-incomplete-defs --disallow-untyped-defs`: clean
- `ruff format --check --line-length=99`: clean

## Compatibility

Purely a bug fix to an error path — no change to any currently-valid input's result. The only behavior change is that `0 ** negative_exponent` now raises the same clean, documented error as every other non-finite formula result, instead of a raw Python exception.


joein:
thanks for the contribution! @bijay-odyssey 

I've made some adjustments and as soon as the CI is green, I'll merge it

PR #1431, opened 2026-09-14
fix: align local mode null handling with the server (IsNullCondition, MatchExcept)

### LINKED ISSUE #1430: Local mode: null handling diverges from the server (IsNullCondition on arrays, MatchExcept on null)
Two null-handling semantics in local mode diverge from the server (lib/segment/src/payload_storage/condition_checker.rs):

1. **IsNullCondition does not match an array field containing a null element.** The server treats such an array as null (`Value::Array(array) => array.iter().any(|x| x.is_null())`; the server's own test fixture in query_checker.rs carries `"null_array": [null, 1]`). Local mode fetches the field with `flat=False` and checks `any(v is None for v in values)`, comparing the array itself to None, so it never matches.

2. **MatchExcept matches a null payload value locally.** On the server, `Match::Except` maps `(Value::Null, _) => false`. Local mode negates `values_match(None, v)` to True, so the point matches.

Reproducer (qdrant-client master, 1.19.0):

```python
from qdrant_client.http import models

client.create_collection("t", vectors_config=models.VectorParams(size=4, distance=models.Distance.COSINE))
client.upsert("t", points=[
    models.PointStruct(id=1, vector=[0.1]*4, payload={"a": [1, None, 3]}),
    models.PointStruct(id=2, vector=[0.1]*4, payload={"a": None}),
])

# Server: point 1 matches. Local: no results.
print(client.scroll("t", scroll_filter=models.Filter(must=[
    models.IsNullCondition(is_null=models.PayloadField(key="a"))])))

# Server: no results. Local: point 2 matches.
print(client.scroll("t", scroll_filter=models.Filter(must=[
    models.FieldCondition(key="a", match=models.MatchExcept(**{"except": [1]}))])))
```

Both reproduced twice with identical output. I have a fix ready with regression tests (red on master, green with the patch, existing suites unaffected) and can open a PR.

Harsh23Kashyap:
Closes #1430

Two one-line fixes in `qdrant_client/local/payload_filters.py`, matching the server semantics in `lib/segment/src/payload_storage/condition_checker.rs`:

1. `IsNullCondition` now matches an array field that contains a null element (server: `array.iter().any(|x| x.is_null())`).
2. `MatchExcept` no longer matches a null payload value (server: `(Value::Null, _) => false`).

Regression tests in `tests/test_in_memory.py`: both fail on master, pass with the patch; the existing `test_in_memory.py` and `test_local_persistence.py` suites pass unchanged.


joein:
This has already been handled in dev branch
Thank you!

PR #1434, opened 2026-09-14
fix(local): honor str and non-list Sequence selectors in local mode

### LINKED ISSUE #1433: Local mode ignores str/tuple selectors for with_vectors and with_payload, silently returning everything
### Description

In local mode (`QdrantClient(":memory:")` / `path=...`), the `with_vectors` and `with_payload` parameters only take effect when passed a `list`. Both are typed `bool | Sequence[str]`, so a bare string or a tuple is a valid selector — but `LocalCollection` checks `isinstance(..., list)` and falls through to "return everything".

### Reproduction

```python

client.create_collection(
    "t",
    vectors_config={
        "a": models.VectorParams(size=2, distance=models.Distance.COSINE),
        "b": models.VectorParams(size=2, distance=models.Distance.COSINE),
    },
)
client.upsert("t", [models.PointStruct(id=1, vector={"a": [0.1, 0.2], "b": [0.3, 0.4]},
                                       payload={"x": 1, "y": 2})])

client.retrieve("t", [1], with_vectors="b")[0].vector
# actual:   {'a': [...], 'b': [...]}   (all vectors)
# expected: {'b': [...]}
# with_vectors=["b"] works correctly

client.retrieve("t", [1], with_payload="x")[0].payload
# actual:   {'x': 1, 'y': 2}           (whole payload)
# expected: {'x': 1}
# same wrong result for with_payload=("x",) and with_vectors=("b",)
```

A `str` is a `Sequence[str]`, so `with_vectors="b"` type-checks and means "only vector b"; iterating it as a selector would also be wrong (it would look up vectors named 'b'-per-character — but the current code does not even do that, it ignores the selector entirely). Returning more data than requested is a silent correctness issue: callers can ship payloads/vectors they explicitly did not ask for, and any code post-processing the selection sees phantom fields.

The same applies to every read path that funnels through `LocalCollection._get_vectors` / `_process_payload`: `retrieve`, `scroll`, `query_points`, `search`, `recommend`, `discover`, groups, etc.

### Proposed fix

In `_get_vectors` and `_process_payload`, normalize `isinstance(x, str)` to `[x]` and match `Sequence` rather than `list`, so bare strings and tuples are honored.

Reproduced on `dev` at bdb2fa67 (Python 3.12, Linux).

betacatsling:
## Summary

Fixes #1433.

In local mode, `with_vectors` and `with_payload` selectors were only honored when passed as a `list`. Both are typed `bool | Sequence[str]`, so a bare string (`with_vectors="image"`) or a tuple (`with_payload=("city",)`) is a valid selector — but `LocalCollection` checked `isinstance(..., list)` and silently returned everything:

- `retrieve(..., with_vectors="image")` returned **all** vectors instead of only `image`
- `retrieve(..., with_payload="city")` returned the **whole** payload instead of `{city: ...}`

This normalizes a bare `str` to a single-element list and matches any `Sequence`, so local mode honors the documented selector contract instead of silently over-returning. The fix lives in `LocalCollection._get_vectors` and `LocalCollection._process_payload`, so it applies to every read path (`retrieve`, `scroll`, `query_points`, `search`, groups, ...).

## Validation

- New regression test `tests/test_in_memory.py::test_in_memory_selector_accepts_str_and_tuple` exercises str / list / tuple selectors for both `with_vectors` and `with_payload` via `retrieve` and `query_points`:

```
$ python -m pytest tests/test_in_memory.py tests/test_local_persistence.py -q
10 passed in 1.54s
```

- The new test fails on `dev` without the fix (`AssertionError: 'text'` — the unrequested vector was returned).
- `ruff format --check` (line-length 99, per `.pre-commit-config.yaml`) passes on the changed files; `mypy` reports no issues for `local_collection.py`.

### All Submissions:


joein:
Hey @betacatsling 

Thanks for pointing it out and proposing a fix!

Actually, we didn't want to accept anything other than what's included into `models.WithPayloadInterface`. 
I've made some changes, so now the interface looks like it was supposed to look like.
But since we did support tuples, in order not to break the interface, I added some helpers to support them,
Though, a single string is still not supported.

joein (approved):

PR #1445, opened 2026-09-16
Clarify local-mode error for wrong-kind query vectors (e.g. 1-D query against a multivector)

### LINKED ISSUE #1444: Local mode: wrong-kind query vector reports the named vector as 'not found' (e.g. 1-D query against a multivector)
### What happened?

In local mode, querying a collection with a vector of the **wrong kind** for the named vector produces a misleading `ValueError`.

When a collection has a named multivector and the query is a 1-D vector (or vice versa: a 2-D query against a dense named vector), the validation falls into the wrong branch and raises:

```
Dense vector mv is not found in the collection        # 'mv' is a multivector, 1-D query given
Multivector  is not found in the collection           # unnamed dense collection, 2-D query given
```

The vector *does* exist — the query format is wrong. The message even shows an empty name in the unnamed case, which makes debugging worse.

### Reproduction (local mode, qdrant-client main)

```python

client = QdrantClient(path=":memory:")
client.create_collection("m", vectors_config={
    "mv": models.VectorParams(size=2, distance=models.Distance.COSINE,
        multivector_config=models.MultiVectorConfig(comparator=models.MultiVectorComparator.MAX_SIM))})
client.upsert("m", points=[models.PointStruct(id=1, vector={"mv": [[1.0, 0.0]]})])

client.query_points("m", query=[1.0, 0.0], using="mv", limit=5)
# ValueError: Dense vector mv is not found in the collection  <- misleading
```

Server-side correctly requires a 2-D query for multivector collections (per the docs), so the local rejection itself is right — the error is just wrong about the reason.

### Suggested fix

When the named vector exists but as a different kind, raise an error that says so:

```
Vector 'mv' is a multivector vector, but the query vector is a single vector (1-D).
A multivector collection requires a matching query format.
```

Truly missing names keep the original "not found" message. Happy to open a PR with the fix + tests.

warter666:
Fixes #1444

## Problem

In local mode, a query vector of the wrong kind for an existing named vector (1-D query against a multivector, or a 2-D query against a dense vector) raised `Dense vector mv is not found in the collection` / `Multivector  is not found in the collection`. The vector exists; the query format is wrong — and in the unnamed case the message shows an empty name, which makes this especially confusing to debug.

The rejection itself is correct (Qdrant server also requires a 2-D query for multivector collections); only the message was misleading.

## Change

`LocalCollection.search()` now distinguishes "name exists but as a different vector kind" from "name missing" via a small helper, producing e.g.:

```
Vector 'mv' is a multivector vector, but the query vector is a single vector (1-D).
A multivector collection requires a matching query format.
```

Truly missing names keep the original "not found" wording. Applies to all three kinds (dense / sparse / multivector).

## Tests

`test_query_vector_kind_mismatch_error` covers both mismatch directions, the truly-missing fallback, and the happy path. Existing local tests pass.


warter666:
Fixed in the new commit: the default vector (empty name) now renders as `'unnamed vector (default)'` in the mismatch message, with a regression test (`test_query_vector_kind_mismatch_unnamed`) covering the 2-D query against an unnamed dense collection. Tests 3/3, lint clean.

warter666:
Added — `test_query_vector_kind_mismatch_sparse` covers all three suggested cases: a sparse query against the configured dense name and the configured multivector name (asserting the accurate kind-mismatch messages), and against a missing name (asserting the preserved `Sparse vector ... is not found` message). Also asserts the sparse happy path still works against a configured sparse vector. Tests 4/4, lint clean.

PR #1447, opened 2026-09-16
fix(local): reject empty relevance feedback

### LINKED ISSUE #1446: Local mode accepts relevance-feedback queries with an empty feedback list
## Problem

An empty relevance-feedback list is rejected by Qdrant server validation, but local mode accepts the same query and runs a normal target-vector search instead.

This makes a query behave differently when an application switches between `QdrantClient(":memory:")` and a server-backed client.

## Reproduction

```python

client.create_collection(
    "demo",
    vectors_config=models.VectorParams(size=2, distance=models.Distance.DOT),
)
client.upsert(
    "demo",
    [
        models.PointStruct(id=1, vector=[1.0, 0.0]),
        models.PointStruct(id=2, vector=[0.0, 1.0]),
    ],
)

query = models.RelevanceFeedbackQuery(
    relevance_feedback=models.RelevanceFeedbackInput(
        target=[1.0, 0.0],
        feedback=[],
        strategy=models.NaiveFeedbackStrategy(
            naive=models.NaiveFeedbackStrategyParams(a=0.5, b=1.0, c=0.7)
        ),
    )
)

print(client.query_points("demo", query=query).points)
```

On current `master` (`cf747f4`), local mode returns:

```text
[ScoredPoint(id=1, score=0.5, ...), ScoredPoint(id=2, score=0.0, ...)]
```

The server rejects the request with `feedback elements must be non-empty`.

I reproduced the local result twice from a clean checkout. The mismatch comes from local query validation checking the naive strategy's `b` bound but not the server's non-empty feedback rule.

## Expected behavior

Local mode should reject an empty feedback list before executing the query, matching the server.

Harsh23Kashyap:
## Summary
- reject empty relevance-feedback requests in both sync and async local clients
- add a congruence test covering empty positive and negative feedback

Closes #1446.

## Validation
- `python -m compileall qdrant_client/qdrant_client.py qdrant_client/async_qdrant_client.py tests/congruence_tests/test_query.py`
- targeted test collected successfully: `tests/congruence_tests/test_query.py::test_query_empty_feedback` (1 selected)
- local rejection check ran twice with identical output: `ValueError: relevance feedback must not be empty`

The server-backed congruence test was not executed because no Qdrant server was available at `localhost:6333`.


joein:
Hey @Harsh23Kashyap 

Can you please create your branches of dev and point them to dev while making PRs? 🙏

PR #1450, opened 2026-09-17
fix(local): reject +-inf in vectors at write time

warter666:
validate_dense_vector and validate_multivector checked isnan, so NaN was rejected at write time while +-inf slipped through, despite both docstrings claiming parity with the server. Once stored, an infinity corrupts silently:

* COSINE normalises it into NaN, and the corrupted point then sorts to the top with a NaN score, ahead of an exact match;
* DOT and EUCLID score it inf;
* an inf among a sparse vector's values scores inf.

Check isfinite over the float32 values that actually get stored, so a value that is finite as a Python float but not as float32 (1e40) is caught as well, and keep numpy's overflow warning out of the user's output.

Query-time checks in distances.py and multi_distances.py still reject only NaN; they are left alone here to keep the change reviewable.

### All Submissions:

### New Feature Submissions:

### Changes to Core Features:


warter666:
Both review points are addressed in 529fabd:

**Write vs query validation split.** `validate_sparse_vector` runs on query paths as well (search inputs, `_recommend_average_sparse` positives/negatives, sparse distance queries), so widening it to reject ±inf changed query behaviour beyond this PR's scope. It is restored to the NaN-only contract, and a new `validate_sparse_vector_at_write` adds the finite (float32) check for stored values. The two write sites — point persistence and `_validate_named_vectors` — switch to it; all query call sites are untouched. Dense/multivector validators needed no split: `_validate_dense_or_multivector` is write-path only.

**Docstring coverage.** Added docstrings to every function touched by the diff (`validate_sparse_vector`, `validate_sparse_vector_at_write`, `to_jsonable_python`, the modified tests, and `assert_internally_consistent`).

The unchanged query-time contract is now pinned by `test_sparse_query_paths_still_accept_infinity`: a sparse query carrying ±inf validates, executes, and returns the expected point. Full `qdrant_client/local/tests` passes (151 tests).

PR #1466, opened 2026-09-21
Fix #1454: Local mode: sparse queries keep zero-score points, while the server's 

### LINKED ISSUE #1454: Local mode: sparse queries keep zero-score points, while the server's indexed path drops them
## What I observed

In local mode (`:memory:`), a sparse query returns points whose dot product is exactly `0.0`
(indices overlap, but the products cancel or a weight is `0.0`). The server drops them —
but only on its indexed search path. Its plain (small-cardinality) path keeps them.

Versions: qdrant-client 1.19.1, qdrant server 1.19.1 (docker). Same on `master` of this repo.

## Reproduction

```python

SV = models.SparseVector
N = 60
QUERY = SV(indices=[1, 2], values=[1.0, 1.0])

def run(client, k):
    """N points; point 1 scores 0.0 against QUERY, the rest score 2.0.
    Points 1..k get tag='in' so a filter on tag='in' passes exactly k points."""
    name = "sparse_zero"
    if client.collection_exists(name):
        client.delete_collection(name)
    client.create_collection(
        name,
        vectors_config={},
        sparse_vectors_config={"s": models.SparseVectorParams(
            index=models.SparseIndexParams(full_scan_threshold=10))},
    )
    client.create_payload_index(name, "tag", models.PayloadSchemaType.KEYWORD)
    client.upsert(name, wait=True, points=[
        models.PointStruct(
            id=i,
            vector={"s": SV(indices=[1, 2], values=[1.0, -1.0]) if i == 1
                    else SV(indices=[1, 2], values=[1.0, 1.0])},
            payload={"tag": "in" if i <= k else "out"},
        ) for i in range(1, N + 1)
    ])
    flt = models.Filter(must=[models.FieldCondition(key="tag", match=models.MatchValue(value="in"))])
    no_filter = client.query_points(name, query=QUERY, using="s", limit=N).points
    with_filter = client.query_points(name, query=QUERY, using="s", query_filter=flt, limit=N).points
    return [p.score for p in no_filter if p.id == 1], [p.score for p in with_filter if p.id == 1]

for label, client in (("server", QdrantClient(url="http://localhost:6333")),
                      ("local ", QdrantClient(":memory:"))):
    for k in (5, 60):
        a, b = run(client, k)
        print(label, f"k={k:<3}", "no_filter:", a, " with_filter:", b)
```

Output (`[0.0]` = point 1 returned with score 0.0, `[]` = dropped):

```
server k=5   no_filter: []  with_filter: [0.0]
server k=60  no_filter: []  with_filter: []
local  k=5   no_filter: [0.0]  with_filter: [0.0]
local  k=60  no_filter: [0.0]  with_filter: [0.0]
```

Negative scores are returned on both sides; only exact `0.0` differs.

## Why it matters

The result count and `limit` consumption differ between the two, so a query that passes
against local mode can behave differently against a real server — which is exactly what
local mode is meant to prevent.

## Why the server differs from itself

`search_nearest_query` (`lib/segment/src/index/sparse_index/sparse_vector_index/read_view/search.rs`)
picks the path by filter cardinality:

- no filter, or cardinality >= `full_scan_threshold` → `search_sparse` → `SearchContext::search`.
  `advance_batch` publishes only `score != 0.0` (`lib/sparse/src/index/search_context.rs`).
- filter with cardinality < `full_scan_threshold` → `search_plain` → `SearchContext::plain_search`,
  which skips only empty overlaps (`values.is_empty()`) and then pushes every score as-is.

Local mode (`qdrant_client/local/sparse_distances.py::sparse_dot_product`) returns a score
whenever indices overlap, and `local_collection.py` only skips `-inf` (no overlap), so `0.0`
is always kept.

## Question

Which behavior should local mode match — drop `0.0` like the indexed path, or keep it like the plain path?
Happy to open a PR once the intended behavior is clear.

nandanadileep:
Fixes #1454

## What changed
Implemented the focused fix requested in #1454 across `qdrant_client/local/local_collection.py`.

## Verification
`python -m pytest -q -x tests/test_local_persistence.py tests/test_qdrant_client.py tests/test_migrate.py tests/test_async_qdrant_client.py tests/test_in_memory.py qdrant_client/local/tests/test_formula.py qdrant_client/local/tests/test_write_atomicity.py qdrant_client/local/tests/test_siphash.py` passed


ljo1010:
This drops `0.0` for every sparse query type, but the server only does that on the nearest-search path. For sparse recommend (`sum_scores`) and context queries the server returns `0.0` scores, and local mode currently matches it — this patch breaks that.

Repro (client 1.19.1, server 1.19.1):

```python
SV = models.SparseVector
name = "t"
pts = [(1, SV(indices=[1], values=[1.0])), (2, SV(indices=[1], values=[1.0])),
       (3, SV(indices=[9], values=[1.0])), (4, SV(indices=[1, 2], values=[1.0, -1.0]))]
for tag, c in (("server", QdrantClient(url="http://localhost:6333")), ("local", QdrantClient(":memory:"))):
    if c.collection_exists(name): c.delete_collection(name)
    c.create_collection(name, vectors_config={}, sparse_vectors_config={"s": models.SparseVectorParams()})
    c.upsert(name, points=[models.PointStruct(id=i, vector={"s": v}) for i, v in pts], wait=True)
    r = c.query_points(name, query=models.RecommendQuery(recommend=models.RecommendInput(
        positive=[1], strategy=models.RecommendStrategy.SUM_SCORES)), using="s", limit=10).points
    print(tag, "sum_scores", sorted((p.id, p.score) for p in r))
    r = c.query_points(name, query=models.ContextQuery(context=[models.ContextPair(positive=2, negative=3)]), using="s", limit=10).points
    print(tag, "context   ", sorted((p.id, p.score) for p in r))
```

Before this patch:
```
server sum_scores [(2, 1.0), (3, 0.0), (4, 1.0)]
server context    [(1, 0.0), (4, 0.0)]
local sum_scores [(2, 1.0), (3, 0.0), (4, 1.0)]
local context    [(1, 0.0), (4, 0.0)]
```
With this patch:
```
local sum_scores [(2, 1.0), (4, 1.0)]
local context    []
```

If the intended behavior is to match the indexed path, the check should be limited to plain `SparseVector` nearest queries rather than keyed on `sparse_scoring`. Also, #1454 is still an open question to the maintainers — the server itself keeps `0.0` on its small-cardinality path.

Issue #1465, opened 2026-09-21
Local `scroll(order_by=...)` returns the same point more than once for array payload values


Harsh23Kashyap:
## Reproduction

On current `dev` (`bdee947`):

```python
from qdrant_client import QdrantClient, models

client = QdrantClient(":memory:")
client.create_collection(
    "items",
    vectors_config=models.VectorParams(size=1, distance=models.Distance.COSINE),
)
client.upsert(
    "items",
    [
        models.PointStruct(id=1, vector=[1.0], payload={"rank": [1, 2]}),
        models.PointStruct(id=2, vector=[1.0], payload={"rank": [3]}),
    ],
)

records, _ = client.scroll("items", order_by="rank", limit=10)
print([(record.id, record.order_value) for record in records])
```

Output:

```text
[(1, 1), (1, 2), (2, 3)]
```

I ran the reproduction twice with identical output.

## Expected behavior

A point should appear at most once in a scroll page. When an ordered payload field contains several values, local mode should choose one ordering value using the same rule as the server.

## Actual behavior

Local mode emits one record for each distinct `(value, point_id)` tuple. Point `1` therefore appears twice. The current deduplication set is keyed by `(value, external_id)`, so it does not deduplicate a point that has multiple values.

## Evidence

Attached screenshot shows both identical runs on `dev`.

joein:
Hey @Harsh23Kashyap 

Local mode works the same way the remote server does

Remote:

```python
from qdrant_client import QdrantClient, models

client = QdrantClient()
if client.collection_exists("items"):
    client.delete_collection("items")

client.create_collection(
    "items",
    vectors_config=models.VectorParams(size=1, distance=models.Distance.COSINE),
)
client.create_payload_index("items", "rank", models.PayloadSchemaType.INTEGER)
client.upsert(
    "items",
    [
        models.PointStruct(id=1, vector=[1.0], payload={"rank": [5, 6]}),
        models.PointStruct(id=2, vector=[1.0], payload={"rank": [7]}),
    ],
)

records, _ = client.scroll("items", order_by="rank", limit=10)
print([(record.id, record.order_value) for record in records])
```
Output:
```
[(1, 5), (1, 6), (2, 7)]
```

If we replace `client = QdrantClient()` with `client = QdrantClient(":memory:")`, we get the same output

This is the intended behaviour of this method.

Closing this for now, please re-open if you need more details 🙏

PR #1467, opened 2026-09-22
fix(local): deduplicate array-valued ordered scrolls

DeepanshuPal:
## Summary

Local ordered scroll expands array-valued payload fields into one candidate per value. It currently deduplicates `(value, point_id)`, so the same point is returned multiple times when its array has distinct values.

This change deduplicates by point ID after sorting. The first eligible value becomes the record's order value, which gives the minimum value for ascending scans and the maximum value for descending scans.

Fixes #1465.

## Verification

- added regression coverage for ascending and descending array-valued order fields
- `pytest -q tests/congruence_tests/test_scroll_order_by.py::test_scroll_array_values_return_each_point_once` (1 passed)
- `git diff --check`

The pre-existing remote congruence test requires a Qdrant server on localhost:6333; it was not available in this environment.


joein:
Hey @DeepanshuPal 

Thanks for the interest in the project

Though, the issue was wrong and local mode's behaviour is the same as the server's one, so there is nothing to fix

